PR #88, opened 2026-01-05
Workflow 10: add date range inputs for testing

clarkemoyer:
Adds optional start_date, end_date, and max_rows inputs to Workflow 10 so we can run smaller test imports (and cap export size) while validating the Zeffy CSV header guardrails.


clarkemoyer:
Closing as obsolete: workflow 10 already includes start_date/end_date inputs on main, and this PR is now conflicting. If we need more testing-specific inputs, we can open a fresh PR off current main.

PR #346, opened 2026-02-25
fix: avoid false failures from LASTEXITCODE

clarkemoyer:
When scripts/Create-GitHubRepo.ps1 succeeds, a stale native $LASTEXITCODE can still cause the GitHub Actions step to end with exit code 1.\n\nThis resets $global:LASTEXITCODE = 0 immediately after invoking the script in the two workflow entrypoints that call it.


clarkemoyer:
### 🔍 Premium AI Code Review Summary

Hi @clarkemoyer! Here is a focused review of your proposed fix to prevent false failures from `$LASTEXITCODE`.

---

### 💡 Core Insights & Potential Issues

> [!WARNING]
> **Stale `$LASTEXITCODE` Risk & PowerShell Behavior**
> In PowerShell, running a script file (`.ps1`) directly using `./scripts/Create-GitHubRepo.ps1` **does not** native-compile or execute as an external process. Therefore, executing it **does not reset or update `$LASTEXITCODE`** unless the script itself runs an external executable (like `gh.exe` or `git.exe`) as its very last action.
> 
> If a previous native command in the same workflow step failed and set a non-zero exit code, running `Create-GitHubRepo.ps1` successfully will **not** clear `$LASTEXITCODE`. Thus, the new check:
> ```powershell
> if ($LASTEXITCODE -ne 0) {
>   throw "Create-GitHubRepo.ps1 failed with exit code $LASTEXITCODE"
> }
> ```
> could still trigger a **false positive failure throw** using that stale value.

> [!NOTE]
> **PR Description Discrepancy**
> The pull request description states:
> *"This resets `$global:LASTEXITCODE = 0` immediately after invoking the script..."*
> However, looking at the diff, there is no assignment to `$global:LASTEXITCODE = 0` or `$LASTEXITCODE = 0`. It only checks and throws.

---

### 🛠️ Actionable Recommendations

To make this completely bulletproof, you can explicitly reset or check the execution status of the script rather than relying solely on `$LASTEXITCODE`.

#### Option A: Explicitly Reset `$LASTEXITCODE` Prior to Running the Script
If you want to ensure the exit code belongs to native commands *within* `Create-GitHubRepo.ps1`, clear it first:
```powershell
$global:LASTEXITCODE = 0
./scripts/Create-GitHubRepo.ps1 @params
if ($LASTEXITCODE -ne 0) {
  throw "Create-GitHubRepo.ps1 failed with exit code $LASTEXITCODE"
}
```

#### Option B: Check PowerShell Error Status (`$?`)
Since `Create-GitHubRepo.ps1` runs in-process, checking `$?` (which indicates if the last command executed successfully without unhandled errors) is often safer for `.ps1` scripts:
```powershell
./scripts/Create-GitHubRepo.ps1 @params
if (-not $?) {
  throw "Create-GitHubRepo.ps1 execution failed."
}
```

---

*Reviewed by Antigravity under your AI Max / Joules plan.*

Issue #27, opened 2025-12-31
Remove all Terraform from this repo


clarkemoyer:
Remove all terraform code and tests from this repo and all refences to it in documentation except for one line in the main readme about deprecated features "Terraform"

PR #367, opened 2026-05-18
feat: bump workflow 19 smoke poll deadline + parallelize dispatch

clarkemoyer:
## Summary

- Splits the post-cutover-smoke job into two steps: dispatch all smokes upfront, then poll each
- Smokes now run in parallel on GHA — total wait converges to ~25 min (slowest single smoke), not 25 * N min
- Initial DNS-propagation sleep 60s -> 120s
- Per-repo poll deadline 5 min -> 25 min (matches new smoke's internal cert-retry budget of 20 min plus Playwright install + screenshot + buffer)
- Error message points operators at the smoke-visual-* artifact for screenshots

## Why

PR #27 in FFC-EX-amargraves.org introduced a cert-aware retry loop in the per-repo smoke that retries HTTPS every 30s for up to 20 min, waiting for Let's Encrypt provisioning. The previous 5-min poll deadline here would time out before the smoke even reached a terminal state.

## Test plan
- [ ] Dry-run dispatch (skip)
- [ ] Single-domain real dispatch: confirm smoke completes within deadline, screenshot artifact accessible from cutover summary

### DIFF
--- a/.github/workflows/19-bulk-cutover-to-github-pages.yml
+++ b/.github/workflows/19-bulk-cutover-to-github-pages.yml
@@ -156,14 +156,17 @@ jobs:
             exit 1
           fi
 
-      - name: Wait for DNS propagation and Pages re-bind
+      - name: Wait for DNS propagation
         shell: bash
         run: |
-          # Apex DNS just flipped; give resolvers ~60s and GH Pages a moment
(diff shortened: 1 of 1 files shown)


clarkemoyer:
Closing in favor of #368 (rebased onto current main; force-push not authorized in this env)

PR #401, opened 2026-06-07
feat(sites-list): generate merged sites list (CSV + JSON) in env-owning repo

clarkemoyer:
## What

Relocates **FFC sites-list generation** into this repo — the one that already owns the export environments (`whmcs-prod`, `cloudflare-prod-read`, `wpmudev-prod`) and the GitHub PAT (`CBM_TOKEN` in `github-prod`) required to dispatch the sibling export workflows. Output is published as both **CSV** (human/spreadsheet, line-based diffs) and **JSON** (structured), which downstream repos pull without credentials.

### Why
The previous generator lived in `FFC-IN-ffcadmin.org`, where the `GH_PAT`/`CBM_TOKEN` secret does not exist (it's an environment secret here). That job had failed **22/22 scheduled runs** because the secret resolved empty. Moving generation to where the secret/environments live removes that whole class of failure.

### Files
- `scripts/update-sites-data.mjs` — merges the three export artifacts (`whmcs_domains`, `domain_summary`, `wpmudev-domain-inventory`) with the curated base list, runs per-domain HTTPS health checks, writes `sites-list/sites_list.csv` + `sites-list/sites_list.json`. **Non-destructive:** membership flags (`In WHMCS`/`In Cloudflare`/`In WPMUDEV`) are preserved when a source export is absent, instead of being overwritten with `No`.
- `.github/workflows/sites-list-generate.yml` — weekly + manual; runs in `environment: github-prod`, validates `CBM_TOKEN`, dispatches the exports **by stable file name** (`7-whmcs-export-domains.yml`, `4-export-summary.yml`, `13-wpmudev-export-sites.yml`), downloads artifacts, regenerates, and commits to `main`.
- `sites-list/` — seed data carried over from the curated ffcadmin list (356 domains) + README with the public raw URLs.
- `.gitignore` — ignore `node_modules/` and `tmp_data/` created at runtime.

### Consumes downstream
`FFC-IN-ffcadmin.org` will pull the published files via raw URL (separate PR), so it needs **no secrets**.

## Review notes
- The dispatch→`gh run list --limit 1` step inherits the prior design's small race window; acceptable for a scheduled job but can be hardened to match dispatched run IDs if you prefer.
- Seeded `Section` is `Unknown` for many domains (carried from the current data); curate in `sites-list/sites_list.csv` and it's preserved on regeneration.
- Pushes directly to `main` with `CBM_TOKEN`, matching the existing pattern in `15-website-provision.yml`.

Draft for your review — nothing here runs until you dispatch it.


clarkemoyer:
Thanks @copilot — addressed the substantive points across commits cbf2751 and b08bac4:

- WHMCS rows keyed by the normalized `domain` column (fallback `domainname`) and included in the domain union.
- Export artifacts resolved by filename under their dir, so a nested upload path (e.g. `artifacts/whmcs/`) is handled.
- Health-check timeout cleared in `finally`; fetch body cancelled to release sockets.
- Curated rows looked up via an O(1) `manualMap`.
- Workflow: added a `concurrency` group, restricted run selection to `workflow_dispatch` with a null guard, guarded commit/push to `refs/heads/main` only, and pinned `csv-parse`/`csv-stringify` to `^6`.

The `sites-list/README.md` table is valid GitHub-flavoured Markdown (single leading `|` per row), so that note is a false positive. Resolving these threads.

PR #480, opened 2026-06-30
Zeffy: per-endpoint exports, drop PII workflow

clarkemoyer:
## Summary

Restructures the Zeffy export into **one dispatch-only workflow per endpoint** and removes the PII-emitting workflow entirely, so the public repo never produces a donor-PII artifact.

This follows the earlier discovery that the combined export had written donor PII (payments buyer email/name; contacts email/phone/name/address) into a public-repo artifact. The fix is structural: no workflow on the public repo writes PII at all.

## Changes

**Workflows (one per endpoint, all `workflow_dispatch`-only, `zeffy-prod` env, OIDC → Key Vault):**

- `44-zeffy-campaigns-export.yml` — **Campaigns Export.** `GET /api/v1/campaigns` returns no personal data, so the artifact (`zeffy_campaigns`) is inherently safe.
- `45-zeffy-payments-export.yml` — **Payments Export (PII masked).** Runs the payments script **without** `-IncludePii`, so buyer email/name/company and receipt URL are blanked. Artifact `zeffy_payments`.
- `46-zeffy-contacts-export.yml` — **Contacts Export (PII masked).** Runs the contacts script **without** `-IncludePii`; email/name/phone/address blanked, only the pseudonymous UUID, giving totals/counts/dates, and country kept. Artifact `zeffy_contacts`.
- **Deleted** `45-zeffy-export-pii.yml` — there is no PII-emitting workflow on the public repo.

**Scripts:** `zeffy-payments-export.ps1` and `zeffy-contacts-export.ps1` keep the `-IncludePii` switch (default off → masked) for **local/private** runs only. No workflow passes it.

**Docs:** `docs/zeffy-api.md` and `.github/workflows/README.md` updated to the per-endpoint, no-PII-workflow model. Also documents explicitly that the **Zeffy API is read-only** — every spec operation is a `GET`; there are no `POST`/`PUT`/`PATCH`/`DELETE` endpoints, so campaigns/contacts/payments cannot be created or modified via the API. The only inbound channel is the `payment.completed` webhook (data delivery, not a write API).

## Validation

- All three workflow YAMLs parse.
- `prettier --check` clean on the workflows, `docs/zeffy-api.md`, and the README.
- PSScriptAnalyzer / `Invoke-Formatter` gates run in CI (pwsh not available in this sandbox); script `=` alignment unchanged from the green #479 baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #481, opened 2026-06-30
Repo-wide security hardening: WHMCS host allowlist, action SHA pins, input hygiene

clarkemoyer:
## Summary

A repo-wide safety sweep (four parallel audit passes: URL-override credential leaks, script injection, PII exposure, supply chain / permissions), applying the unambiguous, low-risk fixes. This extends the Zeffy `base_url` hardening (#480) to the rest of the repo.

**No behavior change for legitimate runs:** the default APIM/origin hosts are in the allowlist, and each pinned SHA is the exact commit the previously-referenced tag pointed at.

### 1. WHMCS credential exfil via `api_url` (highest severity — same class as the Zeffy fix)
**9 WHMCS workflows** (`7, 8, 9, 10, 38, 39, 41, 42, 43`) accept a user-overridable `api_url` dispatch input that flows into a request carrying the WHMCS **identifier/secret + APIM subscription key**. A dispatcher could point it at an attacker-controlled host and capture the credential.

Fix: a host-allowlist guard at **every request chokepoint** — `https` only, host ∈ `{apim-ffc-gateway-prod.azure-api.net, freeforcharity.org}`:
- `scripts/whmcs-api-common.ps1` → `Invoke-WhmcsApi` (covers all triage/write/onboard scripts)
- the four self-contained exports: `whmcs-domain-export.ps1`, `whmcs-products-export.ps1`, `whmcs-payment-methods-export.ps1`, `whmcs-clients-export.ps1`

This protects every current and future caller regardless of where `api_url` originates.

### 2. Supply chain — pin third-party actions to SHA
Pinned all genuinely third-party actions to full commit SHAs (kept a `# vN` comment), so a compromised tag can't inject code:
- `azure/login` v3 → `532459e…` (×13)
- `peter-evans/create-pull-request` v8 → `5f6978f…`
- `EndBug/label-sync` v2 → `5207415…`

### 3. Input hygiene — `3-manage-record.yml`
Moved `record_type` / `record_name` / `record_content` (and `domain`) into `env:` instead of interpolating them directly into the pwsh `run:` block. `record_content` is a free-form string and was the real (latent) injection vector; the others are constrained but moved for consistency.

### 4. PII defense-in-depth — `whmcs-client-add.ps1`
Mask the donor email in the duplicate-client error message (it can surface in workflow logs).

## Audit items intentionally NOT changed (surfaced for your decision)
- **`0-domain-status.yml`** carries `pull-requests: write` that appears unused — candidate for removal, but wanted confirmation it isn't needed by a step I didn't trace.
- **Permission-model consistency**: several workflows declare `id-token: write` at job level rather than top level. Cosmetic; the issue-triggered workflows (15/20/36) that hold `id-token` are already label-gated and need it for OIDC→Key Vault. (One audit suggestion, `id-token: read-metadata`, isn't a valid value and was discarded.)
- **First-party `actions/*`** remain on major-version tags (acceptable per common practice); can pin to SHA later if desired.

## Validation
- All workflow/action YAML parses; `prettier --check` clean.
- PowerShell `Invoke-Formatter` / PSScriptAnalyzer run in CI (pwsh unavailable in this sandbox). The inserted guard blocks contain no aligned-hashtable assignments, so formatter alignment is unaffected.

---

### DIFF
--- a/.github/actions/cloudflare-tokens-from-kv/action.yml
+++ b/.github/actions/cloudflare-tokens-from-kv/action.yml
@@ -49,7 +49,7 @@ runs:
         }
 
     - name: Azure login (OIDC)
-      uses: azure/login@v3
+      uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
       with:
         client-id: ${{ inputs.azure-client-id }}
         tenant-id: ${{ inputs.azure-tenant-id }}
--- a/.github/actions/whmcs-secrets-from-kv/action.yml
+++ b/.github/actions/whmcs-secrets-from-kv/action.yml
@@ -67,7 +67,7 @@ runs:
         }
 
     - name: Azure login (OIDC)
-      uses: azure/login@v3
+      uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
       with:
         client-id: ${{ inputs.azure-client-id }}
         tenant-id: ${{ inputs.azure-tenant-id }}
--- a/.github/actions/zeffy-secrets-from-kv/action.yml
+++ b/.github/actions/zeffy-secrets-from-kv/action.yml
@@ -53,7 +53,7 @@ runs:
         }
 
     - name: Azure login (OIDC)
-      uses: azure/login@v3
+      uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
       with:
         client-id: ${{ inputs.azure-client-id }}
         tenant-id: ${{ inputs.azure-tenant-id }}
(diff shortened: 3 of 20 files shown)


clarkemoyer:
Addressed the `REC_CONTENT` comment in `8742ccf`. The folded value actually parsed correctly (YAML folds the wrapped newline to a single space, so the default stayed `v=spf1 include:spf.protection.outlook.com -all`), but relying on that is brittle — so I shortened each `env:` line and moved the `TXT`/`@`/SPF defaults into the pwsh body, where they can't be wrapped. Inputs are still passed via `env:` (no direct interpolation into the run block).

---

PR #483, opened 2026-06-30
Quality audit: Cloudflare token export hardening (heredoc parity)

clarkemoyer:
## Summary

Output of a second multi-agent audit pass focused on **correctness and quality** (PowerShell robustness, workflow quality, dry-run/masking correctness, consistency/dead-code). Every candidate finding was verified against the actual code; this PR contains the **one** fix that survived verification as both real and low-risk.

### Fix
`cloudflare-tokens-from-kv/action.yml` wrote tokens to `GITHUB_ENV` with `Add-Content "KEY=$value"`, whereas the WHMCS and Zeffy `secrets-from-kv` actions use a randomized **heredoc delimiter** (`Add-EnvVar`) so a value containing a newline (or a delimiter-like line) can't inject extra env vars into later steps. This brings the Cloudflare action to the same hardened, consistent pattern. No behavior change for normal single-line tokens.

## Verified, but intentionally NOT changed
Most agent findings were **false positives** on inspection:
- **dry-run logic** (`if (dry_run -ne 'false') { -DryRun }`) already fails *safe* — the suggested `-eq 'true'` rewrite would make empty input run **live**. Left as-is.
- **Zeffy pagination / `[double]` casts / contact-add dry-run / bulk-staging exit code / retry "off-by-one"** — all checked and correct (e.g. `bulk-staging` `exit 1`s on failure and is called in-process; the retry loop issues all 5 attempts).

## Real findings deferred (need your call — broad or judgment-based)
1. **WHMCS export-script duplication**: `whmcs-domain-export.ps1` reimplements `Invoke-WhmcsApi` and — unlike the others and the shared helper — **lacks the Imunify360 retry/backoff**, so it fails hard on transient bot-protection blocks. Fixing well means either porting the retry logic or refactoring the 4 self-contained exports to dot-source `whmcs-api-common.ps1` (larger change; they're standalone by design).
2. **`timeout-minutes` / `concurrency`**: only 3 of ~55 workflows set either. Worth adding to hang-prone/stateful ones (httrack clone #27, long pollers #19, import #10), but `concurrency` cancel-vs-queue semantics on deploy workflows is a judgment call.
3. **Deprecated `WHMCS_API_ACCESS_KEY`**: still referenced in ~10 files; removable per CLAUDE.md, but touches many files and `accesskey` is a legitimate WHMCS auth param.
4. **Workflow numbering**: filename prefixes intentionally differ from display numbers; aligning them would mean mass file renames (breaking run history / dispatch URLs) — not recommended.

Happy to take on any of #1–#3 as its own PR if you'd like.

## Validation
- `prettier --check` clean; YAML parses. PSScriptAnalyzer/`Invoke-Formatter` run in CI.

---

### DIFF
--- a/.github/actions/cloudflare-tokens-from-kv/action.yml
+++ b/.github/actions/cloudflare-tokens-from-kv/action.yml
@@ -96,10 +96,24 @@ runs:
         Write-Host "::add-mask::$ffc"
         Write-Host "::add-mask::$cm"
 
+        # Write to GITHUB_ENV with a randomized heredoc delimiter (matching the WHMCS/Zeffy
+        # secrets-from-kv actions) so a token that ever contains a newline (or a line that looks
+        # like a delimiter) cannot inject extra env vars into later steps.
+        function Add-EnvVar {
+          param([string]$Name, [string]$Value)
+          $delim = "ghadelim_$([guid]::NewGuid().ToString('N'))"
+          if ($Value -match [regex]::Escape($delim)) {
+            throw "Refusing to write '$Name': value collides with the generated heredoc delimiter."
+          }
+          Add-Content -Path $env:GITHUB_ENV -Value "$Name<<$delim"
+          Add-Content -Path $env:GITHUB_ENV -Value $Value
+          Add-Content -Path $env:GITHUB_ENV -Value $delim
+        }
+
         # Export to GITHUB_ENV so subsequent steps see them as plain env vars,
         # matching the pre-refactor interface (CLOUDFLARE_API_TOKEN_FFC / _CM).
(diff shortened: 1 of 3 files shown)


clarkemoyer:
Good catch — addressed in `6317761`. Added a `Protect-Secret` helper that splits the value on CR/LF and masks each non-blank line, so a multi-line secret can't leak its later lines (and a `::`-prefixed line can't be read as a workflow command). Routed all masking through it in the Cloudflare, WHMCS, **and** Zeffy actions so the three stay consistent with the heredoc env-write hardening. No change for single-line tokens.

---

PR #484, opened 2026-06-30
whmcs-domain-export: add Imunify360 retry/backoff (parity with common helper)

clarkemoyer:
## Summary

Audit follow-up #1. `whmcs-domain-export.ps1` reimplements `Invoke-WhmcsApi` but — unlike the shared `whmcs-api-common.ps1` and the other WHMCS scripts — issued a **single `Invoke-RestMethod` with no retry**. A transient Imunify360 bot-protection challenge (or a 429/502/503/504) on the runner IP therefore failed the export hard, while every other WHMCS call retries with backoff.

This ports the **same proven retry loop** from `whmcs-api-common.ps1` into domain-export's `Invoke-WhmcsApi`:
- Detects the transient signature (`Imunify360|bot-protection|429|502|503|504|timeouts`).
- Retries up to `WHMCS_API_MAX_ATTEMPTS` (default 5) with capped exponential backoff.
- Still **fails fast** on genuine auth/permission errors (they don't match the transient regex).
- Preserves the existing host-allowlist guard and domain-export's richer error diagnostics.

No interface change; the function signature and return shape are unchanged.

## Note
This addresses the retry gap specifically. The deeper duplication (4 self-contained export scripts each reimplementing the helper) is left for a possible later consolidation — out of scope here to keep the change low-risk.

## Validation
- PSScriptAnalyzer / `Invoke-Formatter` run in CI (pwsh unavailable in this sandbox). The ported block mirrors the production-proven loop and introduces no aligned-hashtable assignments.

---

### DIFF
--- a/scripts/whmcs-api-common.ps1
+++ b/scripts/whmcs-api-common.ps1
@@ -101,8 +101,9 @@ function Invoke-WhmcsApi {
     # errors do not match and still fail fast.
     $transientRe = 'Imunify360|bot-protection|too many requests|temporarily unavailable|timed out|The operation has timed out|\b(429|502|503|504)\b'
     $maxAttempts = 5
-    if ($env:WHMCS_API_MAX_ATTEMPTS -and ([int]::TryParse($env:WHMCS_API_MAX_ATTEMPTS, [ref]([int]$null)))) {
-        $maxAttempts = [int]$env:WHMCS_API_MAX_ATTEMPTS
+    $parsedMax = 0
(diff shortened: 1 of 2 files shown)


clarkemoyer:
Good catch — fixed in `4182df2`. The override now parses into a dedicated variable and is only honored when `>= 1`; a `0`/negative/garbage value keeps the default of 5 instead of silently disabling retries. Applied the same clamp to `whmcs-api-common.ps1`, which had the identical pattern, so the two stay consistent.

---

Issue #496, opened 2026-06-30
Verify &amp; document the real environment approval-gate (reviewer) config


clarkemoyer:
Parent: #487

The safety doc's central claim — which environments require a reviewer — is sourced from `CLAUDE.md`, not from live **Settings → Environments**. Confirmed-gated today: `cloudflare-prod-write`, `whmcs-prod`, `github-prod`. **Unknown:** `m365-prod`, `wpmudev-prod`, `zeffy-prod`, and the `*-read` envs.

**Goal:** confirm the actual required-reviewer configuration for every environment and make the doc definitive (remove the "confirmed to require a reviewer" hedge).

Needs a human with repo admin to read/screenshot Settings → Environments (the automation can't read environment protection rules). Then update `docs/workflow-safety-and-approvals.md` and `docs/github-actions-environments-and-secrets.md` to state the real gating per environment.

clarkemoyer:
**Blocked on a human with repo admin** — the automation can't read **Settings → Environments** protection rules (that data isn't exposed to the workflow token / MCP). So this needs someone to confirm, for each environment, whether a **required reviewer** is configured:

- Confirmed gated (per CLAUDE.md + the website-provision flow): `cloudflare-prod-write`, `whmcs-prod`, `github-prod`.
- **Please verify:** `m365-prod`, `wpmudev-prod`, `zeffy-prod`, and the `*-read` environments.

Once confirmed, I'll update `docs/workflow-safety-and-approvals.md` (remove the "confirmed to require a reviewer" hedge and mark the real ✅ set) and `docs/github-actions-environments-and-secrets.md`. The safety doc is written honestly in the meantime. This particularly affects **#498** (workflow 24 / M365 Add Tenant is a write whose only gate would be `m365-prod`).

---
_Generated by [Claude Code](https://claude.ai/code)_

clarkemoyer:
Attempted to read the environment reviewer config directly via the `gh` CLI (installed `gh` 2.95.0 in the session). Result: the agent proxy in this web sandbox **blocks the environments/repo-settings API paths** regardless of client —

- `gh api repos/FreeForCharity/FFC-Cloudflare-Automation/environments` → `403 Access to this GitHub API path is not permitted through this proxy`
- `gh api repos/FreeForCharity/FFC-Cloudflare-Automation` → `403 GitHub access is not enabled for this session`

[...]

Issue #487, opened 2026-06-30
Docs: close admin/training gaps for recent capabilities (lifecycle, safety model, Zeffy/WHMCS ops)


clarkemoyer:
## Why

A review of the admin/training docs against the full set of capabilities (with emphasis on the recent additions — Zeffy read API, WHMCS support/orders/products workflows 38–43, and the OIDC→Key Vault / security hardening in #480–#486) found that the **reference docs are strong**, but three higher-order things are under-expressed for an administrator:

1. **End-to-end narratives** — the journeys an admin actually performs span many workflows but have no single runbook.
2. **Operating policy / decision-trees** — support-template selection, international eligibility, orders/fraud triage, dry-run/approval guarantees.
3. **What the recent hardening changed** — new guarantees (no donor PII on the public repo, WHMCS host allowlist, serialized cutovers, SHA-pinned actions) aren't surfaced anywhere an operator would look.

## Scope (prioritized)

**Immediate**
- [ ] `docs/charity-onboarding-lifecycle.md` — end-to-end runbook: domain → Cloudflare zone → M365 + DKIM → WHMCS client/contacts/order → website provision (15) → support routing. Per-step success criteria + rollback. Ties together workflows 01–04, 24, 34, 15, 38/39.
- [ ] `docs/workflow-safety-and-approvals.md` — per-workflow write level (read-only vs live), which gate behind environment approval, what `dry_run` guarantees, and the concurrency/serialization + host-allowlist + SHA-pinning guarantees added in #480–#486.

**High**
- [ ] `docs/zeffy-reconciliation.md` — operationalize the new Zeffy read exports (44–46) + the WHMCS→Zeffy import (33): cadence, what "reconciled" means, common discrepancies (refunds/recurring/pending).
- [ ] `docs/whmcs-orders-triage-policy.md` — decision tree for the Pending/Fraud/Active backlog; fraud heuristics; accept/hold/cancel SLA.
- [ ] Expand `docs/whmcs-support-tickets.md` — template decision matrix + international-eligibility (TechSoup referral) checklist (currently only in code).
- [ ] Unify the OIDC→Key Vault story (extend `docs/github-actions-environments-and-secrets.md` or add `docs/security-hardening-and-oidc.md`): why it's safer, bootstrap, troubleshooting, rotation SLA.

**Medium / cleanup**
[...]

PR #506, opened 2026-06-30
ci: add read-only workflow to audit environment approval gates

clarkemoyer:
## Summary

Adds **99. Repo - Audit Environment Approval Gates** — a dispatch-only, **read-only** workflow that prints each GitHub Environment's **required reviewers**, wait timer, and deployment branch policy to the job summary (and the log).

This is the runner-side answer to **#496**: environment protection rules can't be read from the agent-sandbox proxy (confirmed with `gh` — `403 not permitted through this proxy`), but a GitHub runner can. It uses the built-in `GITHUB_TOKEN` (repo read is enough to read protection rules), so it needs **no environment and no approval gate** — it runs immediately on dispatch.

Also adds the row to the safety table so the new consistency guard stays green.

## How we'll use it
1. Merge this.
2. You dispatch it (Actions → "99. Repo - Audit Environment Approval Gates" → Run workflow) — I can't dispatch `workflow_dispatch` from the sandbox.
3. I read the result back via the Actions logs and make `docs/workflow-safety-and-approvals.md` definitive (resolving #496, and confirming whether `m365-prod` is gated).

## Validation
- YAML parses; `prettier --check` clean; consistency guard passes (56 workflows, 47 rows). `actionlint`/shellcheck run in CI.

Refs #487, #496.

---

### DIFF
--- a/.github/workflows/99-repo-audit-environment-gates.yml
+++ b/.github/workflows/99-repo-audit-environment-gates.yml
@@ -1,0 +1,58 @@
+name: '99. Repo - Audit Environment Approval Gates [Repo]'
+run-name: 'Audit environment approval gates'
+
+# Read-only. Lists every GitHub Environment's required reviewers, wait timer, and deployment
+# branch policy to the job summary (and the log). Answers "which environments gate a deployment
+# behind a human approval?" — which the docs hedge on because environment protection rules can't
+# be read from the agent sandbox proxy. Uses the built-in GITHUB_TOKEN (repo read is enough to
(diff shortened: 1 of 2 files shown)


clarkemoyer:
Good catch — fixed in `0f3c172`. A per-environment fetch failure now renders an explicit `⚠️ failed to read protection rules` marker instead of being silenced to `{}` (which, for a gate audit, could have misreported a gated env as ungated). The environment name is also URL-encoded (`jq @uri`) when building the API path.

---

PR #534, opened 2026-07-01
fix(ci): add merge_group triggers so the merge queue can function

clarkemoyer:
Root cause confirmed: ALLGREEN merge queue + zero `merge_group:` triggers → merge-group checks never run → 60-min timeout → nothing merges via the queue (0 merge_group runs in repo history). Adds `merge_group:` to 722-ci, 723-codeql, 727-phantom-revert-guard (whose job `if:` safely skips non-PR events). Will be empirically verified by running the NEXT PR through the queue.

### DIFF
--- a/.github/workflows/722-ci.yml
+++ b/.github/workflows/722-ci.yml
@@ -3,6 +3,7 @@ name: '722. Repo - CI Validate and Test [Repo]'
 on:
   pull_request:
     branches: [main]
+  merge_group:
   push:
     branches: [main]
 
--- a/.github/workflows/723-codeql-analysis.yml
+++ b/.github/workflows/723-codeql-analysis.yml
@@ -3,6 +3,7 @@ name: '723. Repo - CodeQL Security Analysis [Repo]'
 on:
   pull_request:
     branches: [main]
+  merge_group:
   push:
     branches: [main]
   schedule:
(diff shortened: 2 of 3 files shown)


clarkemoyer:
**Root-cause correction (post-verification).** After empirically testing the queue with #535, the original analysis in this PR was **partly wrong**:

**What actually blocked PRs** (verified): the repo requires **review-conversation resolution** before merge. #524 had **10 unresolved Copilot review threads** — that is why it sat `blocked` with auto-merge enabled and never entered the queue. #535 hit the same gate (`enqueuePullRequest` → *"A conversation must be resolved"*), plus *"waiting for CodeQL"* on fresh commits. Once threads were resolved and CodeQL finished, `enqueuePullRequest` worked and #535 **merged via the queue in ~90 seconds**.

**Why zero `merge_group` runs still exist:** the ruleset configures the ALLGREEN queue with **zero required status checks**, so the queue has nothing to validate on a merge group and merges immediately — no merge-group CI is ever built. The triggers added by this PR were therefore not the operative fix, but they are the **prerequisite** for the real hardening:

**Recommendation:** add `722 CI Validate and Test` (and optionally CodeQL) as **required status checks** in the ruleset. With the `merge_group:` triggers from this PR in place, the queue will then genuinely build and validate merge groups before merging — today it is effectively a pass-through.

Also for operators: the misleading `gh pr merge --auto` GraphQL "rate limit" error masks these gates — use `gh api graphql` `enqueuePullRequest` directly to see the true blocking reason.

PR #718, opened 2026-07-18
feat(preflight): verify reality before the gate in 720 create + 702 clone-deploy

clarkemoyer:
## Why

Operator directive after the 2026-07-18 incident: six workflow runs were dispatched (and human-approved) against a stale epic-#702 inventory — four `720` creates for repos that **already existed** (one under different casing: `FFC-EX-AllTypeTowing.com`), and a `702` re-clone against that same repo, which is a **finished migration already cut over on its apex**. All failed or were canceled harmlessly, but only by luck of loud failure modes. The per-target verification that would have prevented it is now a **technical control inside the workflows**, not a process recommendation.

### `720. Repo - Create GitHub Repo` — ungated `preflight` job
Fails fast when the repository name is already taken **in any casing** (API lookup is case-insensitive; the error reports the canonical name and points at 702/736 for regenerate/retire). A doomed create never reaches the `github-prod` gate, so it can't consume a human approval. Transient/unparseable API responses fail safe.

### `702. Domain - Deploy Static Clone` — ungated `preflight` job + `force` input
Before the gate, the preflight:
1. **Requires the target repo to exist** (clear "create it first with 720" error otherwise).
2. **Resolves canonical repo casing** via the API and feeds it downstream — github.io paths are case-sensitive, so a wrong-case guess probes (and displays) the wrong URL.
3. **Refuses to re-clone a live site** — default Pages URL serving 200, or apex serving 200 with a GitHub server header (the cut-over shape) — unless the new `force=true` input is explicitly passed.

### Latent CR bug fixed in `701`'s verify probe (found by the new tests)
HTTP header lines end in `\r`; the `awk` status-code parse kept it, so `"200\r" != "200"` — the 701 serving probe would have reported **every** live site as "not serving", and the new 702 preflight would have let cut-over sites through. Both parsers now `tr -d '\r'` first. The fake `curl` in the test harness emits real CRLF headers precisely so this stays pinned.

## Tests (AGENTS.md workflow-logic harness)
- `test_720_preflight.py` (3): case-variant name refused with canonical name; unparseable-but-existing response still refuses; free name proceeds.
- `test_702_preflight.py` (6): missing repo fails with 720 guidance; unparseable fails safe; live-on-default-URL refused; **cut-over-apex refused** (the AllTypeTowing shape); `force=true` overrides and emits canonical name; not-yet-live proceeds with canonical casing.
- `test_701_verify_probe.py` (3): serving detected despite CRLF headers (the regression anchor); non-GitHub server and non-200 report not-serving.
- Harness additions (additive): fake `curl` with real CRLF header output, no-op `sleep` so retry loops run instantly. All 8 modules pass.

## Checks
- actionlint (+shellcheck) clean on all three workflows (701's pre-existing input-count note unchanged)
- Prettier ✅ · consistency guard ✅ (87 workflows, 78 rows) · catalog regenerated
- Safety-table row 702 updated to name the preflight guard

---


clarkemoyer:
All three review findings addressed in 45af08f:

1. **Fail-open on org-list errors (both scans)** — valid and the most important catch: `2>/dev/null … || true` would have silently disabled the sibling guard on any transient API error, which defeats the point of a guard built to stop stale-inventory dispatches. Both preflights now capture the `gh api` exit separately and **refuse on listing failure** ("failing safe rather than guessing"); `force`/`Force=true` skips the scan explicitly and says so in the summary.
2. **720 domain extraction** — the original `${domain#*-*-}` shortest-match did strip exactly `FFC-EX-` (single `#` is shortest match, so `FFC-EX-the-trendylittlegeek.com` → `the-trendylittlegeek.com`), but it was fragile and unobvious; replaced with literal `FFC-EX-`/`ffc-ex-` prefix strips.
3. **Hyphenated-domain regression test** — added exactly as requested: `FFC-EX-the-trendylittlegeek.com` scans for `the-trendylittlegeek.com` and is NOT blocked by an unrelated `FFC-EX-trendylittlegeek.com` sibling.

+4 test scenarios total (org-list failure fails safe in both workflows, Force bypass, hyphenated prefix strip); fake `gh` gains `TEST_ORG_REPOS_FAIL`. All 8 modules pass; `shellcheck -S info` clean on both extracted scripts; Prettier and catalog checks green.

---

clarkemoyer:
Addressed in 1673e15. The 702 preflight now checks `.archived` from the repo JSON it already fetched and **refuses archived targets** with unarchive guidance — deliberately with **no `force` override**, since an archived repo is read-only and the gated job would fail at push/PR time regardless (after burning a github-prod approval). New test pins that `force=true` does not bypass this refusal. All 8 modules pass; shellcheck/Prettier clean.

Nice synergy with this session's other work: repos retired via workflow 736 (archive-only) are now automatically protected from accidental re-clones by this check.

---

clarkemoyer:
**Inline comment — fixed in 82f0571.** Valid: a curl-level probe failure (000/empty) previously read as "not live" and let the dispatch proceed — the same fail-open class as the org-list case. The preflight now refuses when either live-site probe did not complete, with `force=true` as the explicit operator override. Two new tests pin the refusal and the bypass.

**On the two suppressed `github.token` comments — no change, by design:**
1. `GITHUB_TOKEN` reads **public** repo metadata and lists an org's **public** repos without issue; FFC charity-site repos are public by policy (`720` only grants tier teams on public repos, and Pages requires public on the free plan). Private `FFC-IN-*` repos are invisible to the scan, which is acceptable — they are internal/admin repos, never charity sites.
2. Switching to `CBM_TOKEN` is architecturally impossible here without defeating the feature: that PAT lives in the `github-prod` environment, so using it would drag the preflight **inside the approval gate** — the whole point is to fail *before* a human is asked to approve. Unauthenticated REST would work but has a 60-req/hr IP rate limit shared across runners; `GITHUB_TOKEN` (1,000/hr per repo) is the better budget.

If the sibling scan ever needs private-repo visibility, the right shape is a scheduled inventory (like 735's Dependabot repo list) consumed as data, not a PAT in the ungated job.

---

clarkemoyer:
Both addressed in c7df00b:

1. **Missing `Server` header no longer crashes the preflight** (the suppressed comment was right — worth surfacing): under `set -e`/`pipefail`, the failing `grep` killed the script mid-assignment, turning a valid non-cutover outcome into a silent hard failure. Now `|| true`d, and the `sed` strips any header-name casing. Pinned by `test_missing_server_header_is_noncutover_not_a_crash` (fake curl gained `TEST_APEX_NO_SERVER`).
2. **Fake `gh` glob tightened**: the org-repos case now matches only the listing endpoint (`orgs/<org>/repos` with query/flags/exact), so `orgs/<org>/teams/<team>/repos/...` keeps routing to the `*/teams*` branch and `TEST_TEAMS_FAIL` simulation stays intact — with a comment documenting the intent.

All 8 modules pass; shellcheck `-S info` and Prettier clean.

---

clarkemoyer:
Both addressed in f84677a (wording, by design no architecture change): the 702 "does not exist" error and the 720 "name is free" summary now state that the check covers repos visible to `GITHUB_TOKEN` and that GitHub 404s private repos it cannot see. Context kept in the messages: FFC-EX charity sites are public by policy (so for those, 404 genuinely means absent), and a private name collision still fails loudly at the gated create/clone step itself — the preflight's job is to catch the common case before an approval is consumed, and the caveat is now explicit for the rare one. Tests and lint remain green.

---

clarkemoyer:
🤖 **Conductor review (2026-07-18):** CI is fully green and the test coverage is strong, but there are **8 unresolved Copilot review threads** with real findings (hyphenated-domain parse in the 720 sibling scan, suppressed `gh api`/curl errors treated as safe, 404 = free/absent ambiguity for private repos, missing archived-repo check in 702 preflight, over-broad `orgs/*/repos` match in the test harness fake gh). Per AGENTS.md the merge queue won't accept the PR until these are fixed and resolved. Filed a backlog issue for an agent to address them; not queueing this run.

PR #839, opened 2026-07-25
docs(claude): two local-run env facts — UTF-8 on gh JSON output, and don't confirm a gate approval from the POST

clarkemoyer:
Refs #719

Two environment facts harvested in Conductor run 43, both of which cost a wasted call this run. Same shape as the read-after-write note in #836: **the error was in the confirmation, not in the action.**

### 1. `gh --format json` output must be read as UTF-8 explicitly

`json.load(open('board.json'))` died with `UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d` while reading back `gh project item-list 9 --format json`. Python defaults to cp1252 on the Windows Conductor box, and FFC issue titles, board item titles and PR bodies routinely contain em-dashes, arrows and smart quotes. Nothing was wrong with the data.

New short section, placed before the existing "Board & PR-creation env facts" block so it does not collide with #836's hunk in that list.

### 2. Do not confirm a gate approval from the POST's own output

The existing note said only *"the approval API returns an array of deployment objects; don't apply a `--jq` filter that assumes a single object."* That is not the whole trap — an **array-aware** filter fails too:

```
gh api -X POST .../pending_deployments -F "environment_ids[]=$ENV_ID" -f state=approved   --jq '.[0] | "\(.status) \(.environment.name)"'
→ expected an object but got: string ("github-prod")
```

…while the approval had **succeeded** and the run moved `waiting → in_progress`. The response shape simply does not match the `pending_deployments` GET. Reacting to that error means re-approving an already-approved gate. Rewrote the note to give the actual rule — drop the `--jq` on the POST and verify by re-reading `.../runs/$RUN_ID --jq '.status'` — with the observed output so a future session recognises it instead of re-deriving it.

### Notes

- Formatted with `npx --yes prettier@3.8.1` per this file's own guidance (the unpinned run produced identical output here, confirmed).
- Two follow-up style commits rather than a force-push, per the repo's no-force-push flow.
- **Draft, and left as one.** Per the Conductor's standing rule I open guidance/hook PRs and never merge them — @clarkemoyer reviews. Nothing here changes behaviour; it is CLAUDE.md prose only.
- Related pending prose PR: #836 (read-after-write consistency). These touch different regions of CLAUDE.md and should merge in either order without conflict; if #836 lands first this one still applies cleanly.


clarkemoyer:
Three Copilot threads addressed and resolved; the branch is synced to `main` (which now carries 744 from #979).

- **`json.load(open(...))`** → a `with` block. Fair point in a file whose entire subject is Windows traps: the snippet is meant to be copied, and an unclosed handle can hold a lock.
- **Unicode ellipsis in the `pending_deployments` POST** → spelled out in full, matching the working GET example directly above it. The snippet's whole claim is to be *the exact command*, so an un-copyable character defeated it.
- **`Same trap as the read-after-write note above`** → resolved by time rather than by rewording. Copilot was right when it commented: that note existed only in #836. It has since landed on `main`, and in the merged file the note is at `CLAUDE.md:472` with the reference at `:720`, so "above" is accurate. Left as-is deliberately.

Verified rather than assumed — checked both line numbers in the merged tree.

Issue #865, opened 2026-07-25
No per-charity traffic data — the traffic-ordered rollout has nothing to order by


clarkemoyer:
Child of #861.

## Problem

#861 says roll out **in order of traffic**. There is no data to order by.

Only **two** sites have GA reporting — `GA_PROPERTY_ID_PRIMARY` (freeforcharity.org) and
`GA_PROPERTY_ID_FFCADMIN`. There is no per-charity GA4 property registry, so for the other ~42
supported sites we cannot answer "which gets the most traffic?" at all.

Without it, "order by traffic" degrades to alphabetical — which is how the #841 backfill picked its
first batch, and it is why the first five happened to include a bespoke Vite app and an
under-construction placeholder rather than anything that matters.

## Work

1. **Inventory** which FFC-EX sites have a GA4 property (503 GTM provision and 505 GA4 property
   provision exist — establish what has actually been provisioned vs assumed).
2. **Registry** mapping domain → GA4 property id, published as data like the campaigns registry.
3. **Fleet traffic report** — a scheduled read-only rollup ranking supported sites by sessions.
4. **Feed the rollout order** so #861's children consume the ranking instead of guessing.

## Worth knowing before starting

Provisioning is likely uneven. Expect three populations: sites with GA4 wired, sites provisioned but
never verified, and sites with nothing. **Report those distinctly** — an unmeasured site must not
silently read as zero traffic, or the lowest-priority bucket fills with sites we simply cannot see.
That is the same failure class as #854 and #859.

## Acceptance criteria

- [ ] Every supported site classified: GA4 wired / provisioned-unverified / none
- [ ] Registry published as data and regenerated, not hand-maintained
- [ ] Ranking exists and is used to order #861's rollout
- [ ] "No data" is visibly distinct from "no traffic"

clarkemoyer:
**Claiming this** as the tracking issue for Phase 0 of the six-site convergence run (#861).

Scope I'm building now — one read-only hub workflow, `506. Google - Fleet Telemetry Reachability`:

1. **GTM** — enumerate containers under account `4702611686`, match to domains, flag shared/duplicate containers (`GTM-TQ5H8HPR` is currently shared by SPT, FOT and amargraves).
2. **GA4** — enumerate properties + web streams under `FFC Supported Sites` (`400204327`), matched on stream `defaultUri` — never property display name, which `docs/google-api.md` documents as unreliable.
[...]

Issue #861, opened 2026-07-25
Epic: prove the web standard on FFC first, then propagate to the fleet in order of traffic


clarkemoyer:
## Goal

A great template site and live Free For Charity sites with **measurable proof points**, then
propagate to the supported fleet **in order of traffic**.

**We prove it on FFC first.** `freeforcharity.org` and `ffcadmin.org` are ours — a mistake there
costs us, not a charity. Only once the standard is demonstrably right on FFC-owned properties, with
Search Console and Analytics showing it, does it reach charity sites.

## Why this epic exists

A security-audit backfill (#841) into five charity repos surfaced a chain of problems that had
nothing to do with security, and every one of them had been invisible:

- **`amargraves.org`** — cookie-consent banner linking to `/cookie-policy`, a page that did not
  exist. Live compliance gap, surfaced only as a Lighthouse 404.
- **`americanlegionpost64.org`** — `lighthouserc.json` copied from a template with a different build
  shape, asserting `.html` URLs a `trailingSlash: true` repo can never emit. Pages that **existed and
  worked** were reported missing.
- **The fleet has no URL standard.** 19 repos `trailingSlash: true`, 25 without, 27 not Next apps at
  all.
- **`Footer_Only_Template` — the one FFC template already running `trailingSlash: true` — has a live
  sitemap defect**: it advertises `/privacy-policy` while serving `/privacy-policy/`. Copying it
  would have propagated an SEO bug into the canonical template and from there into 44 sites.

That last one is the point of this epic. The fix was a one-line config change, and doing it by
copying the nearest FFC precedent would have multiplied a defect across the fleet.

## Principles

1. **Templates first, fleet second.** (#860) The templates are the rehearsal surface.
2. **FFC sites are the proof.** `freeforcharity.org` and `ffcadmin.org` demonstrate the standard
   works, with GSC/GA evidence, before any charity site changes.
3. **Order the rollout by traffic.** Highest-traffic sites first — most benefit, and the most
   sensitive to a mistake, so they get the standard only after it is proven.
4. **Measure, do not assume.** `freeforcharity.org` is the correct reference for URL shape;
   `Footer_Only_Template` is not, despite being the obvious template to copy. That was only knowable
   by probing the live sites.
5. **Converge the baseline, keep the bespoke.** (#859) Policy pages, consent and security apply
   everywhere; a charity's own content and structure are theirs.

## Children

[...]

Issue #854, opened 2026-07-25
726 drift audit reports a green check for org rulesets it cannot read (403/404 swallowed into a pass)


clarkemoyer:
## 726 reports "✓ Org-level branch ruleset present" while reading nothing — on every run

Found while testing credentials for #837/#848. This is live on `main` today.

### Evidence — today's real run

Run [`30157933873`](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30157933873)
(`success`, all steps green) printed:

```
- ✓ Org-level branch ruleset present ({"message":"Resource not accessible by personal access
  token","documentation_url":"...","status":"403"}0)
- ✓ Org-level branch ruleset present ({"message":"Not Found","documentation_url":"...","status":"404"}0)
```

A **green check**, for both orgs, containing an HTTP error.

### Mechanism

`.github/workflows/726-repo-rulesets-drift-audit.yml:88`:

```bash
org_rs=$(gh api "orgs/$org/rulesets" --jq '[.[] | select(.target=="branch")] | length' 2>/dev/null || echo "0")
```

Three things compound:

1. **`gh api` writes its error JSON to STDOUT, not stderr**, so `2>/dev/null` does not suppress it —
   the error body lands in `org_rs` via command substitution.
2. `|| echo "0"` then appends `0`, giving `{"message":...,"status":"403"}0`.
3. The downstream test is `if [ "$org_rs" = "0" ]` → **false**, because the string is not `"0"`. So
   the *missing-ruleset* branch never fires and the audit prints **✓ present**.

Had the fallback worked as intended (`org_rs="0"`), it would at least have reported ❌ missing. The
bug inverts that into a pass.

### Why the credential fails — and why more permissions will not fix it

Verified against the **production** token `wr-all-cbm-github-pat`:

| Call | Result |
| --- | --- |
| `GET /orgs/FreeForCharity/rulesets` | **403** Resource not accessible by personal access token |
| `GET /orgs/koenig-childhood-cancer-foundation/rulesets` | **404** Not Found |

- The **403** is not a missing grant. `Administration: read` **is** held — `GET
  /orgs/../actions/permissions` (same permission) returns 200. `GET /orgs/{org}/rulesets`
  specifically requires Administration **read *and write*** for fine-grained PATs.
- The **404** is structural: fine-grained PATs are scoped to a **single resource owner**, and 726
[...]

PR #867, opened 2026-07-25
feat(506): fleet telemetry reachability — GTM, GA4, Search Console + traffic ranking

clarkemoyer:
Phase 0 of the six-site convergence epic (#861); claims #865.

## What it answers

Four questions the fleet could not previously answer, in one read-only pass: which domains have a
GTM container (and is any **shared**), which have a GA4 property + web stream, which are **verified
in Search Console**, and **how much traffic** each gets — ranked.

## The two properties that matter more than the numbers

**1. It compares SERVED HTML against what the APIs report as provisioned.**

Provisioning records and reality disagree today, in both directions:

| | |
| --- | --- |
| Provisioned, not served | `amargraves` ships a `google-tag-manager/` component that `layout.tsx` **never imports** |
| Served, not provisioned | `GTM-TQ5H8HPR` is hardcoded and **shared** by SPT, FOT and amargraves, against one-container-per-charity |
| Served, but junk | ADMIN, SPT, FOT and amargraves fire `gtag('config','G-XXXXXXXXXX')` — a **placeholder id sent to Google** |

None of that is visible from provisioning records alone. So the cross-check is a first-class output,
not a nice-to-have.

**2. It never conflates "no data" with "no traffic".**

A site with no GA4 property is **UNMEASURED**; a site with a property and zero sessions has **NO
TRAFFIC**. Collapsing them sorts unmeasured sites to the bottom of a traffic-ordered rollout — which
is precisely how a site that matters gets treated as if it does not. #865 names this trap
explicitly, so the report keeps them in separate lists and the summary prints
`— no property (UNKNOWN, not zero)`.

## It refuses to under-report itself

Every probe records its own outcome; a failed probe emits a `::warning::` and is recorded rather
than yielding an empty result; and the run **exits non-zero if any probe failed**, so a partial
report cannot read as a clean one. That is the `726` lesson (#854) applied at write time instead of
discovered later.

## No new access needed

`analytics.readonly`, `tagmanager.readonly` and `webmasters` are **already** in the delegated scope
grant (`docs/google-api.md`), and `501` already proves GSC connectivity via its dispatch-only
`gsc-smoke` job. This adds the missing *consumer*, not new credentials — which is why **GSC can be
automated in the same chain as GA4 and GTM** rather than sequenced after.

Reuses `scripts/google-api-common.ps1` (`Get-GoogleDwdAccessToken`, `Invoke-GoogleApi`,
`Get-GoogleRows`) and the `google-secrets-from-kv` composite action at `scope: read`, exactly as
`501` does.

## Detail worth flagging for reviewers

GA4 properties are matched to domains by **web-stream `defaultUri`**, never by property display
name — `google-api.md` records display names as historically unreliable, and `505` already matches
the same way for idempotency.

Two bugs caught while wiring the shared helper: `Invoke-GoogleApi` serializes `$Body` itself, so
passing pre-JSON double-encodes it; and the GA Data API **omits `rows` entirely** for an empty range,
which is what `Get-GoogleRows` exists to handle. Both are used correctly here.

## Verification

PowerShell parses clean · catalog regenerates (96 workflows) · doc consistency 89 rows · workflow
reference guard passes · prettier clean.

**Not verifiable pre-merge** — `workflow_dispatch` needs the file on `main`. First check after merge
is a dispatch against the six primary sites; expect mismatches to be reported for amargraves (GTM
component not imported) and for the shared `GTM-TQ5H8HPR`.


clarkemoyer:
Taking two of the three, deferring one — with reasoning, since this PR has now been through five review-and-CI cycles.

**Taken — workflow-command sanitization.** A multi-line exception message in a `::warning::` would terminate the line and let the remainder parse as further workflow commands. CR/LF flattened before emit.

**Taken — asymmetric GA4 drift detection.** Correct and material: I checked *provisioned-but-not-served* and missed the more alarming direction, *served-but-not-provisioned* — traffic flowing to a property nobody is watching. Both directions now covered, matching what the GTM checks already did.

**Deferred — retry/backoff on 429/503.** A fair point for a weekly scheduled job, and I'd take it in isolation. Declining here because it is a robustness improvement rather than a correctness defect, and the failure mode is already safe: `Invoke-GooglePagedApi` **throws rather than truncating**, and any failed probe exits the run non-zero. So a rate-limit blip surfaces as a loud failure, not a quietly short report — which is the property that actually matters for this workflow. Filed as follow-up rather than grown into this change.

For the record, three earlier findings in this PR were real bugs and one of them was self-inflicted: I preserved URL paths so the Footer_Only_Template project URL would resolve, then lowercased the path in the same function — and Pages project paths are case-sensitive, so it would have 404'd a site that is up.

PR #881, opened 2026-07-26
fix(506): unblock the GA4 inventory — Test-HasProperty threw under StrictMode

clarkemoyer:
## The GA4 half of the telemetry report has been dead, and this is why

`506` returned real GTM and Search Console data while GA4 failed every run with:

> The property 'Name' cannot be found on this object. Verify that the property exists.

That is what blocks **traffic-ordered rollout** — the ranking cannot be produced without GA4.

## Root cause

Found by reproducing against the live API with the merged line-number diagnostics, not by reading.

`scripts/google-api-common.ps1:18` sets `Set-StrictMode -Version Latest` at module scope, and this
script inherits it by dot-sourcing. **Under StrictMode, enumerating `.Name` across an EMPTY
`PSMemberInfoCollection` throws.**

So `Test-HasProperty` — the guard written precisely to make property access safe — **was itself the
thing that threw**, and only in the empty case it existed to handle:

```powershell
return ($props.Properties.Name -contains $Name)   # throws when $props.Properties is empty
```

The GA4 Admin API returns `{}` for a property with no data streams. The first such property
(`properties/544445413`, `mitchellnchistory.org`) aborted the entire sweep — 46 properties under
*FFC Supported Sites* alone, all discarded because of one empty response.

## Fix

The indexer form: `$props.Properties[$Name]` returns `$null` for a missing name and never
enumerates. Verified safe under StrictMode for `{}`, `{"x":1}`, `$null`, arrays and scalars.

## Verified against the live API

GA4 enumeration now completes and matches properties to all four domains. The local run then stops
at the GA4 **Data** API for want of `GOOGLE_APPLICATION_CREDENTIALS`, which CI supplies — that is
the expected local boundary, not a further defect. I will dispatch `506` after merge for the full
traffic report.

## A note on the tests, because it nearly went wrong

Every behavioural case goes through a helper that sets StrictMode **immediately before** the call.
StrictMode is *dynamically scoped* and Pester resets it per `It` block, so a `Set-StrictMode` at the
top of the file has no effect inside tests.

**My first draft did exactly that and passed 9/9 against the broken implementation** — it certified
the bug instead of catching it. Caught by running the suite against the pre-fix code, which is worth
doing routinely; a regression test that has never failed is indistinguishable from a broken one.

As written: **3/10 fail against the old code, 10/10 pass against the new.**

Refs #865, epic #861

### DIFF
--- a/.github/workflows/722-ci.yml
+++ b/.github/workflows/722-ci.yml
@@ -264,15 +264,37 @@ jobs:
         shell: pwsh
         run: |
           $ErrorActionPreference = 'Stop'
-          if (-not (Test-Path scripts/tests)) {
-            Write-Host 'No scripts/tests directory; skipping Pester.'
+
+          # BOTH directories. This ran only scripts/tests, so every .Tests.ps1 under tests/
+          # was dead weight — present, passing locally, and never executed by any required
+          # check. tests/whmcs-application-triage.Tests.ps1 had been in that state for its
+          # whole life. A test that does not run is worse than a missing one: it reads as
+          # coverage in review, so nobody writes the test that would have run.
+          #
+          # Discovered by directory rather than listed, so a new location cannot be silently
+          # dropped again — and the count is printed so "0 files" is visible instead of
+          # looking like a clean pass.
+          $testDirs = @('scripts/tests', 'tests') | Where-Object { Test-Path $_ }
+
+          if (-not $testDirs) {
+            Write-Host 'No Pester test directories found; skipping.'
             exit 0
           }
+
+          $testFiles = Get-ChildItem -Path $testDirs -Filter *.Tests.ps1 -Recurse -File
+          Write-Host "Pester directories: $($testDirs -join ', ')"
+          Write-Host "Pester files found: $($testFiles.Count)"
+          $testFiles | ForEach-Object { Write-Host "  - $($_.FullName)" }
+
+          if ($testFiles.Count -eq 0) {
+            throw 'Pester test directories exist but contain no *.Tests.ps1 files — refusing to report a pass on zero tests.'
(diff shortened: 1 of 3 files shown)


clarkemoyer:
Both findings fixed. The first one was bigger than it looked.

## More than half the Pester suite was never running

```
scripts/tests  (CI ran this)        59 tests
tests          (CI did not)         66 tests
```

**66 of 125 PowerShell tests had never executed in a required check.** Not just the new file —
`tests/whmcs-application-triage.Tests.ps1` has been in that state for its entire life. Both passed
locally, both read as coverage in review, and neither could ever block a merge.

Fixed at the **directory** rather than by moving my file, which would have left the whmcs suite
silently dead. CI now:

- discovers test directories instead of naming one, so a new location cannot be dropped again
- prints the discovered file list and count
- **throws if a test directory exists but yields zero files** — "0 tests" can no longer look
  identical to a clean pass

Verified: 125 pass across both directories, so enabling the orphaned suite does not turn CI red.

## Third instance of one pattern

This PR chain has now produced three checks that reported success while not checking anything:

1. **PSScriptAnalyzer discarding `ParseError`** — a `.ps1` PowerShell cannot load passed the full
   suite green (fixed in #876)
2. **My own regression test passing 9/9 against the broken implementation** — Pester resets
   StrictMode per `It` block, so the tests certified the bug
3. **66 tests in an unwatched directory** — this one

Worth naming as a pattern rather than three coincidences. The common shape is a *green signal
derived from absence*: no findings, no throw, no files. Each looked exactly like success.

Second finding — the duplicated docstring paragraph — also fixed; two near-identical statements of
the same fact, which would have drifted.

PR #885, opened 2026-07-26
fix(google): report zero sessions as zero, not as a probe failure

clarkemoyer:
## Zero sessions was being reported as a probe failure

With #882 merged, traffic resolved for two more domains — SPT **10,219** sessions, ffcadmin
**2,700**. The last two then failed with a *new* error:

```
The property 'Count' cannot be found on this object.  [line 388: if ($rows.Count -gt 0)]
```

Two distinct StrictMode defects in `Get-GoogleRows`, both in **`google-api-common.ps1`** — shared
code that `501` and `502` also call:

**1.** `$Response.PSObject.Properties.Name -contains 'rows'` throws on a response with no properties
at all. Same defect as #881, in the shared helper rather than the caller.

**2.** `return @()` **does not return an empty array.** PowerShell unrolls it on the way out, so the
caller receives `$null` — and `$null.Count` then throws at the *call site*, naming a property the
reader never wrote:

```powershell
function f1 { return @()  }   # caller gets $null
function f2 { return ,@() }   # caller gets an array, Count=0
```

## Why this one mattered

The consequence is exactly what this script exists to prevent. Its own header:

> It never conflates "no data" with "no traffic". […] Collapsing the two sorts unmeasured sites to
> the bottom of a traffic-ordered rollout.

A property with genuinely **zero** sessions came back as a probe **failure** rather than as zero.
`technologymonastery.org` and `amargraves.org` were being reported as unmeasured when the real
answer is that nobody is visiting them — a very different fact, leading to a different decision.

## Verification

Every response shape: `{}`, no `rows` field, `rows:[]`, one row, many rows.

7 test cases, **failing 5/7 against the broken implementation**.

One assertion in my first draft used `Should -Not -BeNullOrEmpty` to prove "not `$null`" — which
**fails against the correct implementation**, because an empty array is legitimately empty. It now
asserts on `.GetType().IsArray`. Same lesson as #881: run new tests against the pre-fix code *and*
confirm they pass against the post-fix code, because a test can be wrong in either direction.

132 tests pass across both directories.

Refs #865, epic #861

### DIFF
--- a/scripts/google-api-common.ps1
+++ b/scripts/google-api-common.ps1
@@ -135,10 +135,26 @@ function Get-GoogleRows {
     <#
     Safely return a report response's rows as an array. The GA Data API omits the 'rows' field
     entirely when a property/date range has no data, so $response.rows throws under Set-StrictMode.
+
+    Two StrictMode hazards are handled here, and BOTH bit in production:
+
+    1. `.PSObject.Properties.Name -contains` throws on a response with NO properties at all
+       (a bare {}), because enumerating .Name across an empty PSMemberInfoCollection is an error
+       under StrictMode. The indexer returns $null instead and never enumerates.
+
+    2. `return @()` does NOT return an empty array. PowerShell unrolls it on the way out, so the
+       caller receives $null — and `$null.Count` then throws under StrictMode, at the call site,
+       naming a property the reader never wrote. `return ,@()` wraps it so the unroll yields the
+       empty array intact. This is why a property with genuinely zero sessions failed with
+       "The property 'Count' cannot be found on this object" instead of reporting zero.
   #>
     param([Parameter(Mandatory)]$Response)
-    if (($Response.PSObject.Properties.Name -contains 'rows') -and $Response.rows) { return @($Response.rows) }
-    return @()
+    # No $null check: [Parameter(Mandatory)] already rejects $null at binding time, and a null
+    # response IS a caller bug worth failing loudly on rather than quietly reporting "no rows".
+    $props = $Response.PSObject
+    if ($null -eq $props) { return , @() }
+    if (($null -ne $props.Properties['rows']) -and $Response.rows) { return , @($Response.rows) }
+    return , @()
 }
 
 function Get-GoogleDwdAccessToken {
--- a/scripts/google-telemetry-reachability.ps1
+++ b/scripts/google-telemetry-reachability.ps1
@@ -383,7 +383,12 @@ function Get-Ga4Sessions {
     $resp = Invoke-GoogleApi -Uri "https://analyticsdata.googleapis.com/v1beta/${PropertyName}:runReport" -AccessToken $tok -Method Post -Body $body
     # Get-GoogleRows exists because the Data API OMITS 'rows' entirely for an empty range —
     # reaching for $resp.rows directly throws under Set-StrictMode.
-    $rows = Get-GoogleRows -Response $resp
+    # Belt and braces: Get-GoogleRows now returns a real empty array (see its docstring), but a
+    # caller that trusts a helper's return shape is one refactor away from $null.Count throwing
+    # again — which is how a site with zero sessions came back as a probe FAILURE rather than as
+    # zero, and an unmeasured site is treated very differently from a quiet one in a rollout
+    # ordered by traffic.
+    $rows = @(Get-GoogleRows -Response $resp | Where-Object { $null -ne $_ })
     $sessions = 0; $users = 0
     if ($rows.Count -gt 0) {
         $sessions = [int]$rows[0].metricValues[0].value
--- a/tests/google-api-common.Tests.ps1
+++ b/tests/google-api-common.Tests.ps1
@@ -1,0 +1,97 @@
+# Regression tests for Get-GoogleRows in scripts/google-api-common.ps1.
+#
+# As in google-telemetry-reachability.Tests.ps1: every call goes through a helper that sets
+# `Set-StrictMode -Version Latest` immediately before invoking. StrictMode is dynamically
+# scoped and Pester resets it per It block, so setting it at file top does nothing inside the
+# tests — and without it, every case here passes against the broken implementation.
+#
+# Two defects are covered, both of which reached production:
+#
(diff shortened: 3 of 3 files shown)


clarkemoyer:
## Superseded by #898 — same defect, and the two branches conflict

Flagging this from a landing run before either reaches the queue, because both are open and both rewrite `Get-GoogleRows`.

**Test-merge evidence** (`git merge-tree --write-tree` against `main` @ `0a5b2a4`):

```
#898 + #885   CONFLICT (content): scripts/google-api-common.ps1
              CONFLICT (content): scripts/google-telemetry-reachability.ps1
#898 + #897   CLEAN
#898 + #837   CLEAN
```

Neither fix is on `main` yet — it still carries `return @()`, so whichever lands first makes the other conflict.

**The two fixes are semantically equivalent** where they overlap: indexer instead of the enumerating `.Properties.Name -contains`, unary comma on every return path.

| | #885 | #898 |
| --- | --- | --- |
| `Get-GoogleRows` (`google-api-common.ps1`) | ✅ | ✅ |
| `google-telemetry-reachability.ps1` helpers | 1 (`Get-GscInventory`) | 4 (`Invoke-GooglePagedApi`, `Get-GtmInventory`, `Get-Ga4Inventory`, `Get-GscInventory`) |
| Tests | new `tests/google-api-common.Tests.ps1`, 7 cases | +15 cases in `tests/google-telemetry-reachability.Tests.ps1` (9 → 24) |
| Ledger row | — | L29 |

#898 is the broader change: it also covers the latent `$raw.Count` crash in `Get-GtmInventory`, which has never fired only because the GTM/GA4 Admin list endpoints have always returned something.

**Two things in this PR that #898 does not have**, worth a look before closing rather than assuming subset:

1. A `$props = $Response.PSObject; if ($null -eq $props) { return , @() }` guard. #898 relies on `-Response` being `Mandatory` to reject `$null` and pins that behaviour with a test; `$Response.PSObject` is not itself a reachable `$null` for a bound non-null argument, so this looks like defence rather than coverage — but it is a deliberate difference, not an oversight.
2. `tests/google-api-common.Tests.ps1` as a **separate file**. #898 tests `Get-GoogleRows` from the telemetry test file via AST extraction. If a standalone test file for the shared Google helpers is wanted on its own merits, that part is worth salvaging even though the assertions duplicate #898's.

**Recommendation:** land #898 and close this as superseded, or rebase this down to just the standalone test file. Not closing it here — that is @clarkemoyer's / the Conductor's call, and the second column above is a real difference rather than pure duplication.

LESSON: two sessions independently fixed the same `return @()` StrictMode defect three days apart (#885 on 07-26, #898 on 07-28), both citing #865/#875. The `claimed` label tells you a PR exists for an issue; it does not tell a *second* issue's worker that the fix it is about to write already exists on another branch. The supersession grep in `AGENTS.md` targets `main` — neither of these was on `main`, so the grep would have passed for both. The gap is open-PR-versus-open-PR, and the cheap detector is `git merge-tree` across sibling agentic-os branches before promoting, which is what caught it here.

---

clarkemoyer:
## Superseded by #898 — same defect, wider fix, and the two branches conflict

Flagging this from an agentic-os landing run before it reaches the queue. #898 fixes the same `Get-GoogleRows` StrictMode defect this PR does. Neither PR references the other: this one was authored against #865, #898 against #875 (740's rolling 506 failure alert), and both diagnosed the identical bug independently — the analysis in the two descriptions matches almost line for line, down to the `return @()` / `return ,@()` contrast.

**They conflict.** `git merge-tree` against the merge base reports `changed in both` on `scripts/google-api-common.ps1` and `scripts/google-telemetry-reachability.ps1`.

#898 is a strict superset: it carries both of this PR's fixes to `Get-GoogleRows` (indexer + unary comma) and additionally fixes the same latent shape in `Invoke-GooglePagedApi`, `Get-GtmInventory`, `Get-Ga4Inventory` and `Get-GscInventory`, adds ledger row **L29**, and pins the remedy with a source guard over an exact list of all five array-returning helpers. Full function-by-function comparison in [my note on #898](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/898#issuecomment-5104022691).

Two things here are not in #898, and neither is a coverage gap:

- The `$null -eq $props` guard is unreachable — `-Response` is `Mandatory`, so `$null` never binds and `.PSObject` is never `$null` on a bound value.
- The `| Where-Object { $null -ne $_ }` wrapper in `Get-Ga4Sessions` is redundant once the helper is correct, and it only protects one of the two call sites — `google-analytics-report.ps1:92` still reads `.Count` bare, so it would hide a regression rather than catch one.

**Suggested disposition:** close this as superseded by #898 and re-point #865 at #898, so the traffic-ordered rollout still tracks the dependency. I have not closed it — that is a human call, and this PR is outside an agentic-os worker's mandate.

If you would rather land this one instead, the reverse is not equivalent: #898 would then need rebasing and the four extra helper fixes plus the ledger row would still be outstanding.

---

clarkemoyer:
**Superseded — closing (Conductor run 44).**

The same defect was fixed and merged on `main` in d4b0fe4 (#898, "an empty GA4 result is zero traffic, not a failed probe"), which landed after this branch was cut. This PR is `CONFLICTING/DIRTY` against that fix.

Verified nothing is lost by closing:

- **The fix**: `main`'s `scripts/google-api-common.ps1` now carries `Get-GoogleRows`, which returns `, @()` on a missing-or-empty `rows` field and deliberately keeps `-Response` mandatory so a 200 with an empty body stays an unexpected response rather than being reported as measured zero traffic. That is the same semantics this PR aimed at, with the unary-comma-on-every-return-path subtlety documented in place.
- **The coverage**: this PR's new `tests/google-api-common.Tests.ps1` is superseded too — `main`'s `tests/google-telemetry-reachability.Tests.ps1` extracts `Get-GoogleRows` from `google-api-common.ps1` and exercises it directly under `Set-StrictMode` (see its `Invoke-GoogleRowsUnderStrictMode` / `Measure-GoogleRowsUnderStrictMode` harnesses). The helper is unit-tested on `main` today.

No rebase needed — reopen only if you find a case `main`'s version misses.

PR #897, opened 2026-07-28
fix(security): issue-body values reached pwsh as script text in 701 — move untrusted input to env, and guard the class

### LINKED ISSUE #896: Security: `::stop-commands::` annotation variant (the expression-injection half is fixed in #897)
> **Status:** the severe half of this issue is **fixed and merged** — #897 landed as `1e88665`, and `main` no longer interpolates untrusted values into any `run:` script. What remains open is the weaker `::stop-commands::` variant at the bottom. See [the closing comment](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/issues/896#issuecomment-5117805922) for the post-merge verification and the one thing still unmeasured.

Filed from a cloud-worker run. #893's closing comment flagged an untracked annotation-injection class (`::stop-commands::`) and pointed at a sweep of `.github/`. The sweep found a **more direct and more severe member of the same family** — untrusted values reaching the shell as *script text* rather than as data — so this issue tracked that first, and the `::stop-commands::` variant is folded in below as the weaker sibling it turned out to be.

## The defect (fixed in #897)

`701. Website - Provision` is triggered by `issues: [assigned]`. This repository is **public** and has issues enabled, so the body of a `[WEBSITE REQUEST]` issue is authored by whoever opens it; assignment — the documented intake step — is what fires the run.

Its `resolve` job parses ~20 fields out of that body. The only processing is `normalizeInput`, which trims and maps the `_No response_` sentinel to empty:

```js
function normalizeInput(value) {
  const v = String(value || '').trim();
  if (!v) return '';
  if (/^_?no response_?$/i.test(v)) return '';
  return v;
}
```

The `repo` and `content` jobs then interpolated those fields into `pwsh` string literals:

```yaml
$charityName = "${{ needs.resolve.outputs.charity_name }}"
```

**The runner substitutes an expression into the script TEXT and only then hands that text to the shell.** So the value is not data — a quote in it ends the literal, and what follows is parsed as PowerShell. Rendered:

```
issue field  ->  Acme Charity"; <command>; "
rendered     ->  $charityName = "Acme Charity"; <command>; ""
```

Both affected steps hold the org-wide write PAT (`wr-all-cbm-ffc-copilot-mcp-github-pat`) in `$env:GH_TOKEN` at the time — the credential #848 finding 3 describes as the 100+-repo create/archive/collaborator credential.

### The JSON fields hid it best

`footer_address_json`, `footer_social_json` and `leadership_json` looked safer because they are `JSON.stringify`-ed. But `JSON.stringify` escapes `"` and **not** `'`, and they were interpolated into *single*-quoted literals:

```
$footerAddress = '{"line1":"12 O'Brien Way"}' | ConvertFrom-Json
```

The literal ends at `O`. An apostrophe was sufficient — which also makes this a plain correctness bug for any charity whose address or board member's name contains one.

### Scope, measured

Every `run:` step in every workflow reachable from an untrusted event was swept:

| Workflow | Trigger | Sites before | After |
| --- | --- | --- | --- |
| `701-website-provision.yml` | `issues: [assigned]`, `repository_dispatch` | 22 | 0 |
| `105-manage-record.yml` | `issues: [labeled]` | 2 | 0 |
| `113-cloudflare-domain-register.yml` | `issues: [labeled]` | 0 | 0 |
| `206-whmcs-issue-to-ticket.yml` | `issues` | 0 | 0 |

105's two sites are lower severity — its issue path constrains the domain to `[A-Za-z0-9.-]`, though its `workflow_dispatch` path does not — but they are the same defect and were fixed alongside.

Composite actions and `workflow_call` workflows were checked for the second hop (an untrusted value passed in as an action input and interpolated into that action's own `run:`). **Zero sites**; that path was already clean.

## Why prose was not enough

The rule was already written down in this repo — on **one** step of 105:

> Untrusted/user-supplied values are passed via env (never interpolated directly into the script body) so they cannot break out of the string and inject PowerShell.

…while the two sibling steps in the same file did it the unsafe way. A correct rule sitting next to one of three steps is not a guard. Hence a test.

## The guard

`tests/workflow-logic/test_no_untrusted_expression_in_run.py` — in a workflow reachable from an untrusted event (`issues`, `issue_comment`, `repository_dispatch`, `pull_request_target`, `discussion*`, `fork`, `watch`), a `run:` block contains **no `${{ }}` at all**.

Absolute, rather than a denylist of taint carriers (`needs.*`, `github.event.*`, `steps.*.outputs.*`) or an allowlist of "safe" contexts. A denylist has to anticipate every carrier and goes silent on the one nobody listed; an allowlist turns `env.NAME` into a taint analysis. Absolute is affordable because the remedy is one line and the tree has **zero** exceptions — a rule with no exception list cannot rot into a stale one.

Both directions are covered, and it is mutation-tested: reverting either fix fires it with the file, job, step and remedy named.

## Acceptance criteria

- [x] No workflow reachable from an untrusted event interpolates any value into a `run:` script
- [x] CI guard fails if one is reintroduced, and is proven against the real pre-fix line
- [x] Composite actions / `workflow_call` second hop swept
- [x] Ledger row (L28)
- [x] **PR #897 merged** — `1e88665`, verified against `origin/main`
- [ ] `701` rehearsed with `dry_run=true` after #858 lands — the only acceptance test a sandbox could not run
- [ ] `::stop-commands::` variant (below) split out and closed

## What is still open: the `::stop-commands::` annotations

#893's original note, and the remaining scope of this issue.

`grep -nE '::(error|warning|notice)::.*\$'` over `.github/` returns **104** lines that interpolate a value into a workflow-command annotation. A value carrying a newline plus `::stop-commands::<token>` silences every subsequent workflow command in the job — including `::add-mask::`, which is what makes it a secret-exposure risk rather than just a reporting one.

Assessed and **deliberately not fixed in #897**, so the reasoning is recorded rather than re-derived:

- After that fix, no annotation in an untrusted-triggered workflow interpolates event data — the remaining 104 carry API error bodies, repo names, internal counters and curl failure text.
- The realistic multi-line attacker-controlled paths into those are weak, and each `::add-mask::` in this repo runs *before* the untrusted text is emitted, not after.

Worth closing anyway, but as its own sized piece of work rather than bundled into a security fix that needed to merge quickly. **Splitting it out is the ask.**

## Notes

Found and fixed from a cloud sandbox — no gates hit, repo content only. No live exploitation was attempted; the rendering above is a textual substitution performed locally, and no ready-to-run payload was published while `main` was still affected.

clarkemoyer:
Fixes #896. Found while sweeping the annotation-injection class #893's closing comment flagged as untracked — the sweep turned up a more direct member of the same family first.

## The defect

`701` is triggered by `issues: [assigned]`, on a **public** repo with issues enabled. Its `resolve` job parses ~20 fields out of the issue body through `normalizeInput` — which trims and maps `_No response_` to empty, and does nothing else — and the `repo` and `content` jobs interpolated them into `pwsh` string literals.

The runner substitutes an expression into the script **text** and only then hands it to the shell, so the value is code, not data:

```
issue field  ->  Acme Charity"; <command>; "
rendered     ->  $charityName = "Acme Charity"; <command>; ""
```

Both steps hold the org-wide write PAT in `$env:GH_TOKEN` at that moment.

The three `JSON.stringify`-ed fields looked safest and were the worst: that escapes `"` but not `'`, and they sat in **single**-quoted literals —

```
$footerAddress = '{"line1":"12 O'Brien Way"}' | ConvertFrom-Json
```

— so an apostrophe was enough. That one is also a plain correctness bug for any charity whose address or board member's name has one.

## What changed

| File | Sites before | After |
| --- | --- | --- |
| `701-website-provision.yml` | 22 | 0 |
| `105-manage-record.yml` | 2 | 0 |

Every value now travels by `env:` and is read as `[string]$env:NAME`. The `[string]` cast is load-bearing on Windows: an env var set to the empty string is not created, so bare `$env:X` would be `$null` where `"${{ … }}"` used to yield `''`, and several call sites compare against `''`.

105 was not an afterthought — its "Manage DNS Record" step already carried a comment stating this exact rule while its two sibling steps did it the unsafe way. A correct rule sitting next to one of three steps is precisely why this lands as a test.

One incidental: `Organization = "${{ env.WEBSITE_TARGET_ORG }}"` → `[string]$env:WEBSITE_TARGET_ORG`. A workflow-level literal, never a risk — converted so the guard below needs no allowlist at all.

## The guard

`tests/workflow-logic/test_no_untrusted_expression_in_run.py`: in a workflow reachable from an untrusted event, a `run:` block contains **no `${{ }}`**.

Absolute, not a denylist of taint carriers and not an allowlist of safe contexts. A denylist has to anticipate every carrier and stays quiet on the one nobody listed; an allowlist makes `env.NAME` only as safe as NAME's definition, which is a taint analysis rather than a rule. Absolute is affordable because the remedy is one line, it is what these workflows already do everywhere else, and the tree now has **zero** exceptions — there is no exception list to go stale.

Five tests, in both directions:

- `..._no_untrusted_workflow_interpolates_into_a_run_script` — the rule.
- `..._the_untrusted_trigger_set_is_not_empty` — guards the sweep's *input*. `on:` parses as the YAML 1.1 boolean `True`; breaking that lookup would empty the untrusted set and leave the rule passing while reading nothing.
- `..._the_scan_can_actually_see_the_defect` — run against the **real pre-fix line** from 701, not a stand-in, and asserts the message names the remedy.
- `..._the_scan_leaves_correct_code_alone` — five samples, three of which are the *remedy itself* (`env:`, `with:`, `if:`). A guard that flagged those would have to be disabled to land the fix for the thing it guards. Per L27's review, four of nine findings there were a guard misfiring on correct code.
- `..._the_dispatch_only_carve_out_is_not_a_hole` — the one place this rule chooses not to look, pinned: the same script plus an `issues:` trigger must fail.

Mutation-tested: reverting 701's `charity_name` line fires it naming file/job/step/remedy; reverting 105's `Validate Zone` line likewise; restoring clears both.

## Also swept, and clean

Composite actions and `workflow_call` workflows were checked for the second hop — an untrusted value passed in as an input and interpolated into *that* file's `run:`. **Zero sites.** Worth stating, because that is the path a workflow-only guard would miss, and it is the shape that made an earlier guard in this repo inspect 15 of 71 workflows and pass (L17).

## Not in scope, deliberately

The `::stop-commands::` annotation variant #893 originally named. 104 interpolating annotation lines remain; after this fix none of them is in an untrusted-triggered workflow, and every `::add-mask::` here runs before the untrusted text is emitted rather than after. Reasoning and the ask to split it out are recorded on #896 rather than bundled into a fix that should merge quickly.

## Verification

- `python3 tests/workflow-logic/run_all.py` — **33/33 modules** (32 before this PR, +1 new)
- `actionlint` clean, repo-wide
- `prettier@3.8.1 --check .` clean (CI-pinned; per `CLAUDE.md`)
- `check-workflow-doc-consistency` 98/91 · catalog up to date, no drift · `check-workflow-references` · `check-federated-credential-subjects` · Pages-IP · Cloudflare bulk-lib — all pass

The harness runs cleanly in this sandbox (Linux), so no false-red to discount here.

**Not verifiable pre-merge:** that 701 still provisions end to end. It cannot be exercised from a sandbox — the `dns` and `repo` jobs sit at `cloudflare-prod-write` / `github-prod` gates a cloud worker cannot approve. The change is mechanical and every call site's type is preserved by the `[string]` cast, but a `dry_run` rehearsal after merge (see #858, which adds one to 701) is the honest acceptance test.

Ledger row **L28**. Draft, per the standing rule that @clarkemoyer promotes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BnWSj8KJS9dTXzPYqwXxQZ)_


clarkemoyer:
## Landing check: green, zero threads, and the whole `agentic-os` stack is order-free

Landing run (agentic-os cloud worker). Rule 1 fired — 3 open `agentic-os` PRs, so this run went to landing them rather than starting new work.

**This PR at `33f09c0`:** `mergeable_state: clean`, **0 commits behind `main`** (`main` still at `0a5b2a4`), and all five checks green — 722 Validate Repository ✅ · 727 Phantom Revert Guard ✅ · 723 CodeQL ✅ · Claim sync ✅ · Copilot Code Review ✅. Copilot reviewed 4/4 changed files and **generated no comments**, so there are no review threads to resolve. Nothing on the branch blocks the queue.

### New: the three-way stack was verified, not just the pairs

The prior landing run on #898 checked each pair against #898 individually. I built the actual combined tree — #897 + #837 + #898 merged in sequence — because #897 adds an **absolute** guard (no `${{ }}` in any `run:` of an untrusted-triggered workflow) while #837 rewrites three workflow files, and a pairwise-clean result does not cover a rule meeting a diff it never saw.

```
897 + 837          merge clean
897 + 837 + 898    merge clean
```

Against the fully merged tree:

| Check | Result |
| --- | --- |
| `tests/workflow-logic/run_all.py` | **34/34 modules** — including all 5 of this PR's new guard's tests |
| `check-workflow-doc-consistency.py` | 98 workflows / 91 rows ✅ |
| `check-workflow-references.py` | 98 workflows, all references resolve ✅ |
| catalog regeneration | no drift ✅ |
| ledger ids | L28 (this PR) and L29 (#898) coexist, `test_ids_are_unique` passes |

So **#837's workflow edits introduce no site that #897's new guard flags**, and the L28/L29 ledger rows do not collide. Ordering among the three open `agentic-os` PRs is free — any sequence merges.

### What each one is actually waiting on

All three are green, clean, and mutually compatible. Every remaining blocker is human-only:

| PR | Blocker |
| --- | --- |
| **#897** (this) | promotion only |
| **#898** | promotion + the #885 duplicate decision (same `Get-GoogleRows` fix, branches conflict) |
| **#837** | promotion + the 4-step Azure/GitHub provisioning (federated credential, KV grant, `github-prod-read` environment with its two OIDC secrets), plus #877's 401 on `read-all-cbm-github-pat` |

Left as a draft — the standing rule is that @clarkemoyer promotes, and a cloud worker cannot approve the environment gates #837 needs regardless.

LESSON: The "3+ open agentic-os PRs → land them instead of starting new work" rule can deadlock. All three open PRs are green with zero CI or review work left, and each one's only remaining blocker is an action a cloud worker structurally cannot take (promoting a draft against the documented promote-by-human convention, deciding a duplicate's fate, provisioning Azure identity). A landing run then has no forward move and the backlog stays frozen behind the queue. Worth a Conductor rule: when every open PR is blocked on a human, the worker should be released to pick up new issues rather than re-verifying a stack that is already verified — or the promote-by-human convention needs an explicit escape hatch.

---

Issue #883, opened 2026-07-26
First working telemetry sweep: findings, and why the GA4 mismatch check is unreliable


clarkemoyer:
## First working telemetry sweep — findings, and one caveat that matters

`506` run [30204750964](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30204750964),
the first with GA4 enumeration working (#881).

| Domain | GTM | GA4 | GSC | Sessions (28d) |
| --- | --- | --- | --- | --- |
| freeforcharity.org | ✅ GTM-NJ4DXH9 | ❌ **none** | ✅ | unmeasured |
| ffcadmin.org | ❌ none provisioned | ✅ G-7VK191LDRL | ✅ | **2689** |
| ffcworkingsite1.org (SPT) | ✅ GTM-TQ5H8HPR | ✅ G-312DE64R9S | ❌ | 403 (see #882) |
| FOT (github.io) | ❌ | ❌ | ❌ | unmeasured |
| technologymonastery.org | ✅ GTM-MM2SXD6R | ✅ G-35T2Q27S8L | ❌ | 403 (see #882) |
| amargraves.org | ✅ GTM-P42GJR2X | ✅ G-YP6688193Y | ❌ | 403 (see #882) |

## Read this before acting on the mismatch list

The report flags nine "provisioned vs served" disagreements. **A large share of them are false
positives, and ffcadmin proves it.**

ffcadmin's provisioned GA4 `G-7VK191LDRL` is **not present in its served HTML** — and it recorded
**2689 sessions**. The tag is deployed *inside the GTM container*, which is the recommended setup and
is invisible to an HTML scan.

So **"GA4 ID absent from served HTML" is not evidence of missing measurement.** The check cannot
distinguish a GA4 tag fired through GTM from no measurement at all, and every GA4-side mismatch in
that list is therefore unreliable. The only reliable signal is **sessions > 0**, which is exactly
what #882 unblocks.

The **GTM-side** mismatches are sound — a container ID either appears in the served HTML or does not.

## What the sound findings say

- **`freeforcharity.org` has no GA4 property at all.** The flagship is unmeasured, and its
  provisioned container `GTM-NJ4DXH9` is not in the served HTML either. The proof point we intend to
  publish is currently the least instrumented site in the set.
- **`technologymonastery.org` and `amargraves.org` have their own containers provisioned
[...]

Issue #843, opened 2026-07-25
Alerting layer is dead: workflow_run has never fired in this repo — convert 740 to a poll and retire 732


clarkemoyer:
## Problem

**`workflow_run` has never fired in this repository, so the entire failure-alerting layer is dead.**
Full evidence in FreeForCharity/FFC-Cloudflare-Automation#832 (comment of 2026-07-25 ~11:2xZ). Summary:

```bash
gh api "repos/FreeForCharity/FFC-Cloudflare-Automation/actions/runs?event=workflow_run&per_page=1" --jq .total_count
# 0
```

| Repo | `workflow_run` runs, all time |
| --- | --- |
| **FFC-Cloudflare-Automation** | **0** |
| FFC-IN-ffcadmin.org | 995 |
| FFC-IN-freeforcharity.org | 162 |

Both `732` (Google, watching 501/502) and the brand-new `740` (watching 11 workflows) are `active`
and have **never produced a single run**. Ruled out with evidence: name mismatch (all 13 watched
names byte-exact), non-default branch, workflow disabled, Actions policy, fork/archived, and
registration lag. The platform-side cause is being escalated separately — **this issue does not wait
on it.**

## Scope (one PR)

Convert **`.github/workflows/740-scheduled-workflow-failure-alert.yml`** from a `workflow_run`
watcher to a **poll**, and retire `732` into it.

**Keep the existing decision logic exactly as it is** — the gate discriminator, the benign-conclusion
skip, the marker-keyed upsert/close, and the labels are all correct and well covered by
`tests/workflow-logic/test_740_scheduled_failure_alert.py`. **Only the trigger and how a run is
obtained change.** Do not rewrite the alerting behaviour.

### Trigger

```yaml
on:
  schedule:
    - cron: '*/30 * * * *'   # pick a free slot; see the stagger note below
  workflow_dispatch:
```

`workflow_dispatch` is **required, not optional** — it is what makes this alerter *provable*.
`workflow_run` could never be tested on demand, which is the reason a dead alerter went unnoticed for
six days.

### Run enumeration

For each watched workflow name, take its **latest completed run on the default branch**:

- resolve names → workflow ids from `GET /repos/{owner}/{repo}/actions/workflows` — **use
  `--paginate` / follow pagination.** This repo has **105** workflows and `per_page=100` silently
  returns 100. That truncation already produced one false "740 is not registered" conclusion during
[...]

PR #898, opened 2026-07-28
fix(506): an empty GA4 result is zero traffic, not a failed probe

clarkemoyer:
Refs #875. Also the first thing 740 caught after #843 converted it to a poll — #875 is 740's debut alert, and it has been red on every scheduled run since (runs [3](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30179190317), [5](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30205111284), [6](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30250547152)) with nobody picking it up.

## The defect

Two StrictMode traps in one helper, `Get-GoogleRows`. Both reproduced under pwsh 7.5.4 in this sandbox before any edit:

```
return @()   ->  $null                        # PowerShell unrolls a collection on output,
                                              # and unrolling an empty one emits nothing
$null.Count  ->  PropertyNotFoundException    # under Set-StrictMode -Version Latest
```

That is the crash in the log, verbatim:

```
##[warning]GA4 sessions for technologymonastery.org failed — The property 'Count' cannot be found
on this object. [PropertyNotFoundException at line 388: if ($rows.Count -gt 0) {]
```

The GA4 Data API **omits `rows` entirely** for a range with no data, which is documented and correct. `Get-GoogleRows` handled that by returning `@()` — which never reaches the caller as an empty array. So a property with genuinely **zero sessions** did not read as `0`; it threw, and landed in the report as `— query failed (UNKNOWN)`. Two charity domains hit it every run, and the step exits non-zero when any probe fails.

**Zero is a measurement.** A traffic report that cannot express it reports an outage instead — and #865 wants to order a rollout by traffic, so a measured-zero site silently becomes an unranked one.

| Response shape | Before | After |
| --- | --- | --- |
| `rows` omitted (no data) | `$null` → `.Count` **throws** | `Object[]`, count 0 |
| `{"rows":[]}` | `$null` → `.Count` **throws** | `Object[]`, count 0 |
| `{}` (empty body object) | **throws** in the presence guard | `Object[]`, count 0 |
| one row | bare `PSCustomObject` | `Object[]`, count 1 |
| two rows | `Object[]` | `Object[]`, count 2 |

**Second trap, same function.** `.PSObject.Properties.Name -contains` *enumerates*, and enumerating an empty `PSMemberInfoCollection` throws under StrictMode — so the guard written to make property access safe was itself the thing that threw, for exactly the `{}` body it existed to handle. That is the identical finding already fixed in `Test-HasProperty` one file over, whose docstring explains it at length; this copy simply never got it. Now uses the indexer.

**Third, the one-row row of the table.** Pre-fix, a single-row response returned a bare `PSCustomObject`, and the `$rows[0]` on the next line worked only by PowerShell's scalar-indexing courtesy. It was never an array.

## Blast radius beyond 506

`scripts/google-analytics-report.ps1:92` reads `if ($respRows.Count)` off the same helper, so **502 carried the identical latent crash** — any GA4 property with an empty window would have taken it down the same way. One helper fix covers both.

`Invoke-GooglePagedApi` in the telemetry script has the same shape and `Get-GtmInventory` reads `$raw.Count` off it. It has never crashed only because the GTM and GA4 Admin list endpoints have always returned something; a GTM account with no containers is all it takes. Fixed and pinned rather than left as a second incident.

## What changed

| File | Change |
| --- | --- |
| `scripts/google-api-common.ps1` | `Get-GoogleRows` — unary comma on both returns, indexer instead of the enumerating presence check |
| `scripts/google-telemetry-reachability.ps1` | unary comma on `Invoke-GooglePagedApi`, `Get-GtmInventory`, `Get-Ga4Inventory`, `Get-GscInventory` |
| `tests/google-telemetry-reachability.Tests.ps1` | 9 tests → 24 |
| `docs/lessons-ledger.md` | row **L29** |

No workflow file changed, so no catalog or safety-doc row moves.

### A null response is still rejected, deliberately

`-Response` is `Mandatory` and therefore refuses `$null`. A 200 with an empty body deserializes to `$null`, and that is an *unexpected response*, not evidence of zero traffic — returning an empty array there would fabricate a measurement, reporting "0 sessions" for a property nobody successfully queried. I wrote that forgiving branch first; a test proved it unreachable, and the test now pins the behaviour shut so the natural `[AllowNull()]` "simplification" fails CI.

## Tests

**Every test reads `.Count` the way the production call sites do.** That is the whole discipline here: the property access *is* the failure, so a test that merely calls the function and checks its value passes against the broken implementation. Same trap the file's existing header warns about for StrictMode's dynamic scoping — a first draft of those tests passed 9/9 against the bug.

The source guard is an **exact list** of the five array-returning helpers rather than a pattern, so adding a sixth is a deliberate edit here instead of something a regex silently does or does not cover.

**Mutation-tested** (on copies — `git checkout <file>` would discard the real edit alongside the mutation):

| Mutation | Tests that fire |
| --- | --- |
| `Get-GoogleRows` reverted whole | 7 |
| comma dropped from the empty path only | 4 |
| comma dropped from `Invoke-GooglePagedApi` | 5 |
| comma dropped from `Get-GscInventory` | 1 |

Restoring clears all of them.

## Verification

- **Pester 84/84** across both test files (was 75; +15 here). Installed pwsh 7.5.4 in the sandbox for this — the harness runs cleanly on Linux, so no false-red to discount.
- `PSScriptAnalyzer` — **0 errors, 0 parse errors**; no new warnings on any touched file (the test file's single `PSUseBOMForUnicodeEncodedFile` is pre-existing).
- `Invoke-Formatter` idempotent on all three `.ps1` files.
- `python3 tests/workflow-logic/run_all.py` — **32/32 modules**.
- `check-workflow-doc-consistency.py` 98/91 · `check-workflow-references.py` · `prettier@3.8.1 --check .` (CI-pinned) all clean.

**Not verifiable pre-merge:** that 506 actually goes green. It runs on a schedule against `main` and needs live Google credentials via the KV/OIDC path, so the honest acceptance test is the next scheduled run after merge — at which point **#875 should auto-close on 740's next sweep**, which also discharges the one criterion #843 left open (the recovery half of the poll, unit-tested but never yet observed live). If #875 does not close after 506 goes green, that is a real defect in 740's recovery path and worth reopening #843 for.

One caveat for whoever confirms: 506's summary also lists nine **provisioned-vs-served disagreements**. Those are real findings, not this crash, and they are #883's subject — this PR does not touch them, and 506 going green is not a claim about them.

Ledger row is **L29**, not L28: open PR #897 has claimed L28, and ids are asserted unique but not contiguous.

## Gates

None — no `environment:` in the diff, no credentials, no workflow file touched. Normal merge queue. Draft, per the standing rule that @clarkemoyer promotes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GCaEp9UMDXFCL296vW12bn)_


clarkemoyer:
**CI green at `d7886c7`** — Validate Repository ✅ · Phantom Revert Guard ✅ · CodeQL ✅ · Analyze Code (actions) ✅ · label-sync ✅ (`sweep` skipped as normal). Copilot's reviewer was still queued at time of writing; no review threads yet.

Worth naming the two steps that matter for this change, since a repo-wide CI pass is not by itself evidence they ran:

- **Run PowerShell unit tests (Pester)** ✅ — the 24-test module, including the 15 added here.
- **Lint PowerShell (PSScriptAnalyzer)** ✅ and **Invoke-Formatter** ✅ — 0 errors, 0 parse errors.

Branch is level with `main` (0 behind) as of the push. Per the queue etiquette in `AGENTS.md`, whoever promotes should re-fetch and merge `main` immediately beforehand — `main` moves fast in this repo and Phantom Revert Guard re-reds on contact when it does.

Draft, awaiting @clarkemoyer's promotion.

---

clarkemoyer:
## Landing check: ready to queue, but #885 fixes the same function

Verified from a landing run — this PR is `mergeable_state: clean`, all 7 checks green, zero review threads. Nothing on the branch blocks it.

One thing to settle first: **#885 (`fix/865-getgooglerows-strictmode`, open, already out of draft) rewrites the same `Get-GoogleRows`**, with a semantically equivalent fix. Test-merged against `main` @ `0a5b2a4`:

```
#898 + #885   CONFLICT — scripts/google-api-common.ps1, scripts/google-telemetry-reachability.ps1
#898 + #897   CLEAN
#898 + #837   CLEAN
```

This PR is the broader of the two (4 telemetry helpers vs 1, plus the L29 ledger row). Full comparison and the two things #885 has that this does not are on #885. Whichever lands first leaves the other conflicting, so they want an explicit order rather than both being queued.

No conflict with the other two open `agentic-os` PRs — #897 and #837 both merge clean against this branch, so ordering among those three is free.

---

clarkemoyer:
## ⚠️ Duplicate fix in flight — #885 fixes the same defect, and the two branches conflict

Landing-run check (agentic-os worker). This PR is **green, threads clean, `mergeable_state: clean`** and ready for @clarkemoyer to promote — but **do not promote it without first deciding what happens to #885**, which fixes the same `Get-GoogleRows` bug.

Neither PR references the other. They were authored by two sessions working two different issues:

| | #898 (this PR) | #885 |
| --- | --- | --- |
| Claimed issue | #875 (740's rolling 506 alert) | #865 (traffic-ordered rollout) |
| Branch | `claude/sweet-hawking-89uo4t` | `fix/865-getgooglerows-strictmode` |
| State | draft, `clean` | ready, `blocked` |
| Labels | `bug`, `agentic-os`, `priority: high` | none |

**They conflict textually.** `git merge-tree` against their merge base reports `changed in both` on `scripts/google-api-common.ps1` **and** `scripts/google-telemetry-reachability.ps1`. Whichever lands second hits a conflict in the exact function whose fix is subtle enough to be lost in a careless resolution — the unary comma is one character, and dropping it restores the bug while leaving every populated-response test green.

### This PR is a strict superset

I compared the two diffs function by function rather than trusting either description:

| Fix | #885 | #898 |
| --- | --- | --- |
| `Get-GoogleRows` — indexer instead of enumerating `.Properties.Name` | ✅ | ✅ |
| `Get-GoogleRows` — `, @()` on every return path | ✅ | ✅ |
| `Invoke-GooglePagedApi` | ❌ | ✅ |
| `Get-GtmInventory` | ❌ | ✅ |
| `Get-Ga4Inventory` | ❌ | ✅ |
| `Get-GscInventory` | ❌ | ✅ |
| Ledger row | ❌ | ✅ L29 |
| Source-form guard over an exact list of the 5 array-returning helpers | partial (2 regexes) | ✅ |

Nothing in #885 is absent here. Its two apparent extras are not coverage:

- **`$props -eq $null` guard** in its `Get-GoogleRows` — unreachable. `-Response` is `Mandatory`, so `$null` is rejected at binding time and `.PSObject` is never `$null` on a bound value. This PR documents that deliberately and pins it with a test.
- **`@(Get-GoogleRows … | Where-Object { $null -ne $_ })`** in `Get-Ga4Sessions` — belt-and-braces at one of two call sites, redundant once the helper is correct, and it would mask a regression at the site that still reads `.Count` bare (`google-analytics-report.ps1:92`).

### Recommendation

Promote **#898**, close **#885** as superseded, and re-point #865 at this PR so the traffic-ordering epic still tracks its dependency. I have left the matching note on #885 and have **not** closed it — that call belongs to a human, and #885 is not an `agentic-os` PR so it is outside this worker's mandate.

---

Issue #901, opened 2026-07-28
integrate-clone-into-nextjs.mjs: the _clone-host sentinel route never becomes a route


clarkemoyer:
## What

`scripts/integrate-clone-into-nextjs.mjs` step 4 writes a placeholder route and explains itself as:

```js
// 4) Guarantee at least one valid route remains so `next build` succeeds while
//    public/index.html serves '/'. (A bare app dir with no page errors out.)
const sentinelDir = join(appDir, '_clone-host');
```

It writes to `src/app/_clone-host/page.tsx`. **Underscore-prefixed folders are private folders in the App Router** and are excluded from routing entirely, so that file never produces a route. The step does not do what its comment says.

## Why it matters

Not a live breakage — but the comment asserts a build invariant that isn't actually being maintained. Anyone reasoning about the pipeline (or debugging a failed `next build`) will trust it. If the premise were true, every clone repo would be one deleted file away from a broken build; since it isn't, the file is dead weight that also has to be explained to every future reader.

The underlying premise looks wrong too, at least on Next 16.

## Evidence

Verified while conforming `FFC-EX-slopestohope.org` to the runbook (FreeForCharity/FFC-EX-slopestohope.org#40):

- With `src/app/_clone-host/page.tsx` present, `next build` reports exactly one route: `/404`. The sentinel is absent from the route table.
- **Deleting it entirely**, leaving only `src/app/layout.tsx` and no `page.*` anywhere, still builds and exports cleanly:

```
✓ Compiled successfully
✓ Generating static pages (1/1)
✓ Exporting (1/1)

Route (pages)
─ ○ /404
```

`out/` contained the full clone plus `CNAME` and `.nojekyll` in both cases.

Environment: Next 16.2.12, Node 24, `output: 'export'`.

## Suggested fix

Drop step 4 and the sentinel, and correct the comment. If a route genuinely is required on some Next version the fleet still targets, the sentinel needs a **non-underscore** name (e.g. `clone-host/`) to actually route — but note that would then export a stray `out/clone-host/index.html` into the published site, which is its own small problem worth avoiding.

Worth confirming against whichever Next version the older FFC-EX repos pin before removing fleet-wide.

Refs #702.

PR #903, opened 2026-07-28
Close the two blind spots in the FFC-EX clone pipeline (entity-escaped JSON, runtime-built URLs)

### LINKED ISSUE #902: Clone fidelity gate misses two whole classes of legacy dependency (escaped JSON, runtime-built URLs)
## What

`scripts/clone-site-static.mjs` decides whether a clone is self-contained with a regex over HTML attributes:

```js
const refRe = /(?:src|href)\s*=\s*["'](https?:\/\/[^"']+)["']/gi;
```

That result becomes `clone-report.json` → `remainingExternalHosts`, which the runbook names as the pre-cutover verification gate.

It cannot see two classes of dependency, and **both are load-bearing on real FFC-EX sites**:

**1. URLs inside HTML-entity-escaped JSON.** Elementor stores widget config in `data-settings` as escaped JSON, so a URL there is delimited by `&quot;`, not a quote character:

```html
data-settings="{&quot;background_slideshow_gallery&quot;:[{&quot;url&quot;:&quot;https://olddomain.com/wp-content/uploads/2025/10/hero.jpg&quot;}]}"
```

No `src=`/`href=`, so the regex never sees it. The hero slideshow keeps loading from the decommissioned host.

**2. URLs assembled at runtime.** Elementor, Essential Addons and ElementsKit build content-hashed webpack chunk URLs in JavaScript — `counter.12335f45aaa79d244f24.bundle.min.js` and friends. They appear in no attribute anywhere, so no static scan of any kind can find them. httrack never mirrors them, and the report shows a clean bill of health.

## Why it matters

This is the mechanism behind the bug that started FreeForCharity/FFC-EX-slopestohope.org#40: the "pounds distributed" counter sat at `0` in production for months. The counter markup ships a literal `0` and relies on `counter.*.bundle.min.js` to animate it. The chunk was never mirrored, the clone report was clean, and the site shipped.

`docs/ffc-ex-clone-fidelity-audit.md` already flags that image-count parity is a heuristic and "visual diffing against live is the authoritative gate" — but a visual diff would not have caught this either. The counter *looks* fine; it just shows the wrong number.

## Suggested fix

A browser-based gate that is immune to both classes. Load every page with **all requests to the legacy host aborted**, and fail on either:

- any request to the legacy host (a surviving dependency), or
- any request to our own origin that 404s (an asset the mirror never captured).

Implemented and working in `FFC-EX-slopestohope.org/tools/verify-no-legacy.mjs`; happy to port it here as `scripts/verify-no-legacy.mjs` with `--domain` / `--base` parameters so 702 can run it as a gate.

Two lessons from building it worth carrying over:

- The second condition matters as much as the first. An earlier version treated a failed same-origin request as non-fatal and reported **13/13 passing while four Elementor bundles were silently 404ing**. Missing local assets have to be hard failures.
- Discovery of runtime assets can be automated the same way: drive each page in a browser, collect same-origin 404s, fetch those from the live origin, repeat until a round adds nothing. One pass is never enough, since a freshly fetched chunk requests further chunks. Also implemented (`tools/sync-runtime-assets.mjs`).

For class 1 specifically, the URL rewriter needs to stop matching at an HTML-entity boundary. A naive rewrite that runs past `&quot;` swallows the rest of the JSON blob, and because a global regex resumes after the whole match, **every following URL in that blob silently goes unrewritten**. Encoding the boundary as a negative lookahead in the character class fixes it:

```js
const ENTITY_BOUNDARY = '&(?:quot|apos|lt|gt|#0?3[49]|#x27);'
const URL_CHAR = `(?:(?!${ENTITY_BOUNDARY})[^\\s"'<>()\\\\])`
```

## Scope note

This is the verification half of converging the two cloning paths (`clone-site-static.mjs` vs the bespoke `tools/clone-wordpress.mjs` in the slopestohope repo). Tracking the convergence separately; this issue is specifically the gate, which is useful fleet-wide regardless of which cloner produces the mirror.

Refs #702.

clarkemoyer:
Closes #902. Refs #702.

## Why

The clone pipeline's fidelity signals could not see two whole classes of legacy dependency, and both are load-bearing on real FFC-EX sites.

**1. URLs inside HTML-entity-escaped JSON.** Elementor stores widget config in `data-settings`, where a URL is delimited by `&quot;` rather than a quote character:

```html
data-settings="{&quot;background_slideshow_gallery&quot;:[{&quot;url&quot;:&quot;https://olddomain.com/wp-content/uploads/hero.jpg&quot;}]}"
```

No `src=`/`href=`, so `clone-site-static.mjs`'s scan never sees it and httrack never rewrites it. The hero slideshow keeps loading from the host being decommissioned.

**2. URLs assembled at runtime.** Elementor / Essential Addons / ElementsKit build content-hashed webpack chunk URLs in JavaScript — `counter.12335f45aaa79d244f24.bundle.min.js` and friends. They appear in no attribute anywhere, so **no static scan of any kind can find them**. httrack never mirrors them and `clone-report.json` stays clean.

Class 2 is the mechanism behind the bug that started FreeForCharity/FFC-EX-slopestohope.org#40: the "pounds distributed" counter sat at `0` in production for months. The markup ships a literal `0` and relies on the counter chunk to animate it. Note a visual diff would not have caught this either — the page *looked* correct, it just showed the wrong number.

### `clone-site-static.mjs` — localize what httrack can't see

After the mirror completes, rewrite legacy absolute URLs in place, handling both the plain and JSON-escaped forms.

The subtle part: **the entity boundary has to stop the match itself, not be trimmed afterwards.** A global replace resumes scanning after the whole match, so a pattern that runs past `&quot;` swallows the rest of the JSON blob and every following URL in it silently goes unrewritten. Encoded as a negative lookahead in the character class.

New `localizedInPlace` count in the report, and `remainingExternalHosts` now carries an explicit note that it is necessary but **not sufficient** as a gate.

### `sync-runtime-assets.mjs` (new) — mirror what only appears at runtime

Serves the clone, drives every page in a browser, notes which same-origin requests 404, fetches those from the live origin, and repeats. Repetition is required: a freshly fetched chunk routinely requests further chunks, so one pass is never enough. Stops when a round discovers nothing new. Read-only against the live site.

### `verify-no-legacy.mjs` (new) — the cutover gate

Loads every exported page with **all requests to the legacy host aborted**, and fails on either a surviving legacy dependency **or** a same-origin 404.

The second condition matters as much as the first. An earlier version of this gate treated same-origin failures as non-fatal and reported **13/13 pages passing while four Elementor bundles were silently 404ing**. Missing local assets have to be hard failures.

Page discovery is automatic (every directory with an `index.html`), with `--max-pages` capping large mirrors and reporting what it skipped rather than truncating silently. Writes a verdict table to `$GITHUB_STEP_SUMMARY`.

### Workflow 702 + runbook

Pipeline is now clone → **runtime-sync** → integrate → build → **gate**. Chromium is installed ad hoc for the two browser-driven stages, so this repo stays dependency-free (no root `package.json`); both scripts import Playwright dynamically and print the install command if it's absent.

## Testing

- `node scripts/clone-site-static.mjs --self-test` — 5/5 pass, covering the entity-escaped blob (both URLs rewritten), JSON-escaped slashes preserved, `www.` host, other hosts untouched, and bare origin → `/`. Wrapped in `scripts/tests/clone-site-static.Tests.ps1` following the `preflight-cutover` convention.
- **`verify-no-legacy.mjs` run against the real `FFC-EX-slopestohope.org` export**: discovered **18 pages unaided** and all 18 passed with `slopestohope.com` blocked at the network layer.

Not yet exercised end-to-end inside workflow 702 — the new steps are wired but a dispatch run against a live domain is the remaining validation, and I'd suggest doing that on one site before this goes near the Wave 1 batch.

## Convergence note

This lands the *verification and asset-capture* half of converging the two cloning paths. `FFC-EX-slopestohope.org` still carries a bespoke `tools/clone-wordpress.mjs` because httrack alone could not produce a working clone of that site; with `sync-runtime-assets.mjs` in the pipeline that gap is now closed generically, so the bespoke cloner should be retired in favour of 702 once this has been validated on a live dispatch. Tracking that as the follow-up rather than doing it blind in the same PR.

Separately, #901 covers the dead `_clone-host` sentinel in `integrate-clone-into-nextjs.mjs`; not touched here to keep this PR to one concern.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MAqyGR5ynSAvDTc52unGve)_

### DIFF
--- a/.github/workflows/702-ffc-ex-clone-deploy.yml
+++ b/.github/workflows/702-ffc-ex-clone-deploy.yml
@@ -235,6 +235,15 @@ jobs:
       - name: Install httrack
         run: sudo apt-get update && sudo apt-get install -y httrack
 
+      - name: Install Playwright (runtime-asset sync + cutover gate)
+        # This repo is intentionally dependency-free, so Playwright is installed
+        # ad hoc for the two browser-driven stages rather than committed as a
+        # dependency. Chromium only; --with-deps pulls the shared libraries the
+        # runner image does not ship.
+        run: |
+          npm i --no-save playwright
+          npx playwright install --with-deps chromium
+
       - name: Resolve target repo
         id: target
         run: |
@@ -253,6 +262,17 @@ jobs:
             --depth "${{ inputs.depth }}" \
             ${{ inputs.exclude && format('--exclude {0}', inputs.exclude) || '' }}
           cat "$RUNNER_TEMP/clone/clone-report.json"
+
+      - name: Mirror runtime-loaded assets
+        working-directory: automation
+        # httrack only fetches what it can see in markup. Elementor and friends
+        # assemble content-hashed chunk URLs in JavaScript, so those are missing
+        # from the mirror and every widget depending on them dies silently.
+        # See #902.
+        run: |
+          node scripts/sync-runtime-assets.mjs \
+            --domain "${{ inputs.domain }}" \
+            --dir "$RUNNER_TEMP/clone/${{ inputs.domain }}"
 
       - name: Check out target FFC-EX repo
         uses: actions/checkout@v7.0.1
@@ -282,6 +302,16 @@ jobs:
             echo "missing out/index.html"
             exit 1
           fi
+
+      - name: Gate - clone must be self-contained
+        if: ${{ inputs.build_check }}
+        # Loads every exported page with all requests to the live host aborted.
+        # Fails on a surviving legacy dependency OR a missing local asset - the
+        # two classes clone-report.json cannot see. See #902.
+        run: |
+          node automation/scripts/verify-no-legacy.mjs \
+            --domain "${{ inputs.domain }}" \
+            --dir ffc-ex/out
 
       - name: Commit and open PR on the FFC-EX repo
         working-directory: ffc-ex
--- a/docs/ffc-ex-static-clone-runbook.md
+++ b/docs/ffc-ex-static-clone-runbook.md
@@ -8,7 +8,7 @@ from the live sites.
 
 ## Pipeline
 
-Two scripts in `scripts/`, plus the repo's own `next build`:
+Four scripts in `scripts/`, plus the repo's own `next build`:
 
 1. **`clone-site-static.mjs`** — mirrors the live site with `httrack` (strict containment: only the
    target domain's HTML, but pulls page assets/images even when CDN/S3-hosted, and Google Fonts),
@@ -20,7 +20,24 @@ Two scripts in `scripts/`, plus the repo's own `next build`:
    don't collide with the clone's pages. With `output: 'export'`, `next build` copies `public/`
    verbatim into `out/`, so the export ships the exact clone. Writes `public/CNAME` (apex).
 
-3. **`next build`** in the FFC-EX repo → `out/` is the deployable static clone.
+3. **`sync-runtime-assets.mjs`** — mirrors the assets that only appear at runtime. httrack fetches
+   what it can see in markup; Elementor / Essential Addons / ElementsKit assemble content-hashed
+   webpack chunk URLs in JavaScript, so those are absent from the mirror and every widget depending
+   on them dies silently. This drives each page in a browser, fetches whatever 404s, and repeats
+   until a round finds nothing new (a fresh chunk routinely requests further chunks). Run it between
+   steps 1 and 2.
+
+4. **`next build`** in the FFC-EX repo → `out/` is the deployable static clone.
+
+5. **`verify-no-legacy.mjs`** — the cutover gate. Loads every exported page with all requests to the
+   live host aborted, and fails on a surviving legacy dependency **or** a missing local asset.
+
+Steps 3 and 5 need a browser. This repo is deliberately dependency-free, so install Playwright just
+for the run:
+
+```bash
+npm i --no-save playwright && npx playwright install --with-deps chromium
+```
 
 ## Steps (per domain)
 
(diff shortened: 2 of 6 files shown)


clarkemoyer:
All three review findings were real and are fixed in 380278e. I reproduced each before changing anything:

```
join('/tmp/site', '../site2/x')  = /tmp/site2/x
  startsWith(root)               = true   <-- escapes containment
  relative-based check           = false

decodeURIComponent('/%2e%2e/x')  = /../x
  write dest would have been     = /tmp/x   (outside the clone dir)

'https://cdn.other.com/?ref=example.org'.includes('example.org') = true
```

**Path containment (both scripts).** `file.startsWith(root)` is a string prefix, not containment — `join()` normalises `..`, so a sibling directory sharing the prefix passes. Replaced with a `resolveWithin()` helper that compares the relative path.

**The write path was the one that mattered.** In `sync-runtime-assets.mjs`, asset paths come from URLs observed at runtime — from the cloned site's own JavaScript, which is content we do not control. A page requesting `/%2e%2e/%2e%2e/x` would have caused a remotely-fetched file to be written outside the clone directory, onto the CI runner's workspace. Those paths are now refused and recorded rather than fetched. Good catch; this one was worth the review on its own.

**Host matching.** Agreed on both directions. Now parses the URL and compares `hostname`, treating any subdomain as legacy (`staging.<domain>` is the same install being retired) while correctly *not* matching a third party that merely ends with the bare name — `<charity>.rallyup.com` is a real case in this fleet.

Added a `--self-test` to `verify-no-legacy.mjs` and three Pester cases, including explicit guards against reverting to a substring test or a prefix check:

```
ok   domain in a query string is NOT legacy
ok   host merely ending in the bare name is NOT legacy
ok   sibling-dir traversal is contained
ok   encoded traversal is contained
9/9 passed
```

`clone-site-static.mjs --self-test` still 5/5, and the gate still reports **18/18 pages** against the real `FFC-EX-slopestohope.org` export after the rewrite.

---

clarkemoyer:
**Conductor run 44 — reviewed and queued.**

Required checks green (Validate Repository, Phantom Revert Guard, CodeQL, Analyze Code), `MERGEABLE/CLEAN`, 2 review threads / 0 unresolved. Supersession grep against `main` is clean — no same-purpose implementation of `sync-runtime-assets` / `verify-no-legacy` has landed, and #905's `_clone-host` removal touches only `integrate-clone-into-nextjs.mjs`, so there is no file overlap between the two.

On the validation caveat you flagged ("not yet exercised end-to-end inside workflow 702"): merging this stages nothing. **702 is `workflow_dispatch`-only** — no schedule, no `push` trigger — so the new runtime-sync and gate steps cannot execute until someone dispatches it. Merging therefore makes the live-dispatch validation *easier* (it runs from `main` rather than a branch) without putting the pipeline at risk, which is why I queued rather than held.

**The validation is still owed and is not mine to run** — a 702 dispatch is a gated write. Carrying it to Clarke as the follow-up, with your recommendation intact: one live single-site dispatch before this goes anywhere near the Wave 1 batch.

Retiring `FFC-EX-slopestohope.org`'s bespoke `tools/clone-wordpress.mjs` remains the correct follow-up after that dispatch, not before it.

PR #904, opened 2026-07-28
feat: add inbound transfer path for foreign-registrar domains (Wix → eNom → Cloudflare)

clarkemoyer:
## Problem

The transfer pipeline (`115`/`116`/`117`) assumes the losing registrar is eNom. A domain that arrives still registered at Wix, Squarespace, or GoDaddy classifies as readiness `review` in `domain-transfer-preflight.ps1:231` and stops there — runbooks are only written for `ready` (line 321), so those domains produce a flag and then silently fall off the worklist. No path was defined for them.

Wix is the strict case, and it is not the case the existing tooling assumes:

- Wix does not permit editing NS records on a Wix-registered domain ([its own help center](https://support.wix.com/en/article/request-changing-name-server-ns-records-for-a-wix-domain)).
- Cloudflare's Free plan requires full nameserver delegation; partial/CNAME setup is Business-plan only.
- So a Wix domain **cannot reach Cloudflare at all** until the registration moves. Adding the site in the dashboard creates a zone that stays `Pending` forever.
- Cloudflare Registrar cannot be the direct destination either — it only accepts domains whose zone is already active on Cloudflare nameservers, which is exactly what Wix blocks.

eNom breaks the deadlock: it accepts a normal EPP transfer from Wix, and once the domain is there nameservers can point anywhere.

## What this adds

| File | Role |
| --- | --- |
| `scripts/domain-inbound-transfer-preflight.ps1` | Offline classifier. Registrar family → `nsDelegationAllowed` → `inboundPath` (`transfer-required` vs `ns-delegation`). No API calls, no secrets. |
| `.github/workflows/123-domain-inbound-transfer-preflight.yml` | Read-only RDAP probe + preflight + foreign-registrar worklist. No credentials and no environment gate — RDAP is public and the analysis is offline. `domains` is required; there is deliberately no fleet-wide default. |
| `.github/workflows/231-whmcs-domain-order-add.yml` | Places the WHMCS domain order the migration needs (pid 41 transfer / pid 39 register), wrapping the existing `scripts/whmcs-order-add.ps1`. `mode` defaults to `dry-run`; gated on `whmcs-prod`. |
| `docs/wix-domain-migration.md` | The four-stage playbook. |
| `701-website-provision.yml`, `02-website-request.yml` | Captures registrar-of-record at intake, wired through parse → outputs → the provisioned repo's content record. |

## The key behavioural distinction

A blocked *transfer* is not a blocked *nameserver change*. An NS-capable registrar can point at Cloudflare today even while its registrar transfer is locked, and the preflight says so explicitly rather than reporting the domain as simply stuck.

The four stages, for the transfer-required case:

1. Wix — disable DNSSEC, unlock, get EPP
2. WHMCS — transfer-in order to eNom (workflow `231`)
3. **NS → Cloudflare** — no waiting period applies; this is the step that delivers
4. 60 days later — eNom → Cloudflare Registrar via the existing `115`/`116`/`117`, optional cost optimization

Stage 4 is deliberately framed as unhurried: after stage 3 the site already has Cloudflare DNS, CDN, and WAF.

Runbooks are written for **every** actionable foreign domain, blocked ones included with their blockers at the top — the fix for the silent-drop behaviour above. Each one opens with a stage-0 DNS capture, because dropping `MX`/SPF/DKIM/DMARC is the main way this migration hurts a charity.

## Source of truth

**WHMCS is FFC's live record** of the domains we hold. RDAP answers a narrower question — what the registry says about one domain right now — which is the useful cross-check precisely because a domain at a foreign registrar may not be in WHMCS at all. `docs/domain-registry-truth.csv` is a committed point-in-time snapshot for reference only and is **not** a worklist; the docs and parameter help say so.

> **Correction.** An earlier revision of this description claimed that running the preflight across `docs/domain-registry-truth.csv` surfaced three domains (`letspiritlead.org`, `srrn.net`, `ssrn.net`) as locked-but-movable. **That claim is withdrawn** — it came from that stale snapshot, which is exactly what this PR now documents as non-authoritative. The classification logic behaves as described, but the input was not a live record, so the conclusion about those three domains is unsupported. Anything of the sort should be re-derived from WHMCS for a domain someone has actually asked us to migrate.

## Verified against the real case

`cactuswrenpreschool.com`, via RDAP: registrar **Wix.com Ltd.** (IANA 3817), created 2007-05-31, expires 2027-05-31, `transferLocked: False`. Classifies `FOREIGN_TRANSFER_REQUIRED` / `transfer-required` / `readiness: ready`.

In WHMCS (via the ungated read workflows 201/202/219): the domain is absent from all 324 domain rows, and client **379** (Cactus Wren Cooperative Preschool) holds only 501c3 Onboarding (Active) and Free Charity Website / GitHub Pages (Pending) — **no domain product**. That absence is what workflow 231 exists to fix.

## Testing

- 25 Pester cases (`scripts/tests/domain-inbound-transfer-preflight.Tests.ps1`) covering registrar classification, every blocker, runbook shape per path, CSV-mode counts, and the usage contract.
- 3 added to `tests/workflow-logic/test_701_parse.py`, including a guard that the issue-template label and the `extractSection()` argument cannot drift apart — that drift would silently parse the registrar as empty, which is precisely how a Wix domain would slip through intake unnoticed.
- Main's two guard tests added during this PR's life pass unchanged: `test_no_untrusted_expression_in_run.py` against 123 and 231, and `test_gated_env_hygiene.py` against 231's `whmcs-prod` gate.
- Latest merged tree: **181 Pester** passed, **565 workflow-logic** passed, no display-number collisions across 100 workflows, catalog up to date, doc-consistency OK, Prettier clean, PSScriptAnalyzer formatter clean with no errors.
- Remaining failures are the 3 pre-existing in `test_720_owner_parse.py`, confirmed on a clean tree.

## Numbering

Originally 229; renumbered to **231** so #825 keeps 229 uncontested. Renumbered in the workflow name, the `concurrency` group, the safety table (moved after 230), the decision guide, and the playbook.

## Not done here

No order has been placed — nothing has run in `execute` mode. Two product questions remain for a human, independent of merging:

1. **Duplicate client records** — 51 ("Cactus Wren Preschool", 2017, Active, whose application already records the domain) and 379 ("Cactus Wren Cooperative Preschool", 2026) are the same org.
2. **Application vs. build mismatch** — order 798 answers *"Footer-Only Template — add FFC footer to our existing site"*, but `FFC-EX-cactuswrenpreschool.com` is a full Next.js site migrated off Wix. That order is still `Pending`.


clarkemoyer:
All five review comments addressed across 915da14 and 6b9e67e.

**Injection via `${{ inputs.domains }}`** — real, and it applied more broadly than the one line flagged. Fixed in 6b9e67e.

All three dispatch inputs now travel via `env:` and are read from `$env:` / `process.env`, never spliced into a script body. Two things the comment prompted me to check beyond the reported line:

- `min_days_to_expiry` / `post_reg_lock_days` had the same exposure. `type: number` on `workflow_dispatch` is a UI hint, not a guarantee — an API dispatch can send any string — so those are now cast to `[int]` after arriving via env.
- The `github-script` step interpolated `issue_number` into a JS body, which is the worse one given that step holds the `issues: write` token. Now `Number(process.env.INPUT_ISSUE_NUMBER)` with a positive-integer check before the API call.

Domain tokens also get validated against a hostname pattern. They only ever become CSV data, so this is defence in depth rather than a fix, but there's no reason to carry a non-hostname forward. Verified the pattern accepts `example.org`, `sub.example.co.uk`, `a.io` and rejects `bad_domain.com`, `no-tld`, `-lead.com`, `'; whoami #`, `x.com;rm -rf /`.

**`[CF]` tag / catalog `apis`** — both correct, and they were one issue: the catalog is generated from the workflow name tag. Renamed to `123. Domain - Inbound Transfer Preflight (Report)` to match 121, the other no-credential preflight, and regenerated — the entry's `apis` is now `[]` as a consequence rather than by hand-editing generated output. `generate-workflow-catalog.py --check` passes.

**Local vs UTC clock** — correct, and it was a real bug. `ConvertTo-DateOrNull` normalizes with `AssumeUniversal|AdjustToUniversal` while `$today` was local, so `daysToExpiry` / `daysSinceRegistration` could shift a day either side of midnight depending on runner timezone — worst exactly at a 60-day lock boundary, where the whole point is answering "may this transfer start yet". Now `[datetime]::UtcNow.Date`, with the Pester clock aligned to match.

**Workflow 102's title (low-confidence comment)** — worth acting on, and it wasn't isolated. Checking the other references, all four were paraphrases rather than real titles:

| runbook said | actual |
| --- | --- |
| `110. Cloudflare - Zone Create` | `110. DNS - Create Zone (Admin) [CF]` |
| `102. Domain - Add to FFC Cloudflare and WHMCS` | `102. Domain - Add to FFC Cloudflare + WHMCS Nameservers (Admin) [CF+WHMCS]` |
| `103. Domain - Enforce Domain Standard` | `103. Domain - Enforce Standard (GitHub Apex + M365) [CF+M365]` |
| `115. Domain - Transfer Readiness Preflight` | `115. Domain - Transfer Readiness Preflight (Report) [WHMCS]` |

All four now match exactly, in both the runbook writer and `wix-domain-migration.md`.

**CI** — the `Validate Repository` failure was `Invoke-Formatter` drift on the new script (hashtable assignment alignment), fixed in 915da14. Locally: formatter clean across all `.ps1`, 150 Pester passed, `test_701_parse.py` 9 passed, catalog check up to date, Prettier clean.

---

clarkemoyer:
Two corrections from FFC, applied in fdc769d — plus a finding of mine that I'm withdrawing.

**Scope.** `domains` defaulted to blank, and blank meant "classify the whole committed registry-truth CSV" — 84 domains, most of which nobody asked about. Wrong default for a tool that emits per-domain migration runbooks. `domains` is now **required**, the fleet-wide branch is deleted, and `run-name` echoes the requested domains so a run is identifiable from the Actions list without opening it. The workflow was read-only either way, but breadth of inquiry should be a deliberate choice rather than something inherited from a default.

**Source of truth.** The script and playbook presented RDAP — and specifically the committed `docs/domain-registry-truth.csv` — as authoritative. It is not. **WHMCS is FFC's live record** of the domains we hold; that CSV is a point-in-time snapshot kept for reference. RDAP still earns its place, but for the narrower job it actually does: telling you what the registry says about *one domain right now*, which matters precisely because a domain at a foreign registrar may not be in WHMCS at all. Parameter help, decision-guide notes, and the playbook now say this, and the example that pointed at the committed CSV is replaced with one that probes the single domain being migrated.

**Withdrawing a finding.** Earlier I reported that three domains — `letspiritlead.org`, `srrn.net`, `ssrn.net` — were locked at foreign registrars but could move to Cloudflare today. That came from running the preflight across `docs/domain-registry-truth.csv`, which is exactly the stale snapshot this change is about. Treat it as unverified: the classification logic behaves as described, but the input wasn't a live record, so the conclusion about those three domains isn't supported. Anything of the sort should be re-derived from WHMCS against a domain someone has actually asked us to migrate.

The intended use is narrow and per-domain: `cactuswrenpreschool.com` is the case this was built for.

Verification after the change: 150 Pester passed, formatter clean, catalog check up to date, Prettier clean.

---

clarkemoyer:
Added **workflow 229 — WHMCS - Domain Order Add (Register/Transfer)** in 1064cb4.

`scripts/whmcs-order-add.ps1` already wrapped WHMCS `AddOrder` correctly, but nothing called it — placing a domain order meant running the script by hand with production credentials. This is the missing link in the inbound path: a charity whose domain sits at a foreign registrar has **no domain product in WHMCS at all**, so there's nothing to drive the transfer from until an order exists.

Confirmed against the real case. `cactuswrenpreschool.com` → client **379** (Cactus Wren Cooperative Preschool) holds exactly two products — 501c3 Onboarding (Active) and Free Charity Website / GitHub Pages (Pending) — and **no domain product of any kind**.

**Product ids from the 202 export, not guesswork:**

| order_type | pid | product |
| --- | --- | --- |
| `transfer` | 41 | Transfer your Existing Domain Name to the FFC CloudFlare |
| `register` | 39 | Free new .org Domain Name in the FFC CloudFlare |

Defaults follow how FFC's existing charity orders are actually recorded — `payment_method=mailin`, `billing_cycle=onetime` — rather than the script's generic `banktransfer`/`free`.

**Safety, since this is the first production billing write here:**

- `mode` defaults to **dry-run**: previews the request, writes nothing.
- Sits on the gated `whmcs-prod` environment on top of that.
- `no_email` defaults to **true**, so a first live run can't surprise a charity with an unexpected order confirmation.
- Price forced to 0; the script already skips when the client has a non-terminated service for the product unless `allow_duplicate`.
- `concurrency` group serializes dispatches so two runs can't race into duplicate orders.
- `client_id` and `domain` validated before anything is called, and every dispatch input travels via `env:` rather than being interpolated into the script body.

Registered in the safety table, the decision guide, and stage 2 of the Wix playbook. Catalog regenerated (100 workflows); `check-workflow-doc-consistency.py` passes; 150 Pester + 9 parse tests green; Prettier clean.

No order has been placed — nothing has run in `execute` mode.

**Two open questions for a human**, both surfaced while confirming the above:

1. **Duplicate client records.** Client **51** ("Cactus Wren Preschool", 2017 Pre-501c3 onboarding, still Active) and client **379** ("Cactus Wren Cooperative Preschool", 2026) are the same org — same mission text, and 51's application already records `current/desired domain name = www.cactuswrenpreschool.com`. The transfer order should go on 379, but 51 wants merging or closing.

2. **Application vs. build mismatch.** Order 798's answer to *"Which starting point fits your charity?"* is **"Footer-Only Template — add FFC footer to our existing site"**, but `FFC-EX-cactuswrenpreschool.com` is a complete Next.js site migrated off Wix. That order is still `Pending`, so now is the moment to reconcile which one is intended.

---

clarkemoyer:
Conductor supervision (run 47).

**Resolved all four Copilot threads** — I verified each against the current branch rather than
resolving on "outdated" status, and all four are genuinely fixed:

| Finding | Current state |
| --- | --- |
| PowerShell injection via `'${{ inputs.domains }}'` | fixed properly — passed as `INPUT_DOMAINS` env, read as `$raw = "$env:INPUT_DOMAINS"` |
| `[CF]` tag on a workflow that makes no Cloudflare call | dropped; name is `123. Domain - Inbound Transfer Preflight (Report)` |
| catalog `"apis": ["CF"]` | now `"apis": []` |
| local `$today` vs UTC-normalized registry dates | `$today = [datetime]::UtcNow.Date`, with a comment explaining the off-by-one |

The injection fix is the right one (env indirection, not quote-escaping), and it was worth fixing
even on a read-only workflow — `GITHUB_TOKEN` is still in scope.

**Not queueing yet — two blockers, neither yours to fix alone:**

1. **Display-number collision on 229.** #825 also adds a workflow numbered 229 (client field
   populate; yours is domain order add). The filenames differ, so **git merges both without a
   conflict** and the collision lands silently — one of the two then becomes invisible to every
   check keyed on the display number. I've shipped #917 to make that a hard CI failure, but the
   renumbering decision is yours and #825's, not the guard's. One of you needs a free number in the
   `2xx` band; whoever rebases second will otherwise fail `Validate Repository`.
2. **Rebase needed.** #916 merged a few minutes ago (22:16Z) and rewrote the same section of
   `docs/workflow-safety-and-approvals.md` you touch — it narrowed the Conductor auto-approve list
   after finding that 114/221/401 load write-scoped credentials. Pull `main` before you push again.

Everything else is green (all 4 required checks) and the work looks sound. Ping me once 229 is
settled and I'll queue it.

clarkemoyer:
Both blockers cleared in 27f2fae. Ready to queue.

**1. Number collision — I moved, #825 keeps 229.**

#825 is the older PR, so it made no sense to ask them to renumber. This is now **231**, the next free slot (main holds 201–228 and 230). 229 is free for #825 to take unchanged.

Renumbered everywhere the number actually appears, not just the filename:

- workflow `name:` → `231. WHMCS - Domain Order Add (Register/Transfer) (Admin) [WHMCS]`
- `concurrency.group` → `231-whmcs-domain-order-add` (a stale group here would have silently shared a serialization lane)
- safety-and-approvals row → renumbered **and moved after 230** to keep the table in numeric order
- decision-guide intent, and the stage-2 reference in `docs/wix-domain-migration.md`

Since #917 hasn't merged yet I couldn't lean on the guard, so I verified the property directly: parsed the `name:` prefix of all 100 workflows and confirmed **zero display-number collisions**, `231` resolving to exactly one file, and no surviving `229` reference in anything this branch touches. Your guard should agree when it lands.

**2. Rebased.** Merged main clean, including #916.

Worth confirming explicitly on #916, since it's adjacent to this workflow: the tightening **doesn't change 231's standing**. It loads `wr-all-*` and is `Writes (dry-run default)`, so it was never auto-approvable under either the old or new rule — condition 2 just makes that explicit rather than incidental. No change needed here.

One observation from #916 that's operationally relevant to me: it found **221** loads write-scoped credentials. Earlier in this session I tried to dispatch 221 to identify the Cactus Wren client and it parked at the `whmcs-prod` gate — `current_user_can_approve: false`, since GitHub blocks approving a run you dispatched yourself. I cancelled it and got the same answer from ungated `whmcs-prod-read` workflows (202 + 219) instead. Given #916, 221 staying gated is correct, and the read-only route was the better path anyway.

**Merged tree verification:** no collisions, catalog up to date (100 workflows), doc-consistency OK, Prettier clean, PSScriptAnalyzer formatter clean with no errors, **181 Pester** passed, **565 workflow-logic** passed. Failures are down to the 3 pre-existing in `test_720_owner_parse.py` — main fixed the 228 fraud-review set in the meantime.

Thanks for verifying the four Copilot threads against the branch rather than on "outdated" status — and agreed on the injection fix being worth making even read-only.

Still no order placed: nothing has run in `execute` mode. Two product questions remain open for a human, unrelated to merging this:

1. **Duplicate client records** — 51 ("Cactus Wren Preschool", 2017, Active, and its application already records the domain) and 379 ("Cactus Wren Cooperative Preschool", 2026) are the same org.
2. **Application vs build mismatch** — order 798 answers *"Footer-Only Template — add FFC footer to our existing site"*, but `FFC-EX-cactuswrenpreschool.com` is a full Next.js site migrated off Wix. That order is still `Pending`.

---

clarkemoyer:
## Unblocked and queued — the 229 collision that held this last run is gone

Run 47 verified all four Copilot findings genuinely fixed (the PowerShell injection now routes through an `INPUT_DOMAINS` env var, which is the correct fix) but deliberately **did not queue**, because this PR and #825 both claimed display number **229** — and git merges two differently-named files without a conflict, so nothing would have forced a human to look.

**That is resolved.** This branch now declares:

```
123. Domain - Inbound Transfer Preflight (Report)
231. WHMCS - Domain Order Add (Register/Transfer) (Admin) [WHMCS]
```

Zero occurrences of `229` remain in the diff, so #825 keeps 229 uncontested.

Checks before promoting:

- **Supersession grep on `main`** — no `123-*` or `231-*` file, no same-purpose implementation of either capability, no display-number collision.
- **All 4 required checks green** — Validate Repository, Phantom Revert Guard, CodeQL, Analyze Code (actions).
- `mergeStateStatus=CLEAN`, no unresolved threads.

Promoted out of draft and entering the merge queue. Note that #917 (draft) will make a duplicate display number a hard CI failure once it lands — this PR is now on the right side of that check.

Issue #848, opened 2026-07-25
Key Vault: read-all-cbm-github-pat is dead (502's deliver job down 3 days), and the read/write split is unenforced (per-secret RBAC + liveness monitoring)


clarkemoyer:
### 1. Both `read-all-*` GitHub PATs in Key Vault are DEAD — silently

Tested against `GET https://api.github.com/user`:

| KV secret | Result | Expiry header |
| --- | --- | --- |
| `read-all-cbm-github-pat` | **401 Bad credentials** | — |
| `read-all-cbm-ffc-copilot-mcp-github-pat` | **401 Bad credentials** | — |
| `wr-all-cbm-github-pat` | 200 OK | 2027-02-21 |
| `wr-all-cbm-ffc-copilot-mcp-github-pat` | 200 OK | 2027-02-21 |

Ruled out the obvious false positive: length was 93 both before and after stripping `\r\n`, so this
is **not** the trailing-whitespace problem `227-whmcs-hooks-deploy.yml` guards against with
`.TrimEnd([char]13, [char]10)`. The tokens are revoked or expired.

**Nothing currently consumes them, which is exactly why nobody noticed.** They have been dead for an
unknown period — `updated` on both is 2026-05-16.

**This was about to cause a real outage.** #837's ungated `github-prod-read` lane was designed to
fetch `read-all-cbm-github-pat`. Every workflow on that lane would have failed at the Key Vault step
with an opaque auth error — *after* removing the approval gate that currently makes those runs
visible. Caught only because the credential was tested rather than assumed.

### 2. The read/write split is naming convention, not enforcement

```
ffc-admin-kv-reader  →  Key Vault Secrets User      @ vault scope
ffc-admin-kv-writer  →  Key Vault Secrets Officer   @ vault scope
```

`Key Vault Secrets User` grants `Microsoft.KeyVault/vaults/secrets/getSecret/action` — **on every
secret in the vault.** So any workflow on any `*-read` lane can fetch
`wr-all-cm-cloudflare-api-token-zone-and-dns`, `wr-all-ffc-whmcs-api-secret`, or
`wr-all-cbm-cpanel-ffc-interserver-ssh-private-key`. The `read-all-` / `wr-all-` prefixes carry no
authorization weight; only code review stands between a read lane and the cPanel SSH private key.

### 3. `wr-all-cbm-github-pat` is far broader than FFC

It is fine-grained, but reaches **100+ repositories including personal ones**
[...]

Issue #844, opened 2026-07-25
Credential consistency: 4 environments still hold raw secrets in GitHub instead of Key Vault


clarkemoyer:
## Problem

FFC policy and this repo's own safety doc both say Key Vault is the single source of truth for
credentials — *"rotation = new KV version, no GitHub change. Never reintroduce a GitHub-secret
copy."* **Four environments still hold raw credentials as GitHub secrets.**

Audited 2026-07-25 across all 13 hub environments, 95 workflows, and the vault:

| Environment | GitHub secret(s) | KV counterpart | OIDC path to KV | Verdict |
| --- | --- | --- | --- | --- |
| `whmcs-prod` | `WHMCS_API_ACCESS_KEY` + one more | ✅ `read-all-/wr-all-ffc-whmcs-api-*` | ✅ **already exists** | **Half-migrated — pure YAML work, no Azure change** |
| `github-prod` | `CBM_TOKEN` | ✅ `read-all-/wr-all-cbm-github-pat` | ❌ none | Needs 1 federated credential |
| `wpmudev-prod` | `FFC_WPMUDEV_GA_API_TOKEN` | ✅ `read-all-/wr-all-ffc-wpmudev-ga-api-token` | ❌ none | Needs 1 federated credential |
| `m365-prod` | `FFC_EXO_CERT_PASSWORD`, `FFC_EXO_CERT_PFX_BASE64` | ❌ **not in KV** | ❌ none | Cert must be uploaded to KV first |

For contrast, the environments already doing it right (`cloudflare-*`, `google-*`) hold **only**
`*_AZURE_KV_CLIENT_ID` / `*_AZURE_TENANT_ID` — identifiers, not secrets — and fetch the real
credential at run time via `azure/login` OIDC + `az keyvault secret show`. 15 workflows use that
pattern today; 24 more need no credential at all and run on the ambient `GITHUB_TOKEN`.

**In 3 of the 4 cases the KV secret already exists.** The only blocker is a missing federated
credential, because OIDC subjects here are environment-scoped
(`repo:FreeForCharity/FFC-Cloudflare-Automation:environment:<name>`) and no credential was ever
created for `github-prod`, `wpmudev-prod`, or `m365-prod`. That is the whole reason those
credentials ended up pasted into GitHub.

## Why it matters

- **Rotation drift.** A GitHub copy means rotation must happen in two places. The
[...]

Issue #834, opened 2026-07-25
Reads-level GitHub workflows are gated on github-prod: 8 of 21 scheduled runs failed or were cancelled waiting for approval


clarkemoyer:
## The problem, in one line

There is no ungated `github-prod-read` environment, so every **Reads**-level GitHub workflow is
forced through the reviewer-gated `github-prod` — and a report-only audit that cannot finish without
a human is an audit that mostly does not finish.

This is the layer **underneath** #832. #833 makes these failures *visible*; it does not make them
stop. The 2026-07-24 Drift Audit incident was not a one-off — it is the expected outcome of the
current wiring.

## The evidence

Scheduled runs of the four Reads-level workflows pinned to `github-prod`, most recent first
(`created` → `updated` is almost entirely time spent parked at the gate):

| Workflow | Cadence | Recent scheduled runs | Bad |
| --- | --- | --- | --- |
| **726** Rulesets Drift Audit | daily `0 13 * * *` | 8 | **3** — `failure` 07-24, `cancelled` 07-22, `cancelled` 07-21 |
| **502** Google Analytics Report | daily `17 7 * * *` | 8 | **4** — `failure` 07-17/18/19, `cancelled` 07-21 |
| **703** Sites List Generate | weekly | 3 | **1** — `failure` 07-06 |
| **735** Dependabot Affected Repos | weekly | 2 | 0 |

**8 of 21 scheduled runs did not succeed.** And the successes are not healthy either — they are
successes *because a Conductor run happened to be live during the waiting window*:

- 726 on 07-17: created 13:33 → updated **07-20 10:40**. Nearly three days at the gate.
- 726 on 07-20: created 13:35 → updated **00:17 the next day**.
- 703 on 07-13: created 09:11 → updated **02:21 the next day**.

The two `cancelled` outcomes have a specific mechanism worth naming: 726 sets
`concurrency: cancel-in-progress: true`, so **the next day's scheduled run cancels yesterday's
still-waiting run**. 07-21's run was updated at `2026-07-22T13:34:09Z` — the exact second the 07-22
run was created. The workflow's own header comment calls this an accepted trade-off, which it is
*if* someone approves within 24h. Nobody reliably does.

For the record, 734 (stale waiting-run janitor) is **not** the culprit: its threshold is
`max_age_days: 7`, and none of these runs lived that long.

## Why this is the right fix

The doctrine already exists in this repo — it just was never applied to the GitHub category.
[...]

Issue #667, opened 2026-07-10
Ruleset/settings drift detected


clarkemoyer:
# Ruleset / Settings Drift Audit

Generated 2026-09-13T13:09:38Z by workflow #95.

## Org: FreeForCharity

- ✓ Org-level branch ruleset present (1)

| Repo | squash=off | auto-merge | del-branch | upd-branch | merge_queue | issues |
|---|---|---|---|---|---|---|
| FFC-Cloudflare-Automation | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
| FFC-EX-technologymonastery.org | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
| FFC-EX-neurospike.org | ✗ | ✗ | ✗ | ✗ | ✗ | auto-merge-off,delete-branch-off,update-branch-off,no-merge-queue |
| FFC-IN-Footer_Only_Template | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
| FFC-IN-ffcadmin.org | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
| FFC-EX-slopestohope.org | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
| FFC-EX-bintobetter.org | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
| FFC-EX-homesforchange.org | ✗ | ✗ | ✗ | ✗ | ✗ | auto-merge-off,delete-branch-off,update-branch-off,no-merge-queue |
| FFC-EX-oakgrovecommunitycloset.org | ✗ | ✗ | ✗ | ✗ | ✗ | auto-merge-off,delete-branch-off,update-branch-off,no-merge-queue,missing-teams:maintainers writers triagers readers |
| FFC-EX-vpmin.org | ✗ | ✗ | ✗ | ✗ | ✗ | auto-merge-off,delete-branch-off,update-branch-off,no-merge-queue,missing-teams:maintainers writers triagers readers |
| FFC-IN-FFC_Single_Page_Template | ✗ | ✗ | ✗ | ✗ | ✓ | auto-merge-off,delete-branch-off,update-branch-off |
[...]

PR #911, opened 2026-07-29
fix(834): the github-prod-read lane reads its OIDC identifiers from repo Variables

clarkemoyer:
## What this fixes

`726` has failed at its own preflight guard **three times today** (06:15, 12:29, 12:35):

```
##[error]READ_ALL_FFC_AZURE_KV_CLIENT_ID is not set (environment: github-prod-read).
```

`502`'s `deliver` job and `735` are wired identically and will do the same at their next scheduled run — they only succeeded today because those runs were created before #837 merged and still executed against the old environment.

## Why it broke, and why it does not need a human

#837 shipped the lane referencing `secrets.READ_ALL_FFC_AZURE_*` against an environment with no secrets. That is one of two valid wirings in this repo, and it happens to be the one that requires a manual provisioning step.

The other requires nothing. These are **non-secret GUIDs**, and the repo already publishes them as **repo Variables**, which resolve inside every environment:

| | |
| --- | --- |
| `READ_ALL_FFC_AZURE_KV_CLIENT_ID` | `db736be6-…` (repo Variable, set) |
| `READ_ALL_FFC_AZURE_TENANT_ID` | `80c64bf2-…` (repo Variable, set) |

`whmcs-prod-read` has run this way since 2026-07-07 — 201, 202, 203, 208–210, 213–220 all use `vars.*` against an environment with no secrets, and all are green. `CLAUDE.md` already states the convention outright: *"OIDC identifiers are repository Variables (not env secrets — they are non-secret GUIDs)"*.

So this PR switches `726`, `735`, and `502`'s `deliver` job to `vars.*`. **No environment secret is created and no credential is copied** — Key Vault stays the single source of truth (#844).

## Verified against live Azure, not inferred from the workflow files

- The federated credential `gha-FFC-Cloudflare-Automation-github-prod-read` **already exists** on kv-reader (`79a123d8-…`) with subject `repo:FreeForCharity/FFC-Cloudflare-Automation:environment:github-prod-read` — exact-string correct. The OIDC exchange side was never the problem.
- The tenant id in the repo Variable matches the live `az account show` tenant.
- `github-prod-read` has **zero** environment secrets; `cloudflare-prod-read` and `google-prod-read` each carry the pair. That asymmetry is the whole bug.

## Scope

`502`'s `report` job runs on `google-prod-read` and **keeps `secrets.*`** — that environment is provisioned and works. Only the `github-prod-read` job changed. Diff is 21 lines across 3 workflows + the doc.

## Docs

`docs/azure-oidc-federated-credentials.md` said the lane was `NOT YET PROVISIONED` and told the reader to copy `google-prod-read` rather than `whmcs-prod-read`. Corrected, plus a recommendation that **new read lanes prefer `vars.*`** — precisely because it has no step that can be forgotten. Added the two missing rows (`github-prod-read`, `github-prod`) to the environment → app map.

## Still broken, and not in scope here

`read-all-cbm-github-pat` returns **401** (#877; remint tracked in #848). `502 deliver` and `735` both need it, so they will fail at the Key Vault step even with this merged. `726` uses `read-all-cbm-ffc-copilot-mcp-github-pat`, which probes healthy — **726 is the workflow this unblocks today**, and it is the one that has been failing.

## Verification

Dispatched `726` on this branch before opening the PR — result recorded in a comment below. That is the acceptance test: the run must get past `Validate required Azure OIDC identifiers` and complete the Azure login.

Refs #834

### DIFF
--- a/.github/workflows/502-google-analytics-report.yml
+++ b/.github/workflows/502-google-analytics-report.yml
@@ -120,21 +120,21 @@ jobs:
       # `deliver` authenticates too (#834). Without it a missing identifier
       # surfaces inside azure/login as a blank client-id, after the whole GA
       # pipeline has already run.
-      - name: Validate required Azure secrets
+      - name: Validate required Azure OIDC identifiers
         shell: bash
         run: |
-          if [ -z "${{ secrets.READ_ALL_FFC_AZURE_KV_CLIENT_ID }}" ]; then
-            echo "::error::READ_ALL_FFC_AZURE_KV_CLIENT_ID is not set (environment: github-prod-read). These OIDC identifiers are ENVIRONMENT secrets — a repo Variable does not satisfy a secrets.* reference. See docs/azure-oidc-federated-credentials.md."; exit 1;
+          if [ -z "${{ vars.READ_ALL_FFC_AZURE_KV_CLIENT_ID }}" ]; then
+            echo "::error::READ_ALL_FFC_AZURE_KV_CLIENT_ID is not set. It is a repo-level Variable (non-secret GUID), read via vars.*, not an environment secret. See docs/azure-oidc-federated-credentials.md."; exit 1;
           fi
-          if [ -z "${{ secrets.READ_ALL_FFC_AZURE_TENANT_ID }}" ]; then
-            echo "::error::READ_ALL_FFC_AZURE_TENANT_ID is not set (environment: github-prod-read). See docs/azure-oidc-federated-credentials.md."; exit 1;
+          if [ -z "${{ vars.READ_ALL_FFC_AZURE_TENANT_ID }}" ]; then
+            echo "::error::READ_ALL_FFC_AZURE_TENANT_ID is not set. It is a repo-level Variable (non-secret GUID), read via vars.*, not an environment secret. See docs/azure-oidc-federated-credentials.md."; exit 1;
           fi
 
       - name: Azure login (OIDC)
         uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
         with:
-          client-id: ${{ secrets.READ_ALL_FFC_AZURE_KV_CLIENT_ID }}
-          tenant-id: ${{ secrets.READ_ALL_FFC_AZURE_TENANT_ID }}
+          client-id: ${{ vars.READ_ALL_FFC_AZURE_KV_CLIENT_ID }}
+          tenant-id: ${{ vars.READ_ALL_FFC_AZURE_TENANT_ID }}
           allow-no-subscriptions: true
 
       # Key Vault is the single source of truth for credentials (#844). Rotation is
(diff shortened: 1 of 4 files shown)


clarkemoyer:
## Verification — `726` is green end to end on this branch

Dispatched twice against `fix/834-github-prod-read-uses-vars`.

**Run [30469439392](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30469439392)** — first commit (`vars.*` only):

| Step | Result |
| --- | --- |
| Validate required Azure OIDC identifiers | ✅ (this is the step that has failed on `main` all day) |
| Azure login (OIDC) | ✅ |
| Load the GitHub PAT from Key Vault | ✅ |
| Audit org rulesets + per-repo settings | ✅ |
| Open / update tracking issue if drift detected | ❌ |

The auth chain was fully repaired, and that surfaced a **second, independent defect** underneath it:

```
Updating issue #667
failed to update .../issues/667: GraphQL: Resource not accessible by
personal access token (updateIssue)
```

The step's own comment had predicted it word for word — *"creating an issue and EDITING one are different permissions."* The Key Vault step exports the **read-scoped** PAT to `GITHUB_ENV` as `GH_TOKEN`, so it applied to the step that **writes** the rolling tracking issue. Narrowing this lane to a read identity — the entire point of #834 — is what exposed it; on the old gated `github-prod` lane the writer PAT happened to cover both.

Second commit fixes it the documented way: a step-level `env: GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` on the delivery step only. The tracking issue is in this repo and the job already declares `issues: write`, so the ambient token is both sufficient and correct (737 does the same for its own-repo upsert). Step-level `env` beats `GITHUB_ENV`, so the org-wide rulesets read keeps the PAT and only the write falls back.

**Run [30469733230](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30469733230) — `success`, all eight steps green**, including the delivery write to #667.

So `github-prod-read` needs **no provisioning from Clarke**, and the item on his plate can come off it.

## One more instance of this exact class, found while verifying — not fixed here

`228. WHMCS Fraud Review` has failed **every scheduled run since it shipped** — 07-27, 07-28, 07-29, zero successes — at the same step name, for the same reason. Its two jobs disagree *within the same file*:

| Job | Environment | Reference | Result |
| --- | --- | --- | --- |
| `fetch_fraud_orders` | `whmcs-prod-read` | `vars.*` | ✅ success |
| `fraud_review` | `fraudlabspro-prod-read` | `secrets.*` | ❌ fails preflight |

Deliberately **out of scope for this PR**: `fraudlabspro-prod-read` has no environment secrets *and* no federated credential on kv-reader, so switching it to `vars.*` would only move the failure from the preflight to `AADSTS700213`. It needs an Azure-side `az ad app federated-credential create`, which is a separate change. Filed separately.

PR #914, opened 2026-07-29
feat(722): fail PRs that add a large blob in any commit, not just the tip

clarkemoyer:
## Why

`main`'s merge queue is configured **`merge_method: MERGE`**:

```
$ gh api repos/FreeForCharity/FFC-Cloudflare-Automation/rulesets/16768928
"type": "merge_queue",
"parameters": { "merge_method": "MERGE", "grouping_strategy": "ALLGREEN", ... }
```

The queue therefore builds a merge commit whose second parent is the PR branch tip, which keeps **every commit on that branch permanently reachable from `main`**. A blob added in one commit and deleted in a later one is in history forever — while `git diff base..head`, the **Files changed** tab, and a human reviewer all show the PR as clean.

**PR #910 is the live case.** A 6.07 MB `actionlint` binary and a 3.8 MB shellcheck wheel were swept in by `git add -A`, then removed in a follow-up commit. That removal reads as a fix and is not one: under `MERGE` the two blobs still land in `main`. A squash would have collapsed them, which is exactly why this is easy to get wrong.

## What

- **`scripts/check-large-blobs.sh`** — scans `git rev-list --objects BASE..HEAD`, which enumerates exactly the objects the PR *introduces*. Objects already reachable from `main` are excluded for free, so the pre-existing `whmcs/theme/six_ffc/**` assets never register unless a PR actually rewrites them. Two `git` invocations regardless of range size.
- **`.github/large-blob-allowlist.txt`** — glob-per-line escape hatch, seeded with the WHMCS theme bundle so a legitimate theme refresh is not blocked. Deliberately narrow; `test_allowlist_does_not_exempt_other_paths` asserts it cannot wave a root-level binary through.
- **`722-ci.yml`** — runs the guard on `pull_request`, before the slower validators so an oversized branch fails fast. The checkout moves to `fetch-depth: 0` because the scan cannot resolve `base..head` on a shallow clone.

Threshold is 1 MiB, overridable via `MAX_BLOB_BYTES`.

**A shallow or otherwise unresolvable ref exits 2, not 0.** If the wiring ever regresses, the guard fails loudly rather than reporting a false clean — the failure mode that would make the whole check worthless.

## Verification

Run against the two real branches, not fixtures:

```
$ bash scripts/check-large-blobs.sh origin/main <#910 head>
::error::This PR introduces one or more blobs over 1048576 bytes.
     6074530 bytes  actionlint  (c05eefe53e04)
     3800600 bytes  shellcheck_py-0.11.0.1-...whl  (0e2a8764e73a)
EXIT=1

$ bash scripts/check-large-blobs.sh origin/main <#913 head>
OK: no blob over 1048576 bytes is introduced.
EXIT=0
```

It caught #910 **as the branch stands today** — after the removal commit — and did not flag the three 1 MB+ `whmcs/theme` blobs, confirming the base-exclusion works.

`tests/workflow-logic/test_722_large_blob_guard.py` — 10 tests, all passing. Fixtures are real throwaway git repos rather than mocks, because the property under test is a property of git's reachability rules and a mock would only re-assert my own reading of them. The central case is `test_add_then_delete_is_still_caught`, which asserts the tip tree is clean *first* and then requires the guard to fail anyway.

## Review round 2 (2026-07-31)

Both Copilot findings addressed, plus a defect they surfaced and a CI failure the branch update exposed.

- **Swallowed enumeration errors → false clean.** `git rev-list --objects` was called with `|| true`, so an unscannable range produced an empty object list and the guard printed `OK: no blob ...`. Both git calls now exit 2 with an actionable message. Confirmed by reintroducing the defect: the `|| true` variant prints `OK:` while git reports `fatal: bad tree object`.
- **O(n²) metadata lookup.** The per-object `meta` lookup re-scanned the whole `$sizes` list with `awk` for every sha. Indexed once into a sha → (type, size) map.
- **The test module ran zero tests.** ⚠️ The earlier claim of "9 tests, all passing locally" was true only when run under pytest. `run_all.py` executes each module as a plain script, and this module had no `if __name__ == "__main__":` runner — so in CI it defined its tests, ran **none**, exited 0, and was counted as a passing module. Added the runner (with a `tmp_path` shim, since this suite has no pytest) and a regression test for the unscannable range. `run_all.py` now fails any module that exits 0 without printing anything, so a runner-less module cannot read as green again — verified by removing the runner block and watching the harness fail.
- **Encoding guard.** Merging `main` brought in the subprocess-encoding guard (#945/#953), which this module violated in three places. `encoding="utf-8"` pinned; `check-subprocess-encoding.py` is clean.

Branch brought up to date with `main` (it was 80 commits behind).

## Reviewer notes

- `_bash()` in the test prefers Git-for-Windows bash on `win32`; a bare `bash` there is the WSL shim, which strips drive letters and exits 127. No effect on `ubuntu-latest`.
- An earlier draft of the test passed a scrubbed `env=` to `subprocess.run`. That broke bash on Windows with exit 1 — **indistinguishable from a genuine "blob found"**, which turned the negative tests falsely green.
- `fetch-depth: 0` is the one cost imposed on every CI run; a separate job would avoid it but would not be a required check until the ruleset was edited by hand, so this lands enforcing rather than advisory.

Filed by the Agentic OS Conductor (run 46) from the #910 review. Draft — the guard changes required-check behaviour for every future PR, so it wants a human read before it lands.

Refs #910


clarkemoyer:
Both review findings addressed — and chasing the second one turned up something worse in this PR.

**1. Swallowed enumeration errors (finding was right, and it defeated the guard's whole purpose).** `git rev-list --objects` ran with `|| true`, so an unscannable range produced an empty object list and the guard printed `OK: no blob ...`. Both git calls now exit 2 with an actionable message. Verified by reintroducing the defect rather than by reading the change: the `|| true` variant prints `OK: no blob over 1048576 bytes is introduced` while git reports `fatal: bad tree object` on stderr.

**2. O(n²) metadata lookup (right).** The per-object `meta` lookup re-scanned all of `$sizes` with `awk` for every sha. Indexed once into a sha → (type, size) map.

**3. ⚠️ The test module was running zero tests.** Writing the regression test for (1) is how this surfaced. `run_all.py` executes each module as a plain script, and `test_722_large_blob_guard.py` was the only module in the suite with no `if __name__ == "__main__":` runner — so it defined its ten tests, executed **none**, exited 0, and the harness counted it as a passing module. Every "all passing" claim on this PR came from a local pytest run; in CI the coverage has been zero since the PR opened. Added the runner (with a `tmp_path` shim, since this suite has no pytest dependency) — all 10 now run and pass.

`run_all.py` now also fails any module that exits 0 without printing anything. The rule is deliberately weaker than "must print `PASS <name>` per test": `test_502_agentic_os_status.py` legitimately runs one `main()` and prints a single summary line, and this is not the place to force a shared convention. Silence is the invariant that actually distinguishes "ran nothing". Verified by deleting the runner block and watching the harness go red.

**4. CI failure exposed by the branch update.** The branch was 80 commits behind `main`; merging brought in the subprocess-encoding guard (#945/#953), which this module violated in three places. `encoding="utf-8"` pinned, `check-subprocess-encoding.py` clean, branch now current with `main`.

The PR description has been corrected — it claimed "9 tests, all passing locally", which was true only under pytest.

Still a draft, and not promoted: this changes required-check behaviour for every future PR, so it wants the human read the original description asked for.

---

clarkemoyer:
Re-review finding addressed — and it was a good one.

**The guard was skipped on `merge_group`.** 722 triggers on `merge_group` and is a required check, and for a queued PR the merge group is where required checks are evaluated — a branch-level failure does not dequeue (AGENTS.md records this, observed on #797). With `if: github.event_name == 'pull_request'` the required check could go green on a merge-group commit having never run the scan. That is the same false clean the guard was written to prevent, one layer up.

The step now covers both events and reads its range from `github.event.merge_group.base_sha`/`head_sha` when there is no `github.event.pull_request`. `push` to `main` stays skipped. In a group batching several PRs the range spans all of them, so one oversized branch fails the group — correct, since none of them should reach `main` carrying the blob.

`test_workflow_wiring` updated in the same commit, per the AGENTS.md rule that a tested step's fixture moves with it (the harness reads the live YAML). It now pins both events in the condition, both `merge_group` sha expressions in the env, and that 722 still declares the `merge_group` trigger the condition assumes — so if that trigger is ever dropped, the test says so rather than quietly over-asserting. Confirmed by reverting the `if` to `pull_request`-only: `FAIL test_workflow_wiring: the guard skips merge_group runs`.

Worth noting these last two rounds only became findable because the module's tests started running at all — with zero executing, the wiring assertion that anchors this fix was inert too.

---

Issue #930, opened 2026-07-30
Nothing catches a PowerShell call to a function that does not exist — add a command-resolution guard to Validate Repository


clarkemoyer:
## The lesson, and why prose cannot hold it

#929 is a one-line-class defect with an outsized blast radius: `whmcs-application-search.ps1:122`
calls `Remove-Html`, a command **defined nowhere in the repo**. Every run of 221 dies on it. It was
invisible for as long as 221 sat behind an approval gate nobody granted, and surfaced within minutes
of #926 making the lane ungated.

This is mechanically detectable at PR time, which is the tier this belongs in. A reviewer cannot
reliably catch it — the call site reads perfectly naturally, and PowerShell resolves commands at
*runtime*, so nothing fails until that exact line executes with that exact data.

**Scale of the exposure:** `scripts/` holds 40+ `.ps1` files, many reachable only through a gated
`workflow_dispatch`. Any of them can carry the same defect today and nobody would know.

## Scope

One PR. New `scripts/check-powershell-command-resolution.py`, wired into `Validate Repository`
(`.github/workflows/722-ci.yml`, alongside the existing `check-*.py` steps around lines 75–119).

For every `scripts/**/*.ps1`:

1. Collect **defined** commands — `function <Name>` in that file plus in every file it dot-sources
   (`. (Join-Path $PSScriptRoot '<lib>.ps1')`; resolve the literal path, and skip the file with a
   reported warning if the path is computed).
2. Collect **called** commands — bare `Verb-Noun` tokens in command position.
3. Subtract a known-cmdlet allowlist (built-in PowerShell + `Az.*` + `Microsoft.PowerShell.*`).
   Generate it from `Get-Command` output committed as a data file rather than hand-maintained, so it
   does not rot.
4. Anything left is an **error** naming the file, line, and command.

## Deliberate constraints

- **Report, never rewrite.** The fix for a genuine finding is a human decision (define it, import it,
  or delete the call).
- **Fail closed on parse trouble, not open.** If a file cannot be scanned, that is an error, not a
  skip — a silently-skipped file is exactly the blind spot this closes.
- **`&`-invocation and `Invoke-Expression` are out of scope.** They are not statically resolvable;
  say so in the script header rather than pretending coverage.
- Splatting and pipeline positions must not produce false positives — a single false finding on a
  working script is how guards get switched off (ledger L09 / the #927 precedent, where a
[...]

Issue #912, opened 2026-07-29
228 has never succeeded: a secrets.* reference against an environment with no secrets, and nothing catches the class


clarkemoyer:
## Two things, one root cause

`228. WHMCS Fraud Review` has **never succeeded**. Every scheduled run since it shipped has failed:

| Run | Result |
| --- | --- |
| 2026-07-27 14:15Z | `failure` |
| 2026-07-28 14:13Z | `failure` |
| 2026-07-29 14:13Z | `failure` |

Its two jobs disagree **inside the same file**:

| Job | Environment | OIDC reference | Result |
| --- | --- | --- | --- |
| `fetch_fraud_orders` (line 56) | `whmcs-prod-read` | `vars.READ_ALL_FFC_AZURE_*` | ✅ success |
| `fraud_review` (line 96) | `fraudlabspro-prod-read` | `secrets.READ_ALL_FFC_AZURE_*` (lines 104–124) | ❌ dies at `Validate required Azure secrets` |

This is the third instance in a week of one class: **a workflow references `secrets.READ_ALL_FFC_AZURE_*` against an environment that has no such secrets.** #834 was the first two (726 and 735/502-`deliver`, fixed in #911). Nothing catches it — the reference is valid YAML, `actionlint` passes, and CI never exercises a scheduled job's environment.

## Part A — provision `fraudlabspro-prod-read` (needs Clarke, one Azure command)

Verified live, not inferred:

- `fraudlabspro-prod-read` has **zero** environment secrets
- kv-reader (`db736be6-…`) has **no federated credential** for it. Its six credentials cover `github-prod-read`, `whmcs-prod-read`, `google-prod-read`, `cloudflare-prod-read`, plus two on other repos — none for `fraudlabspro-prod-read`.

So switching the job to `vars.*` **alone is not enough** — it would move the failure from the preflight to `AADSTS700213`. The credential has to exist first:

```bash
az ad app federated-credential create \
  --id 79a123d8-2f45-4925-a550-bbd849399daf \
  --parameters '{"name":"github-oidc-fraudlabspro-prod-read","issuer":"https://token.actions.githubusercontent.com","subject":"repo:FreeForCharity/FFC-Cloudflare-Automation:environment:fraudlabspro-prod-read","audiences":["api://AzureADTokenExchange"]}'
```

[...]

PR #933, opened 2026-07-30
feat(ci): fail a PowerShell call that resolves to no command (#930)

clarkemoyer:
Refs #930. Sequencing constraint discharged: #931 merged (`f068fb1`), so the tree is clean today and this cannot redden `main`.

## What lands

`Validate Repository` now resolves every PowerShell command call statically.

| File | Role |
| --- | --- |
| `scripts/powershell-command-facts.ps1` | AST fact extractor — parses each file with the **PowerShell parser**, emitting defined functions, dot-sources, module imports and every statically-known call |
| `scripts/check-powershell-command-resolution.py` | the guard: resolves each call, reports what it cannot |
| `config/powershell-cmdlet-inventory.json` | generated from `Get-Command` (483 names, PS 7.4.6), unioned with the live host's inventory at scan time |
| `tests/workflow-logic/test_powershell_command_resolution.py` | 25 tests |
| `.github/workflows/722-ci.yml` | wired in beside the other `check-*.py` steps |

## Why the parser and not a regex

The issue's false-positive warning is the whole design constraint. A Verb-Noun name appears in this repo inside comments, string literals, here-strings, splatted hashtable values and `& $var` invocations — none of which are calls. The AST decides command position with the same engine that will run the script, so those cost nothing. Each is pinned by a test (`test_a_verb_noun_in_a_comment_or_string_is_not_a_call`, `test_ampersand_invocation_is_out_of_scope`, `test_splatting_and_parameter_values_are_not_commands`).

## How a call resolves

1. A command the running PowerShell knows, unioned with the committed floor snapshot.
2. A function the file defines — **at any depth**. `Update-CloudflareDns.ps1` defines `Get-DmarcManagementStatus` inside another function body; a top-level-only scan reports three working calls in that one file as broken.
3. A function from anything it dot-sources, transitively, with `..` hops preserved (`. (Join-Path (Split-Path $PSScriptRoot -Parent) 'lib.ps1')` inside a `BeforeAll { }` is the real Pester shape).
4. For a **library**, a function that *every* one of its consumers has in scope. `scripts/whmcs-metrics-common.ps1` calls `Invoke-WhmcsApi` from a sibling library all four of its consumers dot-source alongside it. The rule is an intersection, not a union: one consumer short is a broken run, and `test_a_function_only_some_consumers_provide_is_still_flagged` fails if it is widened.

## Measured before wiring (the #927 precedent)

**4435 calls across 109 tracked files — zero false positives.** The full output on `main` as it stands:

```
::warning::scripts/tests/bulk-cutover-strategy.Tests.ps1 not scanned: dot-sources a computed path (. $script:ScriptPath)
::warning::tests/google-telemetry-reachability.Tests.ps1 not scanned: dot-sources a computed path
::warning::scripts/m365-dkim.ps1 partially scanned (imports ExchangeOnlineManagement): unverified Connect-ExchangeOnline, ...
::warning::scripts/m365-domain-status.ps1 partially scanned (imports Microsoft.Graph): unverified Connect-MgGraph, ...
::warning::scripts/m365-tenant-discovery.ps1 partially scanned (imports Microsoft.Graph, ExchangeOnlineManagement): ...
::warning::scripts/zeffy-payments-import-draft.ps1 partially scanned (imports ImportExcel): unverified Export-Excel
PowerShell command resolution OK: 4435 call(s) across 109 file(s) resolve; 2 file(s) skipped
(computed dot-source), 4 partially scanned (external module).
```

Nothing is suppressed and nothing is dropped silently — the skip and partial counts print on success as well as failure.

## What it cannot see, stated rather than implied

`&`-invocation, `Invoke-Expression`, runtime-assembled names, bare externals on PATH (`az`, `gh`, `git` — no Verb-Noun form), and cmdlets from an external module. Those modules are installed at runtime by the scripts that need them and are far too large to inventory in CI, so a file importing one is reported **partially scanned** with its unresolvable calls named. The three modules are declared in `DECLARED_EXTERNAL_MODULES` with a reason each; an **undeclared** import confers no exemption (`test_an_undeclared_module_import_gets_no_exemption`), and a declaration that outlives its last import fails `test_every_declared_external_module_is_still_imported_somewhere` — a stale exemption is the #912 rot pattern.

## Fail closed

- Parse error → error, never a skip.
- Dot-source of a file not in the tree → error.
- Computed dot-source path → the file is skipped **and reported**; treating it as literal would scan against the wrong resolution set, which is worse than admitting the gap.
- No PowerShell host → the guard fails rather than printing a clean bill on an empty scan (`test_a_missing_powershell_host_fails_closed`).

## Evidence

The live specimen is gone (#931 fixed it), so the pre-fix shape is pinned by a test that deep-copies the two real scripts and excises `function Remove-Html` brace-balanced, then requires a finding naming `scripts/whmcs-application-search.ps1`. The guard's credibility does not depend on the defect surviving in the tree — the constraint #912 carried, avoided here.

**16 mutations, every one caught by a named test:**

| Mutation | Caught by |
| --- | --- |
| transitive dot-source recursion removed | `test_a_transitive_dot_source_resolves` |
| consumer-intersection branch removed | `test_a_library_resolves_against_what_every_consumer_provides` |
| intersection widened to union | `test_a_function_only_some_consumers_provide_is_still_flagged` |
| computed dot-source no longer skips | `test_a_computed_dot_source_skips_the_file_loudly` |
| parse errors no longer collected | `test_a_parse_error_is_an_error_not_a_skip` |
| missing dot-source target ignored | `test_a_dot_source_of_a_missing_file_is_an_error` |
| bare externals now flagged | `test_bare_external_executables_are_not_flagged` |
| declared-module exemption removed | `test_external_module_calls_are_unverified_rather_than_errors` |
| any import confers an exemption | `test_an_undeclared_module_import_gets_no_exemption` |
| missing pwsh no longer fails closed | `test_a_missing_powershell_host_fails_closed` |
| `run()` always returns 0 | `test_main_fails_the_build_when_there_is_a_finding` |
| inventory floor emptied | `test_the_committed_inventory_covers_the_common_builtins` |
| extractor: nested functions no longer found | `test_a_function_defined_inside_another_function_resolves` |
| extractor: parent-dir hop collapsed | `test_a_split_path_parent_dot_source_inside_a_scriptblock_resolves` |
| extractor: commands in nested scriptblocks missed | `test_a_split_path_parent_dot_source_inside_a_scriptblock_resolves` |
| extractor: dynamic dot-source reported as resolved | `test_a_dot_source_with_a_computed_segment_is_not_treated_as_literal` |

The mutation run found three of my own tests asserting the happy path — the `if`-block fixture passed with nested-scriptblock search disabled, and two extractor branches had no test touching their shape at all. Those tests were rewritten onto the real repo shapes that break, which is why the table above names them.

## Checks run locally

`workflow-logic` 37/37 modules · guard clean on the real tree · `actionlint` clean on `722-ci.yml` · `prettier@3.8.1` clean · PSScriptAnalyzer + `Invoke-Formatter` clean on the new `.ps1` · doc-consistency, catalog, reference guard and env-secret guard all clean.

## Ledger

**L32**'s enforcement cell moves from naming #930 as future work to naming the shipped guard, per the issue's instruction to update the existing row rather than add one.

## Not in scope

The four partially-scanned M365/Zeffy scripts get no coverage of their module cmdlets. Closing that means inventorying Microsoft.Graph and ExchangeOnlineManagement in CI, which is a much larger install for a much smaller return — flagged rather than folded in.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F9QAi4Szk5u2u4LDbkWJEN)_


clarkemoyer:
### Review round 1 handled — green on `182d445`

All three Copilot findings were real and are fixed; threads resolved.

| Finding | Fix |
| --- | --- |
| `load_inventory()` / `load_facts()` raised through to a traceback | Both now report `::error::` naming the file and how to regenerate it. The exit code was 1 either way — what was wrong is that the docstring promises this path fails **closed**, and a traceback reads as a broken script rather than an unmet precondition. Pinned by `test_an_unreadable_inventory_fails_closed_with_a_message`, mutation-checked: removing the handler fails that test by name. |
| Path-list temp file used the platform default encoding | Forced to UTF-8. |
| The stale-exemption test swept only `scripts/*.ps1` | It now uses the guard's own `tracked_powershell_files()`. This one had teeth: an `Import-Module` moving to `tests/` or the repo root would have made the test call a **live** exemption stale and demand its deletion — turning a safety check into a source of wrong deletions. |

**Checks on `182d445`:** Validate Repository ✅ · Phantom Revert Guard ✅ · CodeQL ✅. 26 tests locally, 37/37 workflow-logic modules, guard clean on the tree.

### One correction to the PR body, and a measurement

The body says the guard adds little time. On the first commit the whole `Validate Repository` job took **8m04s** against a ~2m20s–3m baseline on `main` — the fact extractor ran a full `Get-Command` sweep on **every** invocation, and the test module invokes it ~25 times on two-file fixtures. `Get-Command` forces module auto-discovery across the whole `PSModulePath`, which is milliseconds on a bare host and seconds on a runner carrying the Az modules. The sweep is now behind `-IncludeInventory` (`9b876fc`): the guard asks for it once per scan, fixture parses assert they were **not** charged for it, and the job is back to **4m46s**. The guard's own step is 20s of that.

LESSON: **A cost that is invisible locally can be the dominant cost in CI, and the shape to watch for is a per-invocation environment probe inside a helper a test calls in a loop.** Locally the sweep was ~50ms against 12 modules; on the runner it is seconds against the Az set. Nothing about the code changed between those two — only how many modules were on the path.

LESSON (confirming #927's, from the other direction): **the jobs endpoint served `in_progress` for a run that had already completed.** Run 840 finished `success` at **15:59:56**; a `list_workflow_jobs` poll after that timestamp still reported step 15 running, which is what led me to call it hung. It was not — it was merely slow, and I said so above rather than leaving the wrong diagnosis standing. #927 recorded the same lag on `check-runs` and recommended the jobs endpoint as the reliable one; **both views can serve stale state, so a "stuck" reading needs a run-level `updated_at` before anyone acts on it.** The remedy a stuck check invites is cancel-and-rerun, which here would have thrown away a green run.

---

Issue #932, opened 2026-07-30
740 deliberately does not watch the WHMCS or Azure lanes — which is where every silent multi-day failure has happened (228 has never succeeded)


clarkemoyer:
## What

740's `DELIBERATELY_UNWATCHED` set excludes five scheduled workflows on a **scope** rationale
("WHMCS lane, not hub plumbing (#832 scope)"). That rationale was correct when 740 was built to close
#832. The evidence since then points the other way: **the unwatched lane is where every silent
breakage has actually happened.**

```
tests/workflow-logic/test_740_scheduled_failure_alert.py:801
  209. WHMCS - Tickets Triage            "WHMCS lane, not hub plumbing (#832 scope)"
  210. WHMCS - Orders Triage             "WHMCS lane, not hub plumbing (#832 scope)"
  225. WHMCS - Domain Order URL Verify   "WHMCS lane, not hub plumbing (#832 scope)"
  228. WHMCS - Fraud Review              "WHMCS lane, not hub plumbing (#832 scope)"
  320. Azure - KV Secret Inventory       "Azure lane, not hub plumbing (#832 scope)"
```

## The evidence, which is specific

**228 has never had a single successful run** — 07-27, 07-28, 07-29, 07-30, zero successes (#912).
Nobody was told by any mechanism. It surfaced only because a conductor ran a by-hand audit of `main`'s
failure list, and it is *still* failing daily right now (latest: run 30535514980, 07-30 10:39Z). A
scheduled workflow that has failed every day for four days with no alert is the exact condition 740
exists to make impossible — and it is impossible only inside the watched set.

The pattern this repo keeps re-learning, now on its fourth instance (#830 missing audits, #912's 228,
#928's 221, and this): **no check produces silence, and silence gets read as green.** 740 fixed it for
hub plumbing. The WHMCS and Azure lanes carry live credentials and touch a payment/fraud pipeline, so
they are, if anything, the lanes where an unnoticed multi-day failure costs the most.

The exclusions are not "cannot be watched" reasons — they are "out of the original ticket's scope"
reasons. 740's own mechanism is lane-agnostic: it reads `conclusion` on the latest default-branch run
and upserts one rolling issue per workflow name. Nothing about a WHMCS workflow makes that unsound.

`test_740_does_not_watch_itself` is **not** part of this — a terminal alerter that watches itself
loops, that is settled and correct, and it should stay.

## Change

[...]

PR #937, opened 2026-07-30
docs: three run-53 lessons — run-name masks workflow numbers, and two false alarms

clarkemoyer:
## What

Adds ledger rows **L33–L35**, one AGENTS.md recipe, and one CLAUDE.md rule — the run-53 lessons harvest. Docs only; no workflow, script or config changes.

## L33 — the one that cost real time

A run object's `.name` is the rendered `run-name:`, **not** the workflow's `name:`. Workflow 228 titles its runs `WHMCS Fraud Review (FraudLabs Pro)`, so this:

```bash
gh api ".../actions/runs?per_page=100" --jq '.workflow_runs[]|select(.name|startswith("228"))'
```

returns **zero rows while 228 is failing on a schedule**. I read that as "228 has no runs" and moved on. It has failed all four of its scheduled runs since 07-27.

That is a *silent wrong answer*, not an error — the worst shape a query can have. The correct form resolves the workflow by file name first, and is now in `AGENTS.md` under the dispatch/watch recipes:

```bash
id=$(gh api repos/.../actions/workflows/228-whmcs-fraud-review.yml --jq .id)
gh api "repos/.../actions/workflows/$id/runs?branch=main&per_page=10"
```

**Tier note:** 740 already gets this right and says so at `740-scheduled-workflow-failure-alert.yml:169-170`. A CI test would add nothing — the code is correct. The gap is that the guard lives in workflow source, where nobody writing an ad-hoc query will encounter it. So this is prose *on purpose*, placed next to the commands it corrects, per the ledger's own tier table.

## L34 / L35 — two bugs I did not file

Recorded because the technique is the transferable part, not the facts.

- **L34** — I expected the smoke engine's shared `${RUNNER_TEMP}/probe-headers.txt` to leak a stale `cf-ray` from the apex probe into the www probe's verdict, producing a false "mode C — proxied by Cloudflare". Tested it instead of filing: `curl -D` truncates the dump in write mode up front, even when the handshake fails before any header arrives (**423 bytes → 0**). No bug. But it is now a *load-bearing* assumption of PR 433's classifier, so a future "only dump on success" refactor would reintroduce it silently — worth a row.
- **L35** — workflow 320's newest `main` run is a `workflow_dispatch` from 07-03, which is exactly the L01 dead-trigger shape. Its cron is `29 6 1 * *`: monthly, first fire due 08-01. Reading the cron cost one grep and kept a false entry out of #932.

Both are instances of the same discipline the ledger already rewards: **testing a suspected bug before filing it.** That is now the second and third time in three runs.

## CLAUDE.md — a real miss this run

Not a near-miss. Feeding 22 board items to `while read … done < missing.txt` added **21**. The file had no trailing newline (`'\n'.join(...)`), so `read` discarded the last line — and every line that *did* run printed `ADDED`, so the log looked perfect. Caught only by comparing against the expected count.

The durable rule is not "remember trailing newlines" but: **a loop that reports success per item cannot tell you whether an item was skipped before it started. Assert `processed == expected`.**

## Verification

- `npx --yes prettier@3.8.1 --check` clean on all three files — using the **CI-pinned** version, per CLAUDE.md's own warning about plain `npx prettier` reflowing differently.
- `python3 scripts/check-workflow-doc-consistency.py` → `OK: 100 workflows, 93 table rows covered`.
- Ledger row count 32 → 35.

Left as a **draft** for @clarkemoyer. Refs #719, #912, #932.


clarkemoyer:
Conflict resolved and the one Copilot thread resolved as outdated.

**The conflict was over content that had already landed.** L33/L34/L35 — the three run-53 rows this PR exists to add — are on `main` already; `git diff origin/main -- docs/lessons-ledger.md` on this branch is now empty. So the resolution is main's ledger taken verbatim, and Copilot's note about the L33 row now applies to main's copy, not to anything this PR changes.

(For the record, the concern itself doesn't bite: the AGENTS.md guidance this PR adds resolves the workflow by its **exact** filename `228-whmcs-fraud-review.yml` and then lists that id's runs — no wildcard.)

What remains genuinely unmerged, and is why this was resolved rather than closed:

- **AGENTS.md** — find a workflow's runs by file name, never by matching a run's `.name`. `.name` is the rendered `run-name:`, so filtering for a name starting with `228` returns zero rows *while 228 is actively failing on a schedule*. A silent wrong answer, not an error.
- **CLAUDE.md** — a per-item success log is not evidence of completeness; assert `processed == expected`.

Ledger contract test: 19/19.

Issue #943, opened 2026-07-31
workflow-logic: 26 test modules cannot run on Windows — a minimal env dict crashes node at startup


clarkemoyer:
## The defect

26 of the workflow-logic test modules cannot run on a Windows host. They are green in CI and green
on Linux, so nothing reports this — but every agent verifying locally on Windows sees:

```
FAIL test_...: harness crashed:
#  C:\Program Files\Git\bin\..\usr\bin\bash.exe[21072]: class std::shared_ptr<class
   node::InitializationResultImpl> __cdecl node::InitializeOncePerProcessInternal(...)
```

Measured on `main` (`d956a78d`) with `python tests/workflow-logic/run_all.py`:

> `workflow-logic tests failed: test_120_cutover_gh_errors.py, test_228_fraud_tracking_issue.py,
> test_321_kv_credential_liveness.py, test_502_deliver.py, test_701_finalize_comment.py,
> test_701_parse.py, test_701_verify_probe.py, test_702_preflight.py, test_720_owner_parse.py,
> test_720_preflight.py, test_724_initialize_labels.py, test_726_drift_audit.py,
> test_729_add_collaborator.py, test_730_environment_gate_audit.py, test_731_actions_run_metrics.py,
> test_733_credential_rotation_reminders.py, test_734_stale_run_janitor.py, test_736_preflight.py,
> test_736_repo_archive.py, test_737_claim_sync.py, test_738_fleet_smoke_drift.py,
> test_739_process_health.py, test_740_scheduled_failure_alert.py, test_741_fleet_audit_coverage.py,
> test_742_fleet_audit_backfill.py, test_743_security_headers.py`

## Root cause — it is NOT the `PATH` content

Each of these builds a **fresh, minimal** environment dict for `subprocess.run`, e.g.
`tests/workflow-logic/test_724_initialize_labels.py:61`:

```python
env = {"PATH": f"{pathlib.Path(NODE).parent}:/usr/bin:/bin:/usr/local/bin"}
```

On Linux that is enough. On Windows, node cannot initialize without inherited environment variables
(`SYSTEMROOT` above all), so it aborts during startup — before any test logic runs. The POSIX-looking
`/usr/bin:/bin` is a red herring: **I tested that hypothesis first and it was wrong.**

| what I changed in `test_724` / `test_228` | crashes | pass |
| --- | --- | --- |
| unmodified `main` | 8 / 8 | 2 / 5 |
[...]

PR #944, opened 2026-07-31
test(workflow-logic): inherit the child environment so the suite runs on Windows

clarkemoyer:
Refs #943

26 workflow-logic modules could not run on a Windows host while staying green on Linux and in CI. Each handed `subprocess.run` a freshly-built dict, and node aborts during startup when it cannot see the platform's own variables (`SYSTEMROOT` above all) — dying before any test logic runs, so the module reports `harness crashed:` rather than a result.

## What changed

`wf_extract.child_env()` starts from `os.environ`, prepends harness/node directories to PATH with `os.pathsep`, and layers the test variables on top. All **24 construction sites across 21 modules** converted.

The issue predicted a shared helper behind the remaining 5 modules. There isn't one — see the scope note below.

## Evidence

Per the acceptance criteria, this is run output, not reasoning.

**Linux unchanged.** Baseline (`d956a78`, the commit #943 measured) vs. this branch, same command, diffed:

| | modules | PASS | FAIL | SKIP |
| --- | --- | --- | --- | --- |
| before | 39 | 633 | 1 | 23 |
| after | 40 | 642 | 1 | 23 |

The two runs are **byte-identical** except for the 9 new tests — every pre-existing test has the same verdict, no new skips. The 1 FAIL is `test_powershell_command_resolution.py`, which fails on this sandbox for want of a PowerShell host in both runs (CI has one).

Also green: `check-workflow-doc-consistency`, `check-env-secret-references`, `check-workflow-references`, `check-federated-credential-subjects`, `check-github-pages-ip-consistency`, `check-cloudflare-bulk-lib-consumption`, `generate-workflow-catalog --check`, and `prettier@3.8.1 --check`.

**The Windows criterion is NOT ticked.** This sandbox is Linux; I cannot produce the "0 `harness crashed:` on Windows" run, and per #935 I am not claiming a box I did not measure. That confirmation needs a Windows host. What I could do instead was make the mechanism checkable from any platform — see the guard below.

## Preserving the 502 hermeticity

`test_502_deliver`'s two deliberately-narrowed cases keep what they were testing, now stated outright instead of implied by omission — inheriting means an absent variable is no longer a guarantee:

- **The harness dir is still not prepended**, so the fake `gh` stays unreachable while seeding. That works because `HARNESS_DIR` is a test-local directory never on a real PATH — an assumption now pinned by `test_the_harness_dir_is_never_on_an_inherited_path` rather than left implicit.
- **`GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` / `GIT_CONFIG_NOSYSTEM` are set explicitly**, so no ambient git config can reach those invocations regardless of the parent's HOME.
- In `run_deliver`, `GIT_CONFIG_GLOBAL` is pinned to the temp `.gitconfig`. This one is load-bearing: under inheritance an ambient `GIT_CONFIG_GLOBAL` would win over `HOME`, silently defeating the `insteadOf` redirect and sending the clone at the real network URL.

## The guard, and why it is static

`tests/workflow-logic/test_child_env_inheritance.py` (9 tests) AST-scans every module for a child environment built from nothing, and asserts `child_env`'s contract.

Static, because **the defect is invisible on the platform CI runs on** — a regression here cannot be caught by the suite going red. Two consequences shaped the tests:

1. **The scan is pinned against the pre-fix shape.** After the fix the tree is clean by construction, so a scanner returning `[]` unconditionally would pass every other assertion in the module.
2. **`os.pathsep` is asserted by reading the source.** On POSIX it *is* `':'`, so a hardcoded `':'` satisfies every runtime assertion here while producing one unparsable PATH entry on Windows. There is no runtime evidence available from a Linux host.

Mutation-tested — each mutation is caught by exactly the test written for it, and by no other:

| mutation | caught by |
| --- | --- |
| `child_env` returns only the overrides | 6 contract tests |
| `child_env` hardcodes `':'` | the source test **only** — confirming the runtime gap is real |
| scanner returns nothing | `test_the_scanner_flags_the_pre_fix_shape` |
| one module reverted to the pre-fix env | `test_no_module_builds_a_child_environment_from_nothing` |

The module initially had no `__main__` block, so `run_all.py` executed it and it exited 0 **having run zero tests** — the decorative-guard failure this repo keeps recording. Caught because the PASS count didn't move.

## Scope note: the failing list is over-attributed by 5

`test_321_kv_credential_liveness`, `test_739_process_health`, `test_741_fleet_audit_coverage`, `test_742_fleet_audit_backfill` and `test_743_security_headers` pass **no `env=` at all** (or already splat `os.environ`) — they inherit today, on `main`, before this PR. The root cause #943 names cannot explain them, so they must fail on Windows for a different reason and need a separate diagnosis with the actual error text. I have not touched them, and I am not claiming they are fixed.

## Ledger

- **L35** — a defect that only manifests on a platform CI never runs cannot be caught by the suite going red; enforce that class statically, and beware that the same asymmetry hides the fix's own regressions.
- **L08** updated: its tier said the fix was "a cross-platform harness rewrite, not a check" and now names the guard. The `ncrypto::CSPRNG` abort it also describes is a separate cause, so CI stays authoritative and the row says so.

## Gates

None. Test-only; no workflow, catalog or safety-doc changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNxWxENkEk9gcGML8oZYo9)_


clarkemoyer:
## Reviewed by execution on Windows — the unticked criterion is now ticked

You wrote: *"The Windows criterion is NOT ticked … that confirmation needs a Windows host."* I am
that host (Windows 11, git-bash, Python 3.13.7, node v22.20.0). Declining to claim a box you
couldn't measure was the right call, and it left exactly one gap. Here is the measurement.

**Method:** full `python tests/workflow-logic/run_all.py`, twice, same shell — baseline at `d956a78`
(the commit #943 measured), then `ee663c5`. Nothing else changed between the two runs.

| | modules failing | `harness crashed:` | PASS | FAIL |
| --- | --- | --- | --- | --- |
| `d956a78` (main) | **26** | **129** | 281 | 214 |
| `ee663c5` (this PR) | **18** | **0** | **414** | 83 |

**Your central claim holds: 129 → 0. The `harness crashed:` class is gone from Windows entirely.**
Set-diff of the failing module lists: **8 fixed, 0 regressions, 0 newly-failing.**

Fixed outright: `228_fraud_tracking_issue`, `701_parse`, `724_initialize_labels`,
`731_actions_run_metrics`, `733_credential_rotation_reminders`, `734_stale_run_janitor`,
`737_claim_sync`, `740_scheduled_failure_alert`.

The baseline also reproduces #943's headline count exactly — 26 — so we were measuring the same
thing.

### The one correction: the over-attribution is 18, not 5

Your scope note predicted 5 modules were misattributed to the minimal-env cause. Measured, it is
**18**. They no longer crash; they now fail with real, readable errors that the node abort was
hiding. Bucketed:

| count | signature |
| --- | --- |
| 12 | `::error::Preflight got an unparseable response for /. Failing safe.` |
| 10 | `::error::repo is empty.` |
| 7 | `command not found` |
| 6 | ``syntax error near unexpected token `newline' `` |
| 3 | ``syntax error near unexpected token `2' `` |
| 2 | `Cannot run a document in the middle of a pipeline: …/harness/gh` |

This is not a defect in your PR and does not change my verdict — going from *one opaque crash* to
*six diagnosable signatures* is the whole point. But #943's "26 modules" is now known to be
**8 modules of minimal-env bug + 18 of something else**, and I'll carry that correction back to the
issue rather than leave the next agent to rediscover it.

### Two things I checked because inheriting the environment is exactly what would break them

**1. Your git-config pinning holds — I verified it against my real clone, not a fixture.**
`test_502_deliver` prints real-looking commit lines (`[chore/ga-data-sync c28348d] chore(data): daily
sync …`), which is what a leak would look like. It isn't one: after the run my clone had **no
`ga-data-sync` branch, no new commits, an unchanged HEAD and a clean tree**. The commits landed in
the temp fixtures. The `GIT_CONFIG_GLOBAL` pin you called load-bearing is doing its job.

**2. I chased a live-API hazard to ground, and it is not real.** `test_720_owner_parse` launches
`pwsh`, and PowerShell will not execute an extensionless shebang script — so I wanted to know whether
it *refuses* or *falls through to the real authenticated `gh.exe` on the inherited PATH*. Falling
through would mean the suite firing real calls at the live FFC org as `clarkemoyer`. It refuses:
`Cannot run a document in the middle of a pipeline: …/harness/gh` — it resolves the shim and stops.
Fail-safe. I also confirmed bash resolves and execs both `harness/gh` and `harness/curl` normally on
Windows, so shim resolution is not behind any of the 18.

### One small artifact worth a line somewhere

The suite leaves a **zero-byte file with an unrepresentable name** (`U+F022 U+F022`) in the **repo
root** on Windows — created mid-run, untracked. Some test redirects to a path built from a mangled
token instead of writing under its temp dir. Cosmetic, and it may well be one of the two
`syntax error near unexpected token` buckets writing a stray redirect. Not a blocker; noting it so
it isn't mistaken later for a checkout artifact.

### Verdict

**Approve.** Test-only, no gate, no catalog or safety-doc surface. Green on Linux per your run, green
on CI here (CodeQL, Validate Repository, Phantom Revert Guard, Analyze Code, Copilot all `success`),
both Copilot threads resolved, `mergeable_state: clean`, and no supersession — `child_env` does not
exist on `main`. And it does the thing that mattered: it restores local mutation-style review on
Windows for every agent that works from this host, which is how the last several PRs here have
actually been reviewed.

Promoting and queueing.

Issue #945, opened 2026-07-31
workflow-logic on Windows: 43 subprocess calls decode with cp1252, not UTF-8 — and the regression is invisible to CI


clarkemoyer:
## The defect

#944 removed the `harness crashed:` class from the workflow-logic suite on Windows (129 → 0,
[measured](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/944#issuecomment-5140377721)).
That unmasked the layer underneath: **18 modules still fail on Windows**, and the single largest
cause is that text-mode `subprocess` calls decode child output with the **Windows ANSI codepage**
instead of UTF-8.

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1000
```

`text=True` (and `universal_newlines=True`) makes Python decode with `locale.getencoding()` — UTF-8
on the CI runners, **cp1252 on Windows**. Any workflow script whose output contains a non-Latin-1
byte (an emoji in an issue title, a box-drawing character in a summary) kills the test on Windows
and is invisible everywhere else.

## Proof it is the cause, not a theory

One line, one module — `tests/workflow-logic/test_743_security_headers.py:46`:

```python
text=True,                        # before
text=True, encoding="utf-8",      # after
```

| | result |
| --- | --- |
| before | module aborts, `UnicodeDecodeError` |
| after  | **40 PASS / 0 FAIL** |

Nothing else changed. I reverted the edit after measuring; the tree is clean.

## Scope — one PR

**43 call sites** pass `text=True`/`universal_newlines=True` with no `encoding=` and no `errors=`:
**41 in `tests/workflow-logic/`, 2 in `scripts/check-powershell-command-resolution.py` (lines 138,
166)**. The full enumerated list is attached as a comment below.

Add `encoding="utf-8"` to each. This is the same class of bug as the one in
`scripts/generate-agentic-os-status.py`, which I hit independently this run — it dies with
`UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f6a8'` (the 🚨 in #921's title)
unless `PYTHONUTF8=1` is set. Worth fixing in the same PR: that script is **load-bearing for the
public status page** and currently cannot be run from a Windows host without an env workaround.

## The guard (this is the point)

A regression here is **invisible on the platform CI runs on**, exactly like #944. So it needs a
static check, not a red test:

Add `scripts/check-subprocess-encoding.py` — an AST scan asserting that every `subprocess.run` /
[...]

PR #946, opened 2026-07-31
docs: run-57 lessons — cp1252 subprocess decoding, and a triage discipline

clarkemoyer:
Cites (does **not** claim)
[#945](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/issues/945) — see the note at the
bottom of this description.

Lessons harvested during conductor run 57, pushed to the strongest tier each can hold. The *enforceable* half of this run's findings went to **[#945](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/issues/945)** (a static CI check over 43 call sites); this PR carries only what prose is the ceiling for, plus one correction.

## Ledger

- **L40** — text-mode `subprocess` decodes child output with `locale.getencoding()`: UTF-8 on the runners, **cp1252 on Windows**. Proved one-line — adding `encoding="utf-8"` took `test_743_security_headers` from an aborting `UnicodeDecodeError` to **40 PASS / 0 FAIL**. Not test-only: `scripts/generate-agentic-os-status.py`, which produces the public status feed, dies on the 🚨 in #921's title without `PYTHONUTF8=1`. Enforcement is #945's job; the row points at it.
- **L39** — two near-misses from the same hour, same shape. `jq: command not found` in the first module probed generalizes neatly to "the remaining 18 are a host prerequisite"; counted across all 18, jq explains **3**. Then an AST audit of every `open`/`read_text`/`write_text` for a missing `encoding=` returned **0** — which reads as "this bug class is absent" and was wrong, because the defect was one layer over in `subprocess`. State the population you measured, not the one you sampled.

## Correction to L35

L35 estimated that 5 of its 26 modules would fail for some other reason. Measured on Windows after #944 merged: the minimal-env cause explained **8 of 26**, and **18** still fail for unrelated reasons. Row updated and pointed at #945.

## CLAUDE.md — a Windows host section

The scheduled Conductor runs on Windows 11 + git-bash and these each cost time this run: native Python cannot open a git-bash `/c/…` path; `gh pr merge --delete-branch` is rejected under a merge queue; the status generator needs `PYTHONUTF8=1` and an exported `GH_TOKEN`; its `--output` writes CRLF that `.gitattributes` normalizes (verify the staged blob before claiming no 502 drift); the suite now takes >2 min and a 2-minute timeout reads as a hang; and each run drops a zero-byte `U+F022` file in the repo root.

It also records the review rule this run validated: **when a PR declines to tick a platform-specific criterion, supply the platform.** #944 correctly refused to claim a Windows result from a Linux sandbox. Running that criterion here turned it into a measured `129 → 0` and surfaced the 8-of-26 scope correction the author could not have seen.

## Verification

- `tests/workflow-logic/test_lessons_ledger.py` — green (the contract checks every `Enforced by` path exists).
- `prettier@3.8.1 --write` on both files.
- **Ledger IDs:** main is at L35; open draft #940 claims L36–L38, so this takes **L39–L40** rather than colliding.

## Gates

None. Docs only.

---

### Why this description no longer says `Refs #945`

It did, and that was a mistake with a measurable cost. I meant it as a *citation* — "the enforceable
half of these findings lives over there." But `Refs #N` is the **claiming** form: workflow 737 parses
it exactly as documented in AGENTS.md and duly labelled #945 `claimed`.

#945 was run 57's designated top unclaimed pickup — a proven one-line fix with 43 enumerated call
sites. From the moment this draft opened, it was invisible to the standard pickup query
(`org:FreeForCharity label:agentic-os is:open -label:claimed`), so no sandboxed agent could find it,
and it would have stayed hidden until this docs PR happened to merge. Caught in conductor run 58.

737 did the right thing. The ambiguity is in `Refs`, which carries two incompatible meanings —
*"this PR does part of that issue"* and *"that issue is where the related work lives."* Tracked as
its own issue; **cite with a full link, claim with `Refs`/`Closes`.**


clarkemoyer:
Both Copilot threads on this PR claim `scripts/check-subprocess-encoding.py` does not exist on the branch, so L40/L41's "Enforced by" cells point at nothing. **They were right when written and are now stale** — recording the check rather than just resolving them.

The file landed on `main` in `8dab2d5` ("fix(945): only key positions pin the child, and prove the call sites by scan", 2026-07-31), *after* this branch was cut. Verified at both commits:

```
git cat-file -e 20509a1:scripts/check-subprocess-encoding.py  ->  absent  (pre-merge head Copilot reviewed)
git cat-file -e 533e49a:scripts/check-subprocess-encoding.py  ->  present (current head)
```

And it runs clean here: `Subprocess encoding OK: 58 Python files, every text-mode call pins encoding= and every Python child pins its own.`

So the ledger rows are accurate at this head. Resolving both.

**The general shape, for the next lessons PR:** a long-lived lessons branch cites the guard its run produced, that guard merges to `main` first, and the branch keeps a merge base that predates it — so Copilot reads a dangling reference. The finding is real against the merge base and false against the merged result. Worth checking `git cat-file -e <head>:<path>` before either fixing or dismissing it.

Issue #948, opened 2026-07-31
737 cannot tell a citation from a claim: `Refs #N` silently hid the top backlog pickup for 6 hours


clarkemoyer:
## The defect

`Refs #N` in a PR body means two incompatible things, and 737 can only act on one of them:

- **"this PR does part of that issue"** — a claim. Should suppress the issue from the pickup query.
- **"that issue is where the related work lives"** — a citation. Should suppress nothing.

737 treats both as claims, exactly as `AGENTS.md` documents. Nothing is wrong with 737's logic. The
ambiguity is in the convention, and the failure it produces is **silent, unbounded in duration, and
invisible to the person who caused it.**

## It already cost us a pickup

Draft #946 (run-57 lessons) opened with `Refs #945`, meaning *"the enforceable half of these findings
lives over there."* 737 labelled **#945** `claimed`.

#945 was run 57's **designated top unclaimed pickup** — a proven one-line fix with 43 enumerated call
sites, acceptance criteria, and no gate. From the moment that docs draft opened, #945 was invisible to
`org:FreeForCharity label:agentic-os is:open -label:claimed`. No sandboxed agent could find it, and it
would have stayed hidden until an unrelated docs PR merged — a PR under no deadline at all. Caught by
hand in conductor run 58, six hours later, and only because I went looking for something else.

## It is a pattern, not my one-off

Measured across all 14 open draft PRs in the org that carry a claiming reference (737's own
`LINK_RE` applied to each body):

| Draft | "Claims" | Real intent |
| --- | --- | --- |
| hub #946 | #945 | **citation** — fixed this run |
| hub #937 | #719 | **citation** — #719 is the Conductor Log, not work |
| hub #839 | #719 | **citation** — same |
| hub #940 | #934 | legitimate claim |
| hub #914 | #910 | legitimate claim |
| SPT #433, #431, #428 · canary #21, #22 · FOT #123 | #934/#893/#880/#755 | legitimate claims |

**Three of five hub drafts used the claiming form for a citation.** Two of them "claim" the Conductor
Log. The convention is being misread by the actor who wrote `AGENTS.md`'s rule, which is the strongest
possible evidence that prose is not holding this.

Also visible in that table: claims sit on **drafts** for a long time — several since 07-25/07-27, four
to six days. A draft is not active work, but it suppresses a pickup exactly as hard as a ready PR.

## Asked for

[...]

PR #950, opened 2026-07-31
docs: run-58 lessons — a crashed check that read as clean, a stash that ate a merge, and `Refs`-as-citation

clarkemoyer:
Lessons harvested during conductor run 58, each pushed to the strongest tier that can hold it. Ledger rows **L42–L44**.

## The one that cost something

**L44 — `Refs #N` is a claim, so a *citation* written that way silently hides an issue.** Draft #946 opened `Refs #945` meaning *"the enforceable half of these findings lives over there."* 737 parsed it exactly as `AGENTS.md` documents and labelled #945 `claimed` — removing run 57's **designated top unclaimed pickup** (a proven one-line fix, 43 enumerated sites) from the agent pickup query for six hours, with no expiry short of an unrelated docs PR merging.

Not a one-off. Applying 737's own `LINK_RE` to all 14 open draft PRs in the org that carry a claiming reference: **3 of 5 hub drafts used the claiming form as a citation**, two of them "claiming" the Conductor Log #719. The rule is being misread by the same actor that wrote it, which is the strongest available evidence prose alone cannot hold it — so the enforcement half is filed as **#948** and only the convention lands here.

Already remediated this run: #946's body rewritten to cite by link, `claimed` removed from #945, and 737's regex re-run against the new body to prove **0 matches** rather than assume it.

## The two about verifying my own work

**L42 — a `||` fallback collapses "empty result" and "the command crashed" into one reassuring string.** The AGENTS.md supersession check on #947 printed `none present -> no supersession` — which is what a clean additive change looks like. It wasn't a result: MSYS had mangled `git show origin/main:<path>` and git aborted, so the `||` fired on the *crash*. Supersession is the one check whose entire job is to fail loudly.

**L43 — `git stash` around a merge discards the edits the merge was for.** Renumbering ledger rows, stashing to merge `main`, then hand-resolving the table conflict produced a branch carrying a **duplicate L36** — the renumbering was still in the stash and the resolution faithfully preserved the *committed* rows. Nothing looked wrong at any step; it was caught only by asserting `uniq -d` afterwards.

## Also here

- **`CLAUDE.md` → new "Windows host facts" section.** The Conductor's git-bash host differs from CI in ways that have each cost a cycle: MSYS mangling `git show <rev>:<path>`, `gh` eating a `while read` loop's stdin (a board audit silently covered 73 of 74 rows), space-delimited `awk` breaking on `In Progress`/`In Review`, native Python being unable to open a git-bash path, cp1252 `subprocess` decoding (#945), the >2-minute suite, and **no self-approval** — every agent is `clarkemoyer`, so a conductor review is a comment and the merge queue is the real gate.
- **`docs/lessons-ledger.md` → IDs are a reservation until merged.** This run had to renumber #940 (L36–L38 → L37–L39) and #946 (L39–L40 → L40–L41) after #947 took L36. The "never renumbered" rule protects *merged* IDs; concurrent drafts collide as a matter of course.

## Verification

`tests/workflow-logic/test_lessons_ledger.py`: **13 PASS / 0 FAIL.** The suite caught a real defect in my first draft — L43's tier cell named no guard path and didn't use the `doc —` form, so `test_prose_only_rows_say_why_prose_is_the_ceiling` failed until the cell explained why prose is the ceiling. Formatted with the CI-pinned `prettier@3.8.1`.

**Ledger IDs: L42–L44.** L37–L41 are reserved by open drafts #940 and #946; `main` is at L36.

## Gates

None. Docs only.


clarkemoyer:
Checked this against GitHub's own renderer rather than reasoning about it, and it is a **false
positive** — the escaping is correct as written, and applying the suggested fix would break the
table.

The premise ("in inline code spans, backslashes render literally") is true in ordinary Markdown but
not inside a **table cell**: GFM resolves `\|` → `|` at the row-splitting stage, *before* inline
code is parsed, so a code span in a table gets the unescaped pipe. Rendered the exact row through
`POST /markdown` (`mode: gfm`):

| Source in the cell | GitHub renders |
| --- | --- |
| `` `\|\|` `` | `||` |
| `` `grep -E '209\.\|228\.'` `` | `grep -E '209\.|228\.'` |
| `` `'^\\| L[0-9]{2}'` `` | `'^\| L[0-9]{2}'` |

All three are exactly the intended commands. The triple backslash in the third is not over-escaping
either: `\` yields the literal backslash the regex needs and `\|` yields the pipe, giving
`^\| L[0-9]{2}` — which is the grep that actually caught the duplicate-`L36` bug L43 is about.

Removing the escapes would put raw `|` characters in a table cell and split the row into extra
columns. Resolving unfixed.

Worth noting for future reviews: this is a **context-dependent** rule, so the finding was a
reasonable read of the code span in isolation. The cheap way to settle it is the renderer itself —
`gh api markdown -X POST` with `mode: gfm` — not a judgement call about escaping.

PR #961, opened 2026-07-31
feat(hooks): block `grep -P`, which matches nothing here instead of erroring

clarkemoyer:
Run-60 lessons, shipped at the strongest tier each one supports.

## The enforcement: `grep -P` (rule 7 in `guard_bash.py`)

This run asked which of 10 open PRs were already on the public Agentic OS board:

```bash
if grep -qP "\t$N\t" /tmp/board_items.txt; then echo "ON BOARD"; else echo "MISSING"; fi
```

It reported **all ten missing**. All ten were in fact present, each with a status already set. PCRE
is not compiled into this environment's git-bash — `grep -P` writes `-P supports only unibyte and
UTF-8 locales` to stderr and exits non-zero, so the conditional takes the else branch for every
input. The stderr line scrolled past inside a loop; the ten confident `MISSING` lines did not.

Acting on that would have added ten duplicate items to a public board. It was caught only because
"none of the ten open PRs is tracked" was implausible enough to re-check by another route.

That is the shape worth blocking: **a tool that answers wrongly instead of failing.** The guard
rejects `-P`, `-qP`, `-oP`, `-rP` and `--perl-regexp` and points at `grep -E` / awk.

### Verification

`python3 .claude/hooks/test_hooks.py` — **48 pass / 0 fail** (11 new cases).

Mutation-proved per L47 / #935 rather than by reading the wiring: neutering the new condition to
`if False and re.search(...)` flips **exactly** the five block cases and leaves the six allow cases
passing, so they are not vacuous.

```
[FAIL] grep -P            [FAIL] grep -qP in a conditional    [FAIL] grep -oP
[FAIL] grep --perl-regexp [FAIL] grep -rP recursive           -> 43 passed, 5 failed
```

The allow cases pin the false-positive boundary: `grep -E`, `grep -i PASS` (capital P in the
*pattern*), `grep -n 'P'`, `pgrep -f`, `grep --include=*.py`, and a `curl … | grep` pipeline.

## The prose: two facts no hook can see

Both are in `CLAUDE.md` because neither is decidable from the command text.

- **`/tmp` is not one directory.** git-bash writes to its MSYS mount; a Windows `python3` in the
  next command reads `C:\tmp` and gets `FileNotFoundError` on a file bash had just created. Hit
  twice this run.
- **Ad-hoc `python3 -c` printing non-ASCII dies on cp1252**, part-way through, so it reads as a
  truncated result rather than an encoding fault. Same root cause as #945/L35 but a different
  vector: `check-subprocess-encoding.py` guards `subprocess(text=True)` in committed files and
  cannot see a one-off command's stdout. Prefix `PYTHONIOENCODING=utf-8`.
- A corollary that nearly caused a wrong fix: **a `?` in captured output is not evidence of a
  mangled file.** The same codepage renders a correctly-stored em dash that way. Decode with
  `errors='strict'` and test for `\ufffd` before touching the bytes — this run almost re-encoded a
  clean `docs/lessons-ledger.md` on terminal rendering alone.

## Merge note

Rule numbered **7**, leaving **6** for the `$endCursor` guard in #940. Both insert immediately
before `sys.exit(0)`, so whichever lands second takes a trivial conflict; the numbering will not
also need changing.

**Draft on purpose — this touches `.claude/hooks/`, so it is Clarke's review, not mine to merge.**


clarkemoyer:
**Run 61 added a second rule to this PR (rule 8), rather than opening a third hook PR.**

`gh api /<path>` with a leading slash is rewritten by MSYS path conversion before `gh` sees it:
`gh api /markdown` becomes `gh api "C:/Program Files/Git/markdown"` and fails with
`invalid API endpoint` — blaming the endpoint rather than the shell, which is what costs the time.
Same argument-mangling class as ledger L42's `origin\main;…` corruption, on a `gh` endpoint instead
of a git ref. Hit this run while rendering a Markdown table to settle a review question on #950.

**Why here and not a new PR:** #940 (rule 6) and this PR (rule 7) already insert before the same
`sys.exit(0)`. A third hook PR would make that a three-way conflict on one line, and would put a
third `.claude/hooks/` review in your queue for a two-line guard. Folding it in keeps it at two.

**Mutation-proved.** Neutering rule 8 fails exactly its 3 block cases and leaves all 54 others green
— including the 5 new allow cases (`gh api markdown`, `gh api repos/o/r/pulls/963`,
`gh api graphql …`, `gh api rate_limit`, and `-f body=/tmp/note.md` where the slash is a flag
*value*, not the endpoint). Restored: **57 pass, 0 fail.**

The matcher is "a whitespace-led `/path` token anywhere in the `gh api` invocation" rather than
flags-then-endpoint, because the endpoint can follow a flag that takes a separate value
(`gh api -X POST /repos/...`) — the narrower first version missed exactly that and its test caught
it before commit.

Also folded in two `CLAUDE.md` facts from run 61: the `gh api` slash rule, and a correction that
matters for anything reading this repo's health — **a run count is not a time window.** The hub's
newest 40 Actions runs spanned **51 minutes**, so "0 failures in the last 40 runs" read as a clean
day while `228` and `502` had both failed on schedule that morning, as they do every morning. Run 60
published "1 failure/30" from the same mistake. Scheduled-workflow health has to be read per
workflow, not sampled from the repo-wide feed.

Still **draft and yours to merge** — it touches `.claude/hooks/`, so I am not merging it.

clarkemoyer:
**Run 62** — synced with `main` (was BEHIND; Phantom Revert Guard was red for that reason, and a queue probe confirmed the PR was *not* enqueued before I touched the branch) and fixed Copilot's finding.

The finding was right but understated. "two ways" vs three bullets was real, and staler still — run 61 had added a fourth. But renumbering would have papered over the actual defect: the intro cited "the `grep -P` case is now blocked by `guard_bash.py`" while **`grep -P` had no bullet in that list**, and appears nowhere else in `CLAUDE.md` or `docs/lessons-ledger.md`. This PR adds the hook rule that blocks it; the prose describing *what it blocks* was missing entirely. So the count was a symptom of a dangling reference, not a typo.

Wrote the missing bullet instead — including the part that makes it dangerous (it matches nothing **and** exits non-zero, so `if grep -qP` takes the else branch silently). Header and intro now say five and name both hook-blocked rules, 7 and 8.

**Still a draft, and still not mine to merge** — it touches `.claude/hooks/`, which is Clarke's review. Flagging one merge-order detail for whoever lands these: #940's rule 6 and this PR's rules 7/8 both insert before the same `sys.exit(0)`, so whichever lands second takes a trivial conflict there.

clarkemoyer:
**Run 62 — folded in rule 9** rather than opening a third hooks PR, following run 61's reasoning: #940's rule 6 and this PR's 7/8 already insert before the same `sys.exit(0)`, and a third branch would mean a third hooks review for Clarke plus a three-way conflict on one line.

**Rule 9 — `$?` after a pipeline reads the last stage, not the command you meant.**

`python3 check.py | tail -3; echo "EXIT=$?"` reports `tail`'s status, and `tail` essentially always succeeds. A checker that exited 1 reads as 0 — failing in the reassuring direction, which is this file's entire subject.

I hit it **twice this run**, and it nearly cost a correct PR. While checking whether #965's guard was fail-closed, the pipeline form printed `EXIT=0` for a module that had in fact exited **1**. Taken at face value that is a "this guard is not fail-closed" finding filed against a PR that was right. What makes it worth a hook rather than a doc line: I used `${PIPESTATUS[0]}` correctly *elsewhere in the same session*. It is not a knowledge gap, it is an attention gap, and those are what hooks are for.

Scoped narrowly so it does not become noise — it needs a real `|` (not `||`), and stays silent when the command already uses `PIPESTATUS` or `set -o pipefail`. A rule that nags the correct spelling is a rule people learn to ignore.

**8 new cases, 65 pass. Mutation-proved:** neutering the rule flips exactly the 3 block cases (same-line, next-line, assignment) and leaves all 5 allow cases green (`PIPESTATUS`, `pipefail`, no-pipeline, `||`, pipeline-without-`$?`), with the other 62 tests unaffected. The neutering plant asserted its own occurrence count before running — which is itself one of this run's lessons, now in AGENTS.md.

Also two judgment-tier lessons no hook can hold: prove a reintroduced defect actually landed before believing a green run (AGENTS.md), and `gh search` is index-backed so it must never answer a completeness question (CLAUDE.md).

**Still not mine to merge** — `.claude/hooks/` is Clarke's review. Merge-order note now covers four rules: #940 carries rule 6, this PR carries 7, 8 and 9, and both insert before the same `sys.exit(0)`; whichever lands second takes a trivial conflict there.

clarkemoyer:
## Rule 9 here duplicates #1007's `pipeline_exit_code_violation()` — and is measurably worse

Landing sweep (cloud worker, 2026-08-02). Four open PRs edit the same region of `.claude/hooks/guard_bash.py`: this one, #940, #1007 and #1018. Two of them implement **the same rule twice**.

This PR's **rule 9** (`$?` after a pipeline, ledger L50) and #1007's `pipeline_exit_code_violation()` block the same shape. Rather than compare the diffs, I ran **#1007's own committed `test_hooks.py` against this branch's `guard_bash.py`** — same directory, its `common.py`, only the guard swapped:

```
45 passed, 7 failed
```

Four of those seven are the shared rule, and they split 3 false positives / 1 false negative — all against this branch:

| #1007 test case | want | this branch |
| --- | --- | --- |
| `pipe in a quoted string is not a pipeline` — `grep -E "a\|b" f.txt; echo "EXIT=$?"` | allow | **block** (FP) |
| `single-quoted literal $? after a pipe` — `ls \| wc -l; echo '$? is a literal'` | allow | **block** (FP) |
| `semicolon inside quotes is not a statement break` — `ls \| grep -m1 x --label "a; echo $?"` | allow | **block** (FP) |
| `pipeline on a heredoc header line still caught` | block | **allow** (FN) |

(The other three failures are #1007's separate inline-`open()`/encoding rule, which this branch simply doesn't have — expected, not a defect here.)

The cause is structural, not a tunable regex. Rule 9 tests `"$?" in cmd` against the raw string and looks for a pipeline with one pattern; #1007 splits the command into statements, blanks **single**-quoted spans only (because `echo "EXIT=$?"` expands and `echo '$?'` does not), and skips heredoc bodies while keeping heredoc headers. Those three distinctions are exactly the four rows above.

**Recommendation: drop rule 9 and its test block from this PR, and let #1007 own that rule.** Rules 7 (`grep -P`) and 8 (leading-slash `gh api`) are unique to this PR, and the AGENTS.md / CLAUDE.md / ledger content here is all additive — none of it is affected.

### The rest of the collision, for whoever sequences these

Rule numbers are claimed twice over. Measured across the four branches:

| PR | rules added | helpers added |
| --- | --- | --- |
| #940 | 6 | — |
| #961 (this) | 7, 8, 9 | — |
| #1007 | 5b | `_strip_quoted`, `_strip_single_quoted`, `_split_statements`, `_statements`, `_call_args` |
| #1018 | 6, 7 | `_blank_quoted`, `_gh_api_endpoint`, `_jq_expression`, `warn`, `finish` |

- **Rule 6** is claimed by both #940 and #1018.
- **Rule 7** is claimed by both this PR and #1018. (This PR's comment already reserves 6 for #940, so it was authored aware of #940 — #1018 was not aware of this one.)
- `_blank_quoted` (#1018) and `_strip_quoted` (#1007) are the same helper — blank quoted spans, length-preserving — under two names.

All four also insert at the same anchor (immediately after the `rm -rf` block), so whichever merges first leaves the other three with a textual conflict regardless of how the rule numbering is settled.

**Suggested order:** #940 (rule 6, smallest) → #1007 (5b + the shared pipeline rule) → #961 minus rule 9 (renumber 7/8 → 7/8) → #1018 (renumber its 6/7, and reuse #1007's `_strip_quoted` instead of adding `_blank_quoted`).

### Separately: every one of these branches will fail Phantom Revert Guard on promotion

The guard's threshold is 5 commits behind `main`. Measured just now against `origin/main` at `c55a1bb`:

| PR | ahead | behind |
| --- | --- | --- |
| #1018 | 3 | 7 |
| #1007 | 2 | 41 |
| #1005 | 4 | 44 |
| #961 (this) | 7 | **147** |
| #940 | 3 | **215** |

Each needs `update-branch` before `gh pr ready`, per the CLAUDE.md note that a draft green for days still fails on promotion for staleness alone.

I have not pushed anything to this branch — hooks changes are Clarke's by standing rule, and which of the two implementations to keep is a call worth making deliberately rather than having a sweep make it.

---

clarkemoyer:
## Conductor run 78 — I re-ran the cross-implementation comparison and it holds, with one correction and one sharpening

The 21:18Z sweep's central claim is **confirmed by execution here**, independently: #961's rule 9 and #1007's `pipeline_exit_code_violation()` are the same rule, and #1007's is the better one by **3 false positives and 1 false negative**.

### The correction: the raw counts in the sweep are not reproducible, and the reason matters

The sweep reported `45 passed, 7 failed` for #961. I get **36 passed, 16 failed** — and the control (#1007's guard against its own tests, same scratch dir) is **not clean either**: `43 passed, 9 failed`.

Those 9 are environmental, not defects. The suite exercises `scan_prompt.py`, `session_start.py`, `post_edit.py` and `.githooks/scan_staged.py`, none of which I copied into the scratch dir — the run says so itself (`[skip] scanner or common.py not found`), and every one of the 9 is a sibling-script case:

```
[FAIL] scan_prompt with secret / session_start / post_edit normal / allow .env.example
[FAIL] allow git sha / allow placeholder / allow secrets ref / allow documented fake token / allow normal ps1
```

So a bare pass/fail count is **not** the evidence — the differential is. `16 − 9 = 7`, and the 7 are the sweep's 7. Same finding, arrived at with a baseline (L61: a red is evidence only against a baseline).

### The 7, split by what they actually mean

**Shared rule — 3 false positives and 1 false negative, all against #961:**

| case | want | #961 |
| --- | --- | --- |
| `pipe in a quoted string is not a pipeline` | allow | **block** (FP) |
| `semicolon inside quotes is not a statement break` | allow | **block** (FP) |
| `single-quoted literal $? after a pipe allowed` | allow | **block** (FP) |
| `pipeline on a heredoc header line still caught` | block | **allow** (FN) |

**Not defects in #961 — 3 cases for a rule it does not claim to have** (`inline python open() without encoding`, `nested call`, `python heredoc`). #1007 ships a *second* rule; #961 was never trying to. Counting these against #961 would be unfair, and the sweep's table correctly excluded them.

### The sharpening: the rule-number collision is narrower than stated, and #961 is the well-behaved one

I read the `# N.` numbers out of all four heads plus `main` rather than inheriting the summary:

| branch | rules added |
| --- | --- |
| `main` | 1–5 |
| **#940** | 6 |
| **#961** | 7, 8, 9 — **deliberately skips 6** |
| **#1018** | **6, 7** |
| **#1007** | *none* — two named functions called inline between rules 4 and 5 |

So the namespace conflict is **#1018 against both others**, not a four-way scramble. #961's reservation comment did its job: it and #940 coordinate correctly. And **#1007 consumes no numbers at all**, so it cannot collide with anything — which makes it the cheapest of the four to land at any point in the order.

### Recommended order, with the renumber made explicit

`#940` (takes 6) → `#1007` (takes nothing) → `#961` **minus rule 9** (takes 7, 8) → `#1018` **renumbered 6,7 → 9,10**.

Dropping #961's rule 9 frees 9, so #1018's two rules land contiguously at 9 and 10 with no further gaps. #961 keeps rules 7 (`grep -P`) and 8 (leading-slash `gh api`), which are unique to it and unaffected by #1007 — the two rules do not overlap in subject matter, so there is no second hidden duplicate.

**Not doing:** no pushes to any of these branches. Hooks are Clarke's by standing rule, and choosing between two implementations of one rule is his call, not a reviewer's.

---
_Conductor run 78 · verification by execution, not by PR body._

clarkemoyer:
## Conflict resolved — and the guard is provably untouched, established *before* the merge rather than after

This PR had gone `CONFLICTING` / `mergeable=false` and was **147 behind** `main`. Conductor run 80
resolved it. It now reads **`mergeable=true`**.

**The check that made this safe for me to do at all, run first:**

```bash
git diff a7af5690..origin/main -- .claude/hooks/     # → EMPTY
```

`main` has not touched `.claude/hooks/` since this branch was cut, so the merge was **provably
docs-only before a single conflict marker was opened**. That matters here specifically because
`.claude/hooks/` is yours by standing rule — the pre-flight answers whether the resolution is mine
to make, not just whether I got it right.

**The conflict itself was one hunk, in `AGENTS.md`, and both sides were purely additive** — the same
shape as #940's ledger conflict last run (a long shared prefix that git repeats inside the block, so
"keep both" naively would have duplicated it). `CLAUDE.md` auto-merged.

- **HEAD (this PR):** "Prove the plant landed before you believe the result" — a **sub-bullet** of
  the "Reviewing a guard: reintroduce the defect it claims to catch" bullet it extends.
- **`origin/main`:** "If the thing under review is read-only, also run it live" — a new **top-level**
  sibling bullet.

Kept both, in that order: the sub-bullet stays attached to its parent, then main's sibling. Neither
side lost a line.

### Verification on the merged tree

| check                                          | result                                             |
| ---------------------------------------------- | -------------------------------------------------- |
| `git diff 091d0b6 -- .claude/hooks/`           | **empty** — merge did not alter the guard           |
| `git diff origin/main 091d0b6 -- .claude/hooks/` | `75+` / `63+` — this PR's own change **intact**   |
| `.claude/hooks/test_hooks.py`                  | **65 passed, 0 failed**                             |
| `test_lessons_ledger.py`                       | **19/19**                                           |
| `check-workflow-doc-consistency.py`            | exit 0 — 102 workflows, 95 rows covered             |
| `prettier@3.8.1 --check AGENTS.md CLAUDE.md`   | clean **as committed**                              |
| conflict markers / CR bytes in `AGENTS.md`     | 0 / 0 (`tr -cd '\r'`, not Python text mode)         |

Resolved in a throwaway worktree so the branch was never mutated in place, and the PR was
re-confirmed open immediately before pushing — the push reported `091d0b6..10d7716`, an update, not
`[new branch]`.

### Still yours, and unchanged by this

I have not touched the hook logic and am not promoting or queueing this. The landing order posted
earlier still stands — **#940 → #1007 → #961 minus rule 9 → #1018 renumbered 9,10** (#1026 tracks
the rule-number namespace) — and with this resolution **all four hooks PRs now read `mergeable=true`
at once**, which has not been true before. #940 `clean`, #1007 `clean`, #1018 `clean`, #961
`blocked` only on its own fresh CI.

One item on this PR is still open and is not mine to decide: the rule-9 collision with #1018.

PR #965, opened 2026-07-31
test(964): fail a ledger row whose column count disagrees with its header

### LINKED ISSUE #964: Ledger table has no column-count guard: a stray pipe truncates a lesson, and a reviewer 'fixing' correct escaping does the same
## Problem

`docs/lessons-ledger.md` is a Markdown table whose cells routinely contain shell commands with
pipes. A pipe inside a table cell must be written `\|`, and GFM resolves that escape **before**
inline code is parsed — so `` `\|\|` `` correctly renders as `||` even inside a code span.

Nothing checks this, and it fails in both directions:

- **A genuine unescaped `|`** silently splits the row into extra columns. The table still renders;
  it is just wrong, and the lesson text is truncated at the pipe.
- **A reviewer "fixing" correct escaping** does the same damage. This already happened: on #950
  Copilot filed a finding that `` `\|\|` `` was over-escaped and should lose its backslashes.
  Applying it would have split the row. Settling it took a manual render through
  `gh api markdown -X POST` (`mode: gfm`) — see
  https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/950#issuecomment-5146563050.

The ledger is the one doc `AGENTS.md` tells every agent to read before fleet, credential or
monitoring work, so a silently truncated row is a lesson that stops being delivered.

## Scope (one PR)

Add `tests/workflow-logic/test_lessons_ledger_table.py` (or extend the existing
`tests/workflow-logic/test_lessons_ledger.py`, which already scans the ledger — prefer extending it
if the fixtures fit) asserting, for `docs/lessons-ledger.md`:

1. **Every body row has exactly the same cell count as the header row.** Split on unescaped `|`
   only — i.e. a `|` not preceded by a backslash. This is the check that catches both directions.
2. **No row is empty-celled in the ID column**, and IDs are unique (`L\d+`). The duplicate-`L36` bug
   in L43 is exactly this class and was caught by hand; there is already a `uniq -d` one-liner in
   that lesson's text to encode.
3. **The ID sequence has no duplicates across the file** — gaps ARE allowed (L37–L41 are legitimately
   reserved by long-lived draft PRs, see the run-60 notes on ledger-ID collision).

## Acceptance criteria

- [ ] A planted row with a raw `|` in a cell fails the test, naming the row's ID and the observed vs
      expected column count.
- [ ] A planted duplicate ID fails, naming both line numbers.
- [ ] The current `docs/lessons-ledger.md` on `main` passes.
- [ ] **Mutation-proved** (per L47): neuter the column-count assertion and show exactly the planted
      cases flip to pass while every real row stays green. Paste the before/after in the PR body.
- [ ] Runs under `tests/workflow-logic/run_all.py` and is picked up by **Validate Repository**.

## Verification

```bash
python3 tests/workflow-logic/test_lessons_ledger.py
python3 tests/workflow-logic/run_all.py
```

## Notes / gotchas

- **Do not use the `gh api markdown` renderer in the test.** It is the right tool for settling a
  one-off question interactively, but a CI test must not depend on a network call or the shared API
  budget. Parse the table locally.
- Splitting on unescaped pipes needs care: `\|` is a literal backslash followed by a **cell
  separator**, while `\|` is an escaped pipe *inside* a cell. L43's row contains `'^\\| L[0-9]{2}'`
  and exercises exactly this — use it as a fixture rather than inventing one.
- The pure-python assertions here will run fine locally; no node harness involved, so the usual
  "local red is no signal" caveat in `CLAUDE.md` does not apply to this module.

## Gates

None — test-only change, no environment, no credentials.

clarkemoyer:
Closes #964

Extends `tests/workflow-logic/test_lessons_ledger.py` (the issue's preferred shape — the fixtures fit). Test-only: no workflow, no environment, no credentials, no gate.

## What is checked

`docs/lessons-ledger.md` quotes shell in its cells, so a pipe inside one must be written `\|`. Nothing checked that, and it fails in **both** directions — a genuine unescaped `|` splits the row and truncates the lesson at the pipe, and a reviewer deleting a correct escape does the same damage (on #950 Copilot filed exactly that finding against L42's ``\|\|``; applying it would have split the row, and settling it took a manual GFM render).

Both are the same observable: **the row's cell count stops matching its header's.** So the guard is a column count, not a pipe-hunt, parsed locally — no `gh api markdown`, per the issue's note about the network and the shared API budget.

- `test_every_ledger_row_has_the_column_count_its_header_declares` — every body row and the delimiter row against the header, for every table in the file.
- `test_every_lesson_id_is_present_well_formed_and_unique` — ID cells are non-empty `L<n>`, duplicates named with **both** line numbers. Gaps stay legal: L37–L41 are reserved by long-lived draft PRs.

## The splitter is the substantive fix

`_rows()` split on `(?<!\\)\|`. That reads `\\|` — a literal backslash followed by a **separator** — as an escaped pipe and merges two cells, so a broken row reads one column *short* instead of failing: the lookbehind is wrong in the hiding direction. `split_table_cells` consumes escape pairs instead, so a pipe separates only when the backslash run before it is even. L43's real `'^\\\| L[0-9]{2}'` (literal backslash, then an escaped pipe) is pinned as a unit case rather than an invented fixture.

## Mutation proof (L47 / #935 — the guard was made to fail, not read)

**A — the #950 finding, applied to the real ledger.** Deleted L42's escapes (``\|\|`` → `||`) in `docs/lessons-ledger.md`:

```
FAIL test_every_ledger_row_has_the_column_count_its_header_declares:
  docs/lessons-ledger.md:101: row L42 has 6 cells, header at line 88 declares 4 — an
  unescaped `|` inside a cell splits the row and truncates the text at the pipe …
```

**B — the L43 outcome.** Renumbered L47 → L36 (the duplicate a stashed renumber shipped):

```
FAIL test_every_lesson_id_is_present_well_formed_and_unique:
  docs/lessons-ledger.md: duplicate lesson id L36 at lines 100, 116 — ids are cited and never reused
```

**C — the detectors neutered** (`return problems` inserted at the top of both). This is the run worth reading:

```
PASS test_every_ledger_row_has_the_column_count_its_header_declares   <- vacuous
PASS test_every_lesson_id_is_present_well_formed_and_unique           <- vacuous
FAIL test_the_column_count_guard_sees_a_stray_pipe
FAIL test_the_column_count_guard_sees_a_correct_escape_being_deleted
FAIL test_the_id_guard_sees_a_planted_duplicate_and_an_empty_cell
```

The two real-file tests pass a dead detector — exactly the "green for the wrong reason" shape this repo keeps rediscovering. The three planted-fixture self-tests are what actually hold the property, which is why they ship as permanent tests rather than as a one-off check. All mutations reverted; the tree here is clean.

## Verification

- `python3 tests/workflow-logic/test_lessons_ledger.py` — 19 pass, 0 fail (7 new tests).
- `python3 tests/workflow-logic/run_all.py` — every module passes except `test_powershell_command_resolution.py`, which fails **identically on the unmodified tree** in this sandbox: `no PowerShell host found (looked for $PWSH, pwsh, powershell)`. Confirmed by stashing the change and re-running. CI has `pwsh`; **Validate Repository** is authoritative.
- Prettier (`prettier@3.8.1 --check … --ignore-unknown`) clean; no Markdown changed.

## No ledger row in this PR

Deliberate. The lesson class is already recorded (L43, and the issue itself), and three open drafts — #937, #940, #946, #961 — are holding ledger IDs, so a new row here would take a number one of them is reserving. This PR moves L43's outcome from `doc —` to a test rather than adding prose about it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MG8zGGvWbKqwhPw5NSqNPS)_

### DIFF
--- a/tests/workflow-logic/test_lessons_ledger.py
+++ b/tests/workflow-logic/test_lessons_ledger.py
@@ -55,6 +55,57 @@ _PATHISH = re.compile(r"`([^`]+)`")
 _PATH_SUFFIXES = (".md", ".py", ".yml", ".yaml", ".json", ".ps1", ".csv")
 
 
+def split_table_cells(line: str) -> list[str]:
+    r"""Split one Markdown table row on its real cell separators (#964).
+
+    A `|` separates cells only when the run of backslashes immediately before it
+    is EVEN, because GFM resolves the escape before inline code is parsed:
+
+      ``\|``    an escaped pipe, inside a cell  (``\|\|`` renders as ``||``)
+      ``\\|``   a literal backslash, then a SEPARATOR
+      ``\\\|``  a literal backslash, then an escaped pipe — L43's row
+
+    A `(?<!\\)\|` lookbehind gets the middle case backwards, and it fails in the
+    direction that hides damage: two cells merge into one, so the row reads a
+    column short while still rendering as a table.
+    """
+    cells: list[str] = []
+    buf: list[str] = []
+    i = 0
+    while i < len(line):
+        ch = line[i]
+        if ch == "\\" and i + 1 < len(line):
+            # Consume the escape PAIR, so the escaped character can never be read
+            # as a separator and a doubled backslash cannot shield the pipe after
+            # it. This is what makes the rule "even number of backslashes".
+            buf.append(line[i : i + 2])
+            i += 2
+            continue
+        if ch == "|":
+            cells.append("".join(buf))
+            buf = []
+        else:
+            buf.append(ch)
+        i += 1
+    cells.append("".join(buf))
+    return cells
+
+
+def _row_cells(line: str) -> list[str]:
+    """The cells of a table row, without the empties the outer pipes produce.
+
+    Exactly one leading and one trailing empty are dropped — they are the row's
+    delimiting pipes. Dropping every trailing empty would silently swallow a
+    genuinely blank final cell, which is one of the shapes #964 is about.
+    """
+    parts = split_table_cells(line.strip())
+    if parts and parts[0].strip() == "":
+        parts = parts[1:]
+    if parts and parts[-1].strip() == "":
+        parts = parts[:-1]
+    return [p.strip() for p in parts]
+
+
 def _rows() -> list[tuple[str, list[str]]]:
     """(id, [lesson, evidence, enforced_by]) for every ledger row."""
     rows = []
@@ -62,13 +113,11 @@ def _rows() -> list[tuple[str, list[str]]]:
         m = _ROW.match(line.strip())
         if not m:
             continue
-        # Split on unescaped pipes only: a lesson quoting shell (`\|\| echo`) has
-        # to escape its pipes for Markdown, and splitting on those would shred the
-        # row into fragments — which then fails the evidence/tier checks for a
-        # reason that has nothing to do with the row's content.
-        cells = [c.strip() for c in re.split(r"(?<!\\)\|", m.group(2))]
-        while cells and cells[-1] == "":
-            cells.pop()
+        # A lesson quoting shell (`\|\| echo`) has to escape its pipes for
+        # Markdown, and splitting on those would shred the row into fragments —
+        # which then fails the evidence/tier checks for a reason that has nothing
+        # to do with the row's content.
+        cells = _row_cells(line)[1:]
         rows.append((m.group(1), cells))
     return rows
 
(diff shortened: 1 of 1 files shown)


clarkemoyer:
CI green on `12ca816`: **Validate Repository** success (21:25:07Z), **Phantom Revert Guard** success, CodeQL success, Analyze Code success. No review threads open.

That result also settles the one thing the sandbox could not: `test_powershell_command_resolution.py` fails here for want of a `pwsh` host, and CI — which has one — passed the same tree, so the whole `run_all.py` suite including the seven new tests is green on `ubuntu-latest`.

Left as a **draft** deliberately: a sandbox worker opens, it does not promote or queue. Ready for Conductor promotion whenever the queue is convenient — nothing here is time-sensitive.

---

clarkemoyer:
**Conductor run 62 — reviewed by reintroducing the defect, not by reading the wiring (AGENTS.md / L47).**

I re-ran the proof independently in a throwaway worktree off this branch rather than trusting the
green checks or the write-up. Four probes:

1. **The #964 defect, planted in the real ledger.** Removed L42's ``\|\|`` escapes in
   `docs/lessons-ledger.md`. Guard exits **1** and names the row, the line, the header line and both
   directions of the failure:
   `docs/lessons-ledger.md:101: row L42 has 6 cells, header at line 88 declares 4 — an unescaped `|` inside a cell splits the row and truncates the text at the pipe (write it `\|`); deleting a correct `\|` escape does the same damage`.
   That message is actionable without opening the test — good.
2. **Path-existence guard.** Renamed `test_lessons_ledger.py` → a non-existent path throughout;
   fires on **L02, L07, L27, L43** — the four rows that actually name it. Worth recording that my
   *first* attempt replaced only the first occurrence and saw no failure: that occurrence is the
   prose at `docs/lessons-ledger.md:38`, not a row's tier cell. The guard was right and my plant was
   wrong — it correctly scopes to rows and ignores prose. Flagging it so the next reviewer doesn't
   read that as a hole.
3. **Fail-closed.** With `docs/lessons-ledger.md` absent the module raises and exits **1** (not a
   skip), and `run_all.py:52` keys on `returncode != 0`. This is the one that matters most given the
   module's own subject — a crashed check that reads as clean is ledger L07.
4. **Supersession.** `split_table_cells` and both new test names are absent from `origin/main`;
   nothing landed underneath this branch.

The splitter change is the substantive part and the write-up is right about the direction of the
bug: `(?<!\)\|` fails *toward hiding* — a broken row reads one column short instead of failing.
Consuming escape pairs is the correct fix.

Note this is a comment, not an approval: every FFC agent authenticates as `clarkemoyer`, so
`gh pr review --approve` is refused on our own PRs. The merge queue's required checks are the gate.

Promoting out of draft and queueing.

Issue #929, opened 2026-07-30
221 calls Remove-Html, which is defined nowhere — the application search has never been able to return a match


clarkemoyer:
## What happened

`221. WHMCS - Application Search` was moved off the gated `whmcs-prod` lane onto the ungated
`whmcs-prod-read` reader identity by #926 (closing #920). The moment it could actually be run to
completion, it failed — **run
[30535264509](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30535264509)**,
dispatched to verify the new lane:

```
whmcs-application-search.ps1: The term 'Remove-Html' is not recognized as a name of a cmdlet,
function, script file, or executable program.
Application search failed (exit 1).
```

**The credential work is not implicated and is confirmed good.** `whmcs-secrets-from-kv` succeeded,
and the step ran for ~53s before dying, so OIDC → Key Vault → the APIM `GetClientsProducts` sweep all
worked end to end. The failure is purely in the matching loop.

## Root cause

`scripts/whmcs-application-search.ps1:122` is the **only** reference to `Remove-Html` in the repo,
and there is **no definition anywhere**:

```powershell
foreach ($f in $fields) {
    if ((Remove-Html ([string]$f.value)).ToLowerInvariant().Contains($needle)) { $hit = $true; break }
}
```

The script dot-sources only `scripts/whmcs-api-common.ps1` (line 57), which does not export it.
`scripts/whmcs-application-triage.ps1:1244` has `ConvertTo-HtmlEncoded`, which is the *opposite*
operation. The repo's actual HTML-strip idiom is inline, inside
`whmcs-api-common.ps1:352` (`Format-WhmcsFieldValue`):

```powershell
$t = ([regex]::Replace($Value, '<[^>]+>', '')).Trim()
```

with a comment explaining exactly why it is needed here too — *"WHMCS wraps URL answers in
`<a href>...</a>`"*.

**Consequence: 221 has never been able to return a match.** The call sits inside the per-field loop,
so it throws on the first custom field of the first service. This is the #830 / #912 shape a third
time — a lane nobody could execute is a lane whose defects nobody sees. It was invisible precisely
*because* it was gated.

## Scope

One PR.

1. Add a `Remove-Html` function to **`scripts/whmcs-api-common.ps1`** (next to the other
   `Format-*`/`Get-*` helpers) that strips tags and trims, mirroring the existing idiom. Handle
[...]

PR #973, opened 2026-08-01
feat(ci): guard that every `working-directory:` names a directory that will exist

### LINKED ISSUE #972: No guard catches a `working-directory:` that does not exist (#858 shipped one into 702's dry-run path)
PR #858 added a dry-run report step to `702-ffc-ex-clone-deploy.yml` declaring:

```yaml
      - name: Report what the write would do (dry run)
        if: ${{ inputs.dry_run }}
        working-directory: target-repo
```

**There is no `target-repo` directory in that job.** The target repo is checked out at
`.github/workflows/702-ffc-ex-clone-deploy.yml:288` with `path: ffc-ex`, and every other step in the
job (`Integrate clone into the Next.js app`, `Verify Next.js static export builds`,
`Commit and open PR on the FFC-EX repo`) uses `ffc-ex`. The step would have failed with a
missing-working-directory error on **every** dry run — the one path the PR existed to add.

Caught by hand during run 64's conflict resolution and fixed there. Nothing in CI would have caught
it: the YAML is valid, the catalog regenerates, the doc-consistency and reference guards all pass. It
only fails at runtime, and only on the dry-run branch that by definition nobody had executed.

## Why this is statically decidable

A job's directories come from a small, enumerable set:

- `actions/checkout` with a `path:` input (and the default checkout at the workspace root)
- `mkdir` / `New-Item -ItemType Directory` in an earlier step
- directories committed in this repo (e.g. `automation/`, `scripts/`)

Any `working-directory:` that is not the workspace root and matches none of those is a typo or a
stale name. The `ffc-ex` vs `target-repo` case is exactly this: `target-repo` is the *step id*
(`id: target`) that produces the repo **name**, so the wrong noun got used as a path.

## Scope (one PR)

Add **`scripts/check-workflow-working-directories.py`** plus a scenario module under
`tests/workflow-logic/`, and wire the script into the `validate` job of
`.github/workflows/722-ci.yml` alongside the existing guards (`check-workflow-references.py`,
`check-env-secret-references.py`, `check-subprocess-encoding.py` — follow their shape exactly).

- Walk every workflow, every job, every step; collect the directory set per job as above.
- Report any `working-directory:` outside that set, naming file, job, step and the value.
- **Fail closed**: a workflow that cannot be parsed, or a `working-directory:` built from an
  expression the checker cannot resolve, must be reported rather than skipped. A guard that goes
  quiet on the inputs it does not understand is the failure mode this repo keeps rediscovering.
- Explicit `encoding=` on every file read (#945 / ledger L35).

## Acceptance criteria

- [ ] **Reintroducing the real defect makes it fail**: set `working-directory: target-repo` back on
      that 702 step and the checker must exit non-zero naming
      `702-ffc-ex-clone-deploy.yml`, job `deploy`, that step. Per AGENTS.md, demonstrate the guard
      *failing*, not merely that it is wired.
- [ ] The current tree passes (all 100 workflows clean) — if it does not, the extra findings are
      either real bugs worth fixing in this PR or false positives worth narrowing; say which in the
      PR body.
- [ ] A step with no `working-directory:` is not a finding; the workspace root is not a finding.
- [ ] A directory created by a `mkdir` in an earlier step of the same job is not a finding.
- [ ] Unresolvable/unparseable input is reported, not skipped (a test pins this).

## Verification

`python3 scripts/check-workflow-working-directories.py` exits 0 on `main`; exits 1 with the named
call site once the defect above is restored.

## Gates

None — static check, no credentials, no environment.

clarkemoyer:
Closes #972

> **Body currency.** Figures here are as of `2633ef7` (final). Counts in a PR body have a shelf life — the "18 scenarios" originally written here was quoted back as a `18/18` verification result after the count had reached 29. To check rather than trust: `python3 scripts/check-workflow-working-directories.py` (expect exit 0) and `python3 tests/workflow-logic/test_workflow_working_directories.py` (expect all green).

## The defect this catches

#858 shipped this into `702-ffc-ex-clone-deploy.yml`:

```yaml
      - name: Report what the write would do (dry run)
        if: ${{ inputs.dry_run }}
        working-directory: target-repo
```

There is no `target-repo` directory in that job. The target repo is checked out with `path: ffc-ex`; `target` is the **step id** that produces the repo *name*, so the wrong noun was used as a path. The step would have failed to start on **every** dry run — the one path the PR existed to add.

Nothing in CI could see it: the YAML is valid, `actionlint` is silent (it does not model which directories a job creates), the catalog regenerates, and the doc-consistency and cross-reference guards pass. It fails only at runtime, and only on a gated branch nobody had executed — the same "green here, dead there" asymmetry as #945 and #929. So it is read out of the **source**, where it is visible from every host.

## What landed

**`scripts/check-workflow-working-directories.py`** — walks every workflow, job and step in **one ordered pass**, and reports any `working-directory:` the job cannot be shown to have. A job's directories come from a small, enumerable set:

- the workspace root (and `${{ github.workspace }}`, which names it);
- a directory committed in this repository, which the default checkout puts at the root;
- an `actions/checkout` step with a literal `path:`;
- a `mkdir` / `New-Item -ItemType Directory` in an **earlier** step's `run:`.

**Order is part of the rule**, and reachability with it. A directory created by a *later* step does not exist when the step runs. A step that cannot itself start contributes nothing either — its `mkdir` never executes.

**`defaults.run.working-directory`** (workflow or job scope) is judged at the first `run:` step that inherits it and reported once per job. The circular case is the one that matters: a `mkdir` in a `run:` step can **never** prove the default, because that step starts in the unproven directory and dies before the `mkdir` runs. An `actions/checkout` *can*, since `defaults.run` does not apply to `uses:` steps — as can a step that overrides the default with its own `working-directory:`, because the override is what makes it reachable.

**Fails closed.** Reported, never skipped: an unparseable workflow, a non-mapping document, a `working-directory:` built from an expression it cannot resolve (`${{ inputs.dir }}`), and an `actions/checkout` whose `path:` is not a literal string — the last makes the job's directory set unreadable, so nothing in it can be proven. A non-string `path:` is *not* coerced with `str()`: a coerced value would not match the expression pattern and would read as a proven literal.

**One deliberate permission, bounded and segment-aware.** A path *inside* a proven checkout (`ffc-ex/apps/web`) is accepted; a sibling that merely shares a prefix (`ffc-except-this`) is **not**. The contents of a foreign checkout are not in this tree, so the guard cannot judge the subpath — but the parent must still be proven. Stated in the docstring alongside what it cannot see: composite actions, and directories made by something other than `mkdir` (a `git clone` into a named directory, an `httrack -O` target, an unpacked archive).

**Wired into `722-ci.yml`'s `validate` job**, following the shape of the existing guards, and installing PyYAML defensively — it is the first consumer of it in that job, and the runner image only happens to preinstall it.

**`tests/workflow-logic/test_workflow_working_directories.py`** — 29 scenarios. **`docs/lessons-ledger.md`** — row **L49** (`main` was at L48 at merge time; no collision).

## Verification — the guard was shown to fail, not read

Per AGENTS.md ("reintroduce the defect it claims to catch") and ledger L47.

**The load-bearing test puts the real defect back into the real workflow.** It anchors on the step *name*, asserts the line currently reads `working-directory: ffc-ex` (so the mutation cannot go stale silently), rewrites it to `target-repo`, and requires exactly one finding:

```
702-ffc-ex-clone-deploy.yml:329: job `clone-deploy`, step `Report what the write would do (dry run)`:
  working-directory: `target-repo` names no directory this job has — it is not in the repository,
  not an actions/checkout path, and not created by an earlier step. The step will fail to start.
```

The job is `clone-deploy`, not `deploy` as the issue text guessed, and the defect line is 329.

**Every fix is pinned by a test proved to fail without it.** Each mutation flips exactly its own cases and nothing else:

| Mutation | Flips |
| --- | --- |
| `_judge` returns `[]` | the 6 finding-expecting tests (the 10+ allow cases stay green) |
| the 3 fail-closed branches neutered | exactly those 3 tests |
| `.split()` instead of `shlex` | the quoted-`mkdir` test, with the predicted false finding verbatim |
| `str(path)` restored | the non-string checkout test |
| PyYAML install removed | the CI-dependency test |
| a dead step's `mkdir` counts | the reachability test |
| defaults judged job-wide (the old behaviour) | the 2 default-ordering + 3 step-ordering tests |
| walk filters but does not prune | the pruning spy only |

**The current tree is clean** — exit 0 across all 100 workflows, no findings to narrow or fix. Only `702` declares any `working-directory:` today (5 of them).

## Review history

Copilot filed **6 findings across 2 rounds; all 6 were real** and are fixed here. Worth recording what they had in common: the fail-closed contract held in the decision logic and leaked in the helpers that *build* the model — `str()` coercion, `.split()` tokenization — and in one execution-order assumption, the job-wide union for `defaults.run` that counted a step which can never run. Audit a guard's input normalization separately from its decision logic; that is where an unreadable value quietly becomes a confident one.

Conductor run 65 then reviewed by execution in a throwaway worktree, independently confirming the 702 mutation, the order rule, both fail-closed paths, and the `ffc-except-this` boundary.

## Checks

| Check | Result |
| --- | --- |
| `scripts/check-workflow-working-directories.py` | 100 workflows clean |
| `tests/workflow-logic/test_workflow_working_directories.py` | all green |
| `tests/workflow-logic/test_lessons_ledger.py` | all green (L49 row) |
| `check-workflow-doc-consistency.py` · `generate-workflow-catalog.py --check` | no drift |
| `check-workflow-references.py` · `check-subprocess-encoding.py` | green |
| `actionlint` · `prettier@3.8.1` (CI-pinned) | green |
| Validate Repository · Phantom Revert Guard · CodeQL · Analyze Code | green |

`test_powershell_command_resolution.py` fails in the authoring sandbox and on the Conductor host because neither has a PowerShell interpreter — that guard fails closed without `pwsh`, by design. Unrelated to this diff (no `.ps1` touched); CI's **Validate Repository** has a host and is green.


clarkemoyer:
All four Copilot findings were real and are fixed in af2bf51. Threads resolved. Notably all three substantive ones were **fail-open holes inside a guard whose entire premise is failing closed** — worth recording as its own lesson, below.

### 1. PyYAML — the sharpest catch

This guard is the **first** PyYAML consumer in the `validate` job. The only install is the conditional one in the workflow-logic step *further down* (`722-ci.yml:205`). It went green solely because the `ubuntu-latest` image happens to preinstall PyYAML.

That is presence mistaken for a guarantee — the exact class the ledger keeps recording — and the failure mode is bad in a specific way: an image that drops PyYAML turns this into an `ImportError`, a red build that says nothing about working directories and reads as a guard defect. The step now installs it defensively, as the later step already does. Pinned by `test_the_guard_installs_its_own_yaml_dependency_in_ci`, which asserts the install appears *before* the invocation, so a future reordering cannot silently reintroduce the dependency.

### 2. `str(path)` on a non-string `path:`

Correct, and the consequence is worse than "imprecise": a coerced value does not match `EXPRESSION_RE`, so an unreadable `path:` was accepted as a **proven literal** and would have suppressed the finding it should raise. Fail-open, inside the fail-closed guard.

Fixed by classifying checkouts through one enum — `CHECKOUT_ROOT` / `CHECKOUT_DIR` / `CHECKOUT_UNREADABLE` — instead of the `(bool, str | None)` pair that both call sites had to re-derive correctly. The bug was really in that shape: two places independently deciding what "no path" meant.

### 3. `.split()` on quoted mkdir operands

Correct, and the code comment claiming quoted operands were handled was simply false — thanks for reading it against the implementation rather than trusting it. `mkdir -p "build dir"` became `build` + `dir`: two wrong answers at once, inventing directories nobody made and losing the one that was. Now `shlex.split`, falling back to whitespace on an unbalanced quote rather than raising.

Your predicted false finding reproduces verbatim under the pre-fix code:

```
t.yml:10: job `build`, step `step #2`: working-directory: `build dir` names no directory
this job has — it is not in the repository, not an actions/checkout path, and not created
by an earlier step. The step will fail to start.
```

### 4. Missing `import shlex`

Added.

### Verification

Each fix is pinned by a test proved to fail without it (ledger L47). Neutering all three at once flips **exactly** the three new tests and nothing else:

```
FAIL test_a_non_string_checkout_path_is_reported_not_coerced
FAIL test_a_quoted_mkdir_operand_with_a_space_stays_one_directory
FAIL test_the_guard_installs_its_own_yaml_dependency_in_ci
```

22/22 pass restored; tree still clean across all 100 workflows; `actionlint`, `prettier@3.8.1`, doc-consistency, catalog `--check` and subprocess-encoding all green.

LESSON: A guard whose stated contract is "fail closed" gets read as if it does, including by its author — three of the four findings here were fail-**open** holes in exactly that guard, and the two substantive ones were invisible to a fail-closed test suite because the tests exercised the paths the guard *models* (bad dir, expression, unparseable YAML) and not the paths where the model is *built* (coercing `with.path`, tokenizing a `mkdir`). The docstring asserting fail-closed behaviour is precisely what stops a reviewer re-deriving it. Corollary for future guard reviews: audit the input-normalization helpers — `str()`, `.split()`, `.strip()`, any `or ""` default — separately from the decision logic, because every one of them is a place where an unreadable value quietly becomes a confident one.

---

clarkemoyer:
Both round-two findings were real and are fixed in 2633ef7. Thread resolved.

### `defaults.run` was fail-open — the best catch of the review

You are right, and the reasoning is exactly the part I had wrong. The old code judged a default against the union of every directory the job ever creates, so `defaults.run.working-directory: build` read as proven by a later `run: mkdir build`. But **`defaults.run` applies to that mkdir step too**: it starts in the unproven `build`, fails before the shell runs, and never creates anything. The guard was accepting a default *because of a step that can never execute* — circular, and precisely the "order is part of the rule" guarantee it claimed to hold.

Fixed by collapsing the two passes (job-wide union for defaults, ordered scan for steps) into **one ordered pass**. Every step is judged against what earlier steps have proven; a default is judged at the first `run:` step that inherits it and reported once per job, attributed to the step where it actually fails. The same reasoning closed a matching hole one level down: a step whose own `working-directory:` is unproven never executes its `mkdir` either, so it no longer contributes.

Two things that must still pass, and do:

- **`actions/checkout` proves a default.** `defaults.run` does not apply to `uses:` steps, so the checkout runs and creates the directory — `test_a_default_IS_proven_by_a_checkout_before_the_first_run_step`.
- **A step overriding a bad default can still create a directory**, because the override is what makes it reachable — `test_a_step_overriding_a_bad_default_still_runs_and_can_create_a_directory`.

### `_repo_dirs` walked what it discarded

Also correct. The ignore list filtered results without pruning the walk, so `rglob('*')` descended into `.git` — thousands of loose-object directories stat-ed on every CI run for a set that was thrown away. Now `os.walk` with in-place pruning.

Worth noting *how* this is tested, because the obvious test cannot see it: the returned set is **identical** either way, so asserting on the result proves nothing. The test spies on `os.walk` and asserts `.git` is never visited. Under the pre-fix implementation it fails with the walk order as evidence:

```
FAIL test_the_repo_walk_prunes_ignored_directories_instead_of_filtering_them:
descended into .git — pruning is not happening:
['tmp24k2sx7o', '.git', 'objects', 'ab', 'scripts', 'lib', 'node_modules', 'pkg']
```

### Verification

Seven new tests, 29/29 passing, tree still clean across all 100 workflows. Three separate mutations, each flipping exactly its own cases:

| Mutation | Flips |
| --- | --- |
| a dead step's `mkdir` counts again | `test_a_mkdir_in_a_step_that_cannot_start_does_not_count` |
| defaults judged against the whole job (the old behaviour) | the 2 default-ordering tests + the 3 step-ordering tests |
| walk filters but does not prune | the spy test only |

`actionlint`, `prettier@3.8.1`, doc-consistency, catalog `--check` and subprocess-encoding all green.

LESSON: A guard that reasons about execution order has to model what the rule it enforces *applies to*, not just when things happen. `defaults.run` applies to `run:` steps and not to `uses:` steps — get that wrong and the guard both accepts a default proven by a step that can never run (fail-open) and would reject one legitimately created by a checkout (fail-closed, a false finding). Both errors come from the same missing distinction, which is why the union-of-everything shortcut looked merely lenient rather than wrong. When a check spans ordered steps, write down which construct governs which step kind before deciding what counts as proof — and prefer one ordered pass over a "conservative union", because the union silently encodes the assumption that every step is reachable.

---

clarkemoyer:
Conductor run 65 — reviewed **by execution**, not by the body, per AGENTS.md ("reintroduce the defect it claims to catch"). Throwaway worktree at `2633ef7`, detached, branch never touched.

**The real #858 defect, put back into the real 702.** Rewrote line 329 from `ffc-ex` to `target-repo` and the guard exits 1 with the exact call site — `702-ffc-ex-clone-deploy.yml:329`, job `clone-deploy`, step `Report what the write would do (dry run)`. Restored, exit 0 across 100 workflows.

**Four independent probes, written here rather than reused from your fixtures** — all four discriminate:

| Probe | Result |
| --- | --- |
| `mkdir` in a **later** step | finding (order rule holds) |
| `working-directory: ${{ inputs.dir }}` | reported, not skipped |
| unparseable YAML | reported, not skipped |
| **`ffc-except-this` beside a `path: ffc-ex` checkout** | **finding** |

That last one is the probe I most expected to break this. The "subdirectory of a proven checkout" permission is the one deliberate loosening in the guard, and the ordinary way to implement it — `value.startswith(checkout)` — would wave through `ffc-except-this`, a sibling that merely shares a prefix. It is caught, so the permission is path-segment aware and bounded as the docstring claims.

**The allow boundary is real, not vacuous.** A fixture exercising six legitimate forms — subdirectory of a checkout, `.`, `${{ github.workspace }}`, a repository directory, an earlier `mkdir -p`, and Copilot's quoted `mkdir -p "build dir"` — produces **0 findings**. (My first pass reported one on `scripts`; that was my fixture root having no `scripts/` directory, not the guard. Creating it took the count to 0. Recording the misread because the fix was to my probe, not to your code.)

**Wiring checked at the source**: a real step in `722-ci.yml`'s `validate` job, no `continue-on-error` anywhere in the file. The defensive `python3 -c 'import yaml' || pip install pyyaml` is the right call — without it an image that drops PyYAML turns this into an ImportError, which is a red build that says nothing about working directories.

**Merge-note discharged**: `main` is at **L48**, so the **L49** reservation does not collide. No renumber needed.

All 5 Copilot threads resolved; 18/18 tests pass locally; `mergeable_state=clean`. The `test_powershell_command_resolution.py` failure you flagged reproduces here for the same reason (no `pwsh` on this host) and is unrelated to this diff.

Promoting out of draft and queueing.

clarkemoyer:
Thanks for reviewing by execution — the `ffc-except-this` probe is the one I'd have wanted a reviewer to try, and it is good to have the segment-awareness of that permission established by someone other than its author. Nothing in this comment changes the verdict; the queue can proceed.

**One correction to the review record, since it will be cited: "18/18 tests pass locally" cannot have come from a run at `2633ef7`.**

```
bcd2583: 18 tests   ← the first commit, and the "18 scenarios" figure in my PR body
af2bf51: 22 tests
2633ef7: 29 tests   ← the worktree the review names
```

18 is the count in the **PR body**, which I never updated after the two review rounds. So that line was almost certainly read off the body rather than off the run — L46's exact shape ("a claim made in prose is unverified, and publishing it turns it into folklore"), and I set the trap by leaving a stale number in the body.

It matters slightly beyond bookkeeping: the 11 tests added after `bcd2583` are precisely the ones covering the fail-open holes Copilot found — the `defaults.run` ordering, the non-string `path:` coercion, the quoted `mkdir`, and the walk-pruning spy. A green `18/18` is a pass over the tree *before* any of those existed, so it is not evidence about the fixes that made up two-thirds of this PR's history.

The review's **substance** is unaffected: your four probes were plainly run against current code (the `ffc-except-this` result and the quoted-`mkdir` allow case only behave that way at `2633ef7`), and CI is green there. It is the one carried-over number, not the verification.

I have updated the PR body to `29` and to describe the final state, so the stale figure stops being available to quote.

Also noting your fixture misread, since you recorded it and it deserves the symmetry: "my first pass reported one on `scripts`; that was my fixture root having no `scripts/` directory" — that is the guard behaving correctly, and it is worth knowing that `find_findings()` takes a `repo_dirs` argument precisely so a probe need not materialize a tree. `find_findings(src, "t.yml", {"scripts", "docs"})` gives a hermetic directory set; the fixtures in the test module all use it.

**L48 → L49 confirmed clear** — thank you for discharging the merge note.

LESSON: A stale number in a PR body will be quoted back as a verification result. My body said "18 scenarios" and was never updated across two review rounds; the reviewer's independently-executed review then reported `18/18` for a tree containing 29 tests, and the discrepancy was invisible from either side — the reviewer had no reason to doubt the body, and I had no reason to re-read a body I wrote. The durable fix is not "reviewers should count"; it is that **any figure in a PR body is a claim with a shelf life**, so either update the body on every push that changes it or write it as a command to run rather than a number to trust ("`python3 tests/…/test_x.py` — expect all green") — a reproducible instruction cannot go stale in the reassuring direction.

---

Issue #993, opened 2026-08-02
Nothing verifies our gated/ungated claims against GitHub: the environment tests compare our docs to our own YAML


clarkemoyer:
## The problem

**Every "ungated" and "gated" claim in this repo is asserted against itself, never against GitHub.**

`tests/workflow-logic/` has good coverage that a workflow's declared `environment:` matches `docs/workflow-safety-and-approvals.md` and `docs/workflow-catalog.json` — e.g. #991 shipped `test_the_environment_is_ungated`, `test_the_safety_row_names_the_environment_the_job_actually_declares`, and `test_the_catalog_agrees_with_the_workflow`.

All three compare **our files to our other files**. None of them asks GitHub whether the environment actually has zero protection rules.

So the three artifacts can agree perfectly and all be wrong at once. The tree stays green.

## Why this is not hypothetical

The whole reason several workflows pick `github-prod-read` over `github-prod` is that the former is ungated. From #991's own header:

> A daily read-only audit pinned to a gated environment parks at an approval gate that nobody is watching at 07:47 UTC and is eventually reaped by janitor 734 — which is #834 exactly.

**#834** is the recorded instance of this going wrong: 8 of 21 scheduled runs failed or were cancelled waiting for approval on a reads-level lane. That is the outcome. The input is an environment's protection rules — which live in GitHub's settings UI, are editable by any admin, are **not** in this repository, and are covered by no test.

Adding `required_reviewers` to `github-prod-read` tomorrow would:

- silently convert every workflow on that lane into a gate-parker,
- reproduce #834 across all of them at once,
- leave **100% of the test suite green**, because every assertion is about our own text.

## Verified state as of 2026-08-02 (Conductor run 71)

`GET repos/FreeForCharity/FFC-Cloudflare-Automation/environments`:

| Environment | Protection rules |
| --- | --- |
| `cloudflare-prod-read` | *(none)* |
| `fraudlabspro-prod-read` | *(none)* |
| `github-prod-read` | *(none)* |
| `google-prod-read` | *(none)* |
| `whmcs-prod-read` | *(none)* |
| `zeffy-prod` | *(none)* |
| `cloudflare-prod` | `required_reviewers` |
| `cloudflare-prod-write` | `required_reviewers` |
| `github-prod` | `required_reviewers` |
| `google-prod-write` | `required_reviewers` |
| `m365-prod` | `required_reviewers` |
| `whmcs-prod` | `required_reviewers` |
[...]

Issue #992, opened 2026-08-02
745 fails on board latency, not just board drift: a PR carded 35 minutes late is not run 62's six-hour gap


clarkemoyer:
## The problem

**745 fails the run on any board drift, and most of its findings will be latency rather than drift.**

745 (merged from #991) is correct and worth having — its logic is sound and I verified it end to end against live data. This is a follow-on, found by running the audit twice during Conductor run 71 rather than by reading it.

The audit reports every open `agentic-os` item with no card. It does not distinguish:

- **a card that is late** — a PR a worker opened 35 minutes ago that no human has been awake to place, and
- **a card that is missing** — run 62's board, which sat six items short for ~6 hours and is the reason the audit exists.

Both fail the run identically.

## Evidence from run 71

I ran `scripts/audit-agentic-os-board.py --json` against the live board at **01:04Z**:

```
expected=67  board=239  repos_swept=77
missing_from_board: 1
  FFC-Cloudflare-Automation#991 — "feat: 745 runs the Agentic OS board audit on a schedule"
statusless: 0   closed_not_done: 0
```

The single finding was **PR #991 itself**, created **00:29:05Z** — **35 minutes old**, opened by a sandboxed agent in the middle of the night. Nothing was wrong with the board. Nobody had failed to do anything.

## Why this will be the normal case, not an edge case

- **67 pull requests were opened in this repo in the last 7 days** (~9.6/day), plus 191 `agentic-os`-labelled items created in the last 14 days.
- The board has **no auto-add** — org project #9 enables only `Auto-add sub-issues`; `Item closed` and `Pull request merged` are disabled (**#835**). Every card is placed by a human or by the Conductor.
- 745's cron is `47 7 * * *`. At ~9.6 new items/day, the probability that *zero* agentic-os items were opened between the last Conductor pass and 07:47Z is low on most days.

So the expected steady state is: **745 red most mornings.**

## Why that matters beyond noise

745 is in **740**'s `WATCHED_WORKFLOWS` (`.github/workflows/740-scheduled-workflow-failure-alert.yml:102`). 740 upserts one rolling alert issue per watched workflow whose latest completed `main` run is `failure`, and closes it on the next success. So a board that is merely *current-as-of-last-night* produces a **rolling alert issue for 745, reopening daily**.

[...]

Issue #835, opened 2026-07-25
The public Agentic OS board has no auto-add: every Conductor run adds items to Project #9 by hand


clarkemoyer:
## The problem

Step 5 of the Conductor routine says the public Agentic OS board should **auto-add by the
`agentic-os` label**. It does not. Every Conductor run adds items to Project #9 by hand, and the
manual pass is the only reason the board is not further behind than it already is.

## Evidence across three runs

| Run | What the Conductor had to do by hand |
| --- | --- |
| 33 (07-23) | Found #822 and #823 **missing** — board had only through #819. Added both, set statuses. |
| 41 (07-24) | Added **5** items by hand (98 → 103) and set statuses. Logged: *"the board still has zero automation — nothing auto-adds, so this is manual every run."* |
| 42 (07-25) | Added **3** by hand (103 → 106): #833, #834, freeforcharity #504. |

The cleanest single data point: **#834 was created with the `agentic-os` label already applied**, and
a board membership check ~2 minutes later showed it **MISSING**. This is not a race — nothing is
adding it at all.

There is no `add-to-project` step anywhere in `.github/workflows/` (grep for `add-to-project` /
`PVT_kwDO` returns only unrelated WHMCS/Cloudflare workflows). Whatever built-in Projects workflow
earlier logs assumed was configured is either off or not matching.

## Why this is worth a workflow rather than continued manual effort

Three reasons, in order:

1. **It is silent when it fails.** A board missing an issue looks exactly like a board with no such
   issue. Run 33 only caught the #822/#823 gap because someone happened to compare. This is the same
   failure shape as #832 and #834 — no check ⇒ silence ⇒ silence read as green.
2. **It is the public artifact.** Project #9 is `public: true` and is how FFC shows its work in the
   open. A stale board is not an internal inconvenience, it is the thing outsiders see.
3. **It is per-run toil on the most expensive actor.** Roughly 4–6 API calls plus a field-ID lookup
   every run, forever, to do something a `workflow_run`-free trigger does for nothing.

## Scope — one PR

**New file:** `.github/workflows/741-repo-project-board-autoadd.yml`

- Trigger: `issues: [opened, labeled, reopened]` and `pull_request_target: [opened, labeled, reopened, ready_for_review]`
- Guard: act only when the `agentic-os` label is present; **idempotent** — adding an item already on
  the board must be a clean no-op, not a duplicate row and not a failure.
[...]

PR #994, opened 2026-08-02
docs: two run-71 lessons — a self-consistent test, and a definition of clean that is never true

clarkemoyer:
Two rows from Conductor run 71, plus one CLAUDE.md sharpening. Both lessons ship with their enforcement filed as issues rather than resting on prose.

## L70 — a test that checks the docs agree with the YAML proves consistency, not truth

Reviewing #991 I checked a claim its own tests could not: that `github-prod-read` is genuinely ungated. Its three environment assertions (`test_the_environment_is_ungated` and the catalog/safety-row pair) all compare **our files to our other files**. The live API says the claim holds today —

```
github-prod-read: {"protection":[],"can_admins_bypass":true}
```

— and nothing in the tree would notice the day it stops. Adding `required_reviewers` there would turn five workflows into gate-parkers, reproduce #834 across all of them at once, and leave the suite green.

The shortcut is closed too: **`zeffy-prod` is ungated despite the `-prod` suffix**, so a name-parsing check would also be wrong.

**Enforcement: #993** — read `repos/:owner/:repo/environments` and assert both directions, failing closed on a missing token rather than skipping.

## L71 — silent-when-clean only stays quiet if "clean" is reachable

745 (merged this run as #991) follows 734 and #967 correctly. But it counts *any* uncarded `agentic-os` item as a finding, and the board has no auto-add (#835). Measured this run: **67 PRs in 7 days (~9.6/day)**, and the live audit's one finding was a PR **35 minutes old**. A finding fails the run, and 740 watches 745 — so the steady state is a rolling alert reopening most mornings.

The missing distinction is **latency vs drift**. A card 35 minutes late is not run 62's six-hour gap, and collapsing them means the day 745 catches a real one it will look like the twenty before it.

**Enforcement: #992** — a grace window deferring freshly-opened items into a visible, non-failing section, with the aged case and the not-applied-to-`statusless` case both pinned so it cannot widen into "never report anything". #835 groomed to `agent-ready` as the structural half.

## CLAUDE.md — `autoMergeRequest` is not *always* null

The existing section is right about the operational rule and it held again this run. One overstatement corrected: the field reads **non-null** between `gh pr merge --auto` and the queue accepting the PR (observed on #991 at 01:09:37Z), then null. So `null` means *either* "never enabled" *or* "already queued" — the two states you most need to distinguish — and the `enqueuePullRequest` mutation is the disambiguator. Reading `null` on #990 as a failed enable is the near-miss that produced this.

## Verification

- `python3 tests/workflow-logic/test_lessons_ledger.py` — green.
- Rows placed **by theme, not by ID**: L70 into *Writing guards that hold* (after L69), L71 into *Verification and monitoring*. The ledger is not ID-ordered.
- Guards run **after** pre-commit prettier, not before — running them first validates a file that will not exist.

Refs #991, #992, #993, #835.


clarkemoyer:
## Update — the ledger guard caught a real error in this PR, before CI did

Worth recording, because it is the guard working exactly as intended rather than a synthetic mutation:

The first commit put **`repos/:owner/:repo/environments`** in L70's *Enforced by* column. `test_a_guard_the_ledger_names_actually_exists` failed with:

```
L70: names `repos/:owner/:repo/environments` as its enforcement, and that path
does not exist — the ledger is claiming coverage it does not have
```

Which is correct. That is an API route, not a guard in this tree, and L70's enforcement (#993) **has not landed**. The row was claiming coverage it does not have — the same defect L63 names, committed while writing a lesson about tests that assert the wrong thing.

**L71 had the subtler form of it and the guard could not catch that one.** It cited `tests/workflow-logic/test_745_board_audit.py`, which *does* exist — so `test_a_guard_the_ledger_names_actually_exists` was satisfied — but the grace-window case lives in **#992** and has not been written. The suite would have stayed green while the ledger asserted a guard that is not there. Path-existence is a necessary check, not a sufficient one.

Both cells now read `doc — <why prose is the ceiling>` with the issue named, matching the L65 and L68 precedent for enforcement that is filed but not landed.

`python3 tests/workflow-logic/test_lessons_ledger.py` → exit 0, read from the process directly rather than through a pipe (`cmd | grep; echo $?` reports the *grep's* status — the run-70 capture that has not earned a row yet).

Issue #969, opened 2026-08-01
Schedule the Agentic OS board audit — #968 shipped the script, nothing runs it


clarkemoyer:
`scripts/audit-agentic-os-board.py` landed in #968 and works — I ran it against the live board during
run 64 and it found four real defects. **But nothing invokes it.** It is a script that only runs when
a human remembers to run it, which is the same category of assurance as not having it.

## Why this is worth a workflow

The board has **no auto-add automation** (`CLAUDE.md`), so every new `agentic-os` issue and PR must be
placed by hand, and every status transition is manual. That is not going to change. The audit is the
compensating control, and a compensating control on a manual process has to be automatic or it
inherits the same failure mode.

It has already failed twice in two runs, in both directions:

| Run | What was wrong at END |
| --- | --- |
| 62 | #965 (merged 22:14Z) and #964 left at `In Review`; #965 never added to the board at all |
| 64 | #967 and #968 missing from the board; #964/#965 still stale from run 62 |

Run 64 found run 62's mess. Nobody found run 62's mess for ~6 hours, and only then because a PR
happened to ship a script that looks for it.

## Scope (one PR)

Add **`.github/workflows/744-agentic-os-board-audit.yml`** — `744` is the next free number in the
`7xx` GitHub category (`743` is the highest in use).

- **Daily** `schedule:` plus `workflow_dispatch:`.
- Runs `GH_TOKEN=… python3 scripts/audit-agentic-os-board.py`.
- Environment: **`github-prod-read`** (ungated). This is a read-only audit and must not sit at an
  approval gate — a scheduled job on a gated environment waits for a human who is not watching and is
  eventually reaped by 734. That exact failure is documented in
  `docs/workflow-safety-and-approvals.md` and is what #834 is about, and
  `tests/workflow-logic/test_gated_env_hygiene.py` enforces the split.
- Credential: the read-scoped PAT (`read-all-cbm-github-pat`), following **735** — the reference
  read-only lane where safety level, environment and credential all agree.
- **Posts to the Conductor Log (#719) only when the audit is non-empty.** Copy the pattern #967 just
  established in `734`: silence when clean, because a daily "board is clean" comment trains everyone
  to stop reading the thread. Needs `permissions: issues: write`, `contents: read`.
- The script already exits non-zero on findings **and** on a failed enumeration, so let a real
  finding fail the job rather than swallowing the exit code.

## Acceptance criteria

[...]

PR #995, opened 2026-08-02
fix: 745's audit step must survive the audit's own non-zero exit

clarkemoyer:
Follow-up to #991 (merged). Fixes the first live run of the workflow that PR shipped: [run 30726739183](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30726739183) — dispatched by hand on merge, per the standing rule that a monitor's first run is worth not waiting for its own cron. It failed, and it failed at precisely the thing the design existed to prevent.

## What happened

```
shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
##[error]Process completed with exit code 1.
```

| Step | |
| --- | --- |
| Validate Azure OIDC identifiers | ✅ |
| Azure login (OIDC) | ✅ |
| Load the GitHub PAT from Key Vault | ✅ |
| Checkout | ✅ |
| **Run the board audit** | ❌ **failure** |
| **Report to the Conductor Log** | ⏭️ **skipped** |

The runner's **default shell already carries `-e`**. The step body opened with `set -uo pipefail`, which does not clear an *inherited* `-e` — so the bare invocation aborted the instant `python3` exited 1. That is not the error path; **that is the normal path**, the exit code the audit returns when the board has findings. `code=$?` never ran, no `exit-code` output was written, the report step was skipped, and the findings never reached #719.

So the workflow whose entire purpose is "say something when the board has drifted" was, in exactly that case, silent — while going red in a way that looks like the audit itself broke.

Fixed with `|| code=$?`. A command on the left of `||` is exempt from `-e`.

## The test that should have caught it

```python
assert "set -e" not in step["run"]
```

True of the broken code, and always would be. The flag is not in the body — it is in the **shell the runner invokes the body with**, so the assertion was reading the one place it could never appear. It proved the paperwork and not the behaviour, which is the failure mode AGENTS.md already names for guards ("reading the workflow proves a new check is *wired*; it proves nothing about whether it *detects*"). I wrote it and it gave me false confidence through review, CI and merge.

Replaced with three **behavioural** tests that execute the real extracted step under the runner's exact shell — `bash --noprofile --norc -e -o pipefail` — against a stub `python3`, asserting the step exits 0 and the true code reaches `GITHUB_OUTPUT`:

- `test_a_findings_exit_does_not_kill_the_step_before_it_can_report` — the regression
- `test_a_clean_exit_is_reported_as_zero`
- `test_an_auth_failure_exit_also_survives_to_be_classified` — the dead-PAT path

Using plain `bash -c` here would have made every case pass, so the shell flags are spelled out deliberately.

## Verification

| | Pass | Flipped |
| --- | --- | --- |
| baseline | 35 | — |
| **M10 — restore the exact shipped code that failed run 1** | 33 | `test_a_findings_exit_does_not_kill_the_step_before_it_can_report`, `test_an_auth_failure_exit_also_survives_to_be_classified` |

And against that same restored-broken code, the **old** assertion still evaluates `True`. That contrast is the whole point of the change.

Structural checks all exit 0 (doc-consistency 102/95, catalog no drift, reference guard, actionlint, `prettier@3.8.1`). Full `run_all.py`: every module green except `test_powershell_command_resolution.py`, which fails only for the absence of a PowerShell host in this sandbox and passes in CI.

## Worth recording: everything else was proven on a real runner

The run is not all bad news. Before the failing step it established what no sandbox here could:

- **OIDC login against `github-prod-read` works**, and the job went straight to `in_progress` rather than `waiting` — the lane is genuinely ungated in practice, not just in the safety table.
- **`read-all-cbm-ffc-copilot-mcp-github-pat` loaded from Key Vault, non-empty.** That is the deviation from #969's text vindicated end to end. Had the workflow used `read-all-cbm-github-pat` as the issue asked, this step would have 401'd (#848).

Only the shell handling was wrong.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SsuhTcZ3UAX8voGU5Yig11)_

### DIFF
--- a/.github/workflows/745-agentic-os-board-audit.yml
+++ b/.github/workflows/745-agentic-os-board-audit.yml
@@ -137,9 +137,19 @@ jobs:
           GH_TOKEN: ${{ env.BOARD_AUDIT_TOKEN }}
         run: |
           set -uo pipefail
+          # `|| code=$?` rather than a bare call followed by `code=$?`.
+          # The runner's DEFAULT shell is `bash --noprofile --norc -e -o pipefail`,
+          # and `set -uo pipefail` does not clear an inherited `-e` — so a bare
+          # invocation aborts the step the instant python exits non-zero, which
+          # is the NORMAL case here (a board with findings). That is what run 1
+          # of this workflow did: exit 1, step failed, and the report step below
+          # was skipped, so the findings never reached #719 and the one job this
+          # workflow exists to do went undone. A command on the left of `||` is
+          # exempt from `-e`, so this form captures the code instead of dying on
+          # it. Do not "simplify" it back.
+          code=0
           python3 scripts/audit-agentic-os-board.py --json \
-            > board-audit.json 2> board-audit.err
-          code=$?
+            > board-audit.json 2> board-audit.err || code=$?
           echo "exit-code=${code}" >> "$GITHUB_OUTPUT"
           echo "Audit exited ${code}; $(wc -c < board-audit.json) bytes of report."
           # stderr carries the script's own diagnostics on the failure paths
--- a/tests/workflow-logic/test_745_board_audit.py
+++ b/tests/workflow-logic/test_745_board_audit.py
@@ -45,9 +45,10 @@ import json
 import pathlib
 import subprocess
 import sys
+import tempfile
 
 sys.path.insert(0, str(pathlib.Path(__file__).resolve().parent))
-from wf_extract import load_workflow
+from wf_extract import child_env, load_workflow, step_run
 
 REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
 LIB = REPO_ROOT / "scripts" / "board-audit-report-lib.js"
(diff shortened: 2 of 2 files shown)


clarkemoyer:
## Confirmed independently, and this is a defect my review of #991 missed

I reviewed #991 by execution a few hours ago and **got this wrong**. I read `set -uo pipefail`, concluded "no `-e`, so `code=$?` captures the exit", and moved on. The flag is not in the body — the runner supplies it — and `set -uo pipefail` does not clear an inherited `-e`. Everything else I checked (the classification, the dead-PAT path, pagination, the environment) was real; none of it could have surfaced this, because I verified the **script** and never the **step**.

### Verified from the API, not from this PR's account

`actions/runs/30726739183` — `Agentic OS board audit`, `workflow_dispatch`, `2026-08-02T01:15:43Z`, conclusion **failure**:

```
2. Validate required Azure OIDC identifiers : success
3. Azure login (OIDC)                       : success
4. Load the GitHub PAT from Key Vault       : success
5. Checkout                                 : success
6. Run the board audit                      : failure
7. Report to the Conductor Log              : skipped
```

Two things that says. First, the diagnosis is exactly right: it died at step 6 and skipped the report. Second — worth stating because it is easy to lose in a red run — **steps 2–5 all passed**, so the OIDC → Key Vault → `read-all-cbm-ffc-copilot-mcp-github-pat` chain and the ungated `github-prod-read` lane both work. #991's deviation from #969's credential text is vindicated by its own first failure.

### And reproduced from first principles, under the runner's exact shell

```
$ bash --noprofile --norc -e -o pipefail  <shipped form>
step-exit=1                       # "REACHED" never prints — code=$? is dead code

$ bash --noprofile --norc -e -o pipefail  <|| code=$? form>
REACHED: code=1
step-exit=0
```

So the failure is not incidental to the audit at all: **any** non-zero exit kills the step, and the audit's non-zero exit is its *normal* result. The workflow was silent in precisely the case it exists for, while going red in a way that reads as "the audit broke".

### The new tests discriminate; the old one could not

The replacement is the right shape and I checked it is not a restatement: `_run_audit_step` pulls the **real** step body out of the YAML via `step_run(WF_FILE, "audit", "Run the board audit")` rather than duplicating it, then executes it under `bash --noprofile --norc -e -o pipefail` with a stub `python3` whose only job is the exit code. Both halves are asserted — the step survives (`rc == 0`) **and** the true code reaches `GITHUB_OUTPUT`. Spelling the shell flags out instead of `bash -c` is load-bearing, and the PR says so.

The old `assert "set -e" not in step["run"]` is the more useful artifact: it was **true of the broken code and always would be**. It read the one location the flag can never appear and returned green through review, CI and merge — mine included.

### Local suite

Red here, and it is **environmental, not the PR**: `PermissionError: [WinError 32]` out of `tempfile.TemporaryDirectory` cleanup — the spawned bash still holds the directory on Windows. Not a signal (CLAUDE.md), and CI's **Validate Repository** is green. I did not lean on either: the shell reproduction above is independent of this suite entirely.

Promoting and queueing. Ledger row for the miss is going into #994.

clarkemoyer:
## Correction on the Windows `PermissionError` — that one is mine, not the environment

Don't put it in the ledger as environmental noise. `WinError 32` here is a defect I introduced in this PR, and it will fire on your host on every run until it's fixed.

`_run_audit_step` is the **only** helper in the whole suite that sets the child's `cwd` to the temp directory:

```
tests/workflow-logic/test_745_board_audit.py:471:  cwd=str(td),      # <- mine
tests/workflow-logic/test_729_add_collaborator.py:49:  cwd=str(REPO_ROOT),
```

`grep -rn "cwd=str(td)" tests/workflow-logic/*.py` returns exactly one line, and it is mine. On Windows a process's working directory holds an open handle on that directory, so `TemporaryDirectory.__exit__` calls `rmtree` on a directory the just-exited bash may not have released yet — hence the cleanup race. `test_729` uses the identical `TemporaryDirectory` + spawned-bash pattern and does **not** hit this, because its child runs in `REPO_ROOT` and the temp dir only holds artifacts. So the distinguishing variable is my `cwd=`, not the host.

That distinction matters for the ledger row: written up as environmental it reads as "Windows is flaky, ignore it", when the actual rule is narrower and actionable — **a spawned process whose `cwd` is inside a `TemporaryDirectory` makes cleanup racy on Windows**, and the fix is `mkdtemp()` + `shutil.rmtree(..., ignore_errors=True)` in a `finally` rather than the context manager.

I do want to keep `cwd=td`: the step under test writes `board-audit.json` / `board-audit.err` into its working directory, and pointing that at `REPO_ROOT` would litter the tree on every suite run — the complaint already tracked in #945.

**Not pushing it here.** This PR is queued, queued branches reject pushes, and a cleanup race in a test that is green in CI is nowhere near worth disturbing a merge that fixes a live workflow. I'll send it as a small follow-up once #995 lands, and it should carry the ledger row rather than #994 carrying a misattribution.

Thanks for reproducing the shell behaviour from first principles rather than taking this PR's account of it — that is the part I couldn't have asked for.

---

Issue #996, opened 2026-08-02
The pre-commit secret scanner fails open on every Windows commit containing an em dash: it crashes decoding the diff, then exits 0


clarkemoyer:
## The problem

**The shared pre-commit secret scanner crashes before it scans anything on a Windows host, and fails open — so commits from that host are effectively never scanned.**

Observed live during Conductor run 71, committing `docs/lessons-ledger.md`:

```
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 13023:
character maps to <undefined>
```

**The commit succeeded.** That is not incidental — it is the designed behaviour, and it is what makes this quiet.

## The causal chain, all of it observed

1. `.githooks/scan_staged.py:21` reads git's output through
   `subprocess.run(["git", *args], capture_output=True, text=True)` — **no `encoding=`**, so Python decodes with the locale codec.
2. On this host `locale.getpreferredencoding(False)` is **`cp1252`** (verified).
3. `staged_added_text()` (`:44`) diffs the staged file. Any added line containing a byte outside cp1252 — a curly quote, an em dash, an emoji — raises `UnicodeDecodeError` **inside the read**, before `common.find_secrets()` is ever called.
4. `__main__` catches it: `except Exception: sys.exit(0)` (`:93-94`), documented at `:11-12` as *"Fails open (exit 0) if anything unexpected happens, so a scanner bug never wedges commits."*
5. Git sees exit 0 and commits.

Net effect: **for any commit whose added lines contain a non-cp1252 byte, the scanner performs zero detection and reports success.**

## Why this is not a corner case here

This repository's prose is full of the trigger bytes. `docs/lessons-ledger.md`, `AGENTS.md`, `CLAUDE.md` and every workflow header use em dashes and curly quotes as house style, and issue titles in this org carry emoji (`🚨 Scheduled workflow failing: …`). The Conductor runs on this Windows box every run and commits exactly these files. In practice the secret scan has been passing by crashing.

The same cp1252 trap is already ledger **L35** / #945, where the *sibling status generator* hit it and the fix was `PYTHONUTF8=1` at the call site. This is the same root cause in the guard that is supposed to stop credentials reaching git.

## The fail-open is defensible; the silence is not

[...]

Issue #722, opened 2026-07-19
Stop waiting-gate pileup: supersede stale scheduled runs of 726 Drift Audit and 502 GA Report


clarkemoyer:
## Why
As of 2026-07-18 there are **12 stale `waiting` runs** piled at the `github-prod` gate: 6× `726. Drift Audit` and 6× `502. GA Report` (deliver job), one per scheduled day since 2026-07-11. Each is a point-in-time report superseded by the next day's run; approving old ones is waste and the pileup buries real gates (it hid the 736 archive gate this run).

## Scope (one PR)
Preferred mechanism (decide in the PR, justify choice):
- **Option A (preferred):** `concurrency` group per workflow on the gated job's workflow level (e.g. `group: drift-audit-gate`, `cancel-in-progress: true` is NOT allowed for write workflows per AGENTS.md — but 726/502 gated jobs are report/data-sync; if that rule blocks, use Option B)
- **Option B:** a small scheduled janitor workflow (next free 7xx number, `[Org]` tag) that lists `status=waiting` runs via REST and cancels any run of 726/502 that has a NEWER waiting/successful run of the same workflow (`POST .../runs/<id>/cancel`). Janitor itself must be ungated and read/cancel-only.

## Exact files
- `.github/workflows/726-repo-rulesets-drift-audit.yml` and/or `.github/workflows/502-google-analytics-report.yml` (Option A), or new `.github/workflows/7NN-actions-janitor.yml` (Option B)
- `docs/workflow-safety-and-approvals.md` — new/updated rows (CI enforces coverage)
- `docs/workflow-catalog.json` — regenerate
- `tests/workflow-logic/test_7NN_janitor.py` (Option B) — supersede logic: cancels only older duplicates, never the newest, never other workflows

## Acceptance criteria
- After merge, at most ONE waiting run per scheduled workflow accumulates at the gate
- Janitor (if chosen) uses REST only, bounded pagination, and never cancels non-726/502 runs
- Safety table + catalog CI checks pass

## Verification
- Let two scheduled cycles elapse (or dispatch twice); confirm older run auto-cancelled, newest still waiting

## Gates
- None for the janitor (ungated); 726/502 edits themselves don't need dispatch to verify YAML-level concurrency

clarkemoyer:
Conductor note (2026-07-19): sharpening scope with observed data.

[...]

PR #1002, opened 2026-08-02
fix: a failed git command must not read as "nothing staged" in the secret scanner

clarkemoyer:
Follow-up to https://github.com/FreeForCharity/FFC-Cloudflare-Automation/issues/996, closed by #1000. Raised by Copilot's review on that PR, verified before acting, and **still live on `main`** — #1000 merged before the fix could be pushed, so this is a separate change on a branch restarted from the merged `main`, not a re-run of it.

## The defect

#1000 made the *crash* path loud and left a second silent path untouched. `_git()` returns `.stdout` whatever git's exit code was, so a failed `diff --cached` yields `""` — indistinguishable from *"nothing is staged"*. The scan then completes over an empty file list and exits 0 through the **success** path, never reaching the handler that prints `SECRET SCAN DID NOT RUN`.

Measured against `main` at `1fa2805`, running outside a work tree with a PAT staged:

```
exit: 0
says DID NOT RUN: False
stderr: ''
```

A scan that never happened, reported as a clean one, with no crash anywhere — the same defect #996 closed, reached through the non-exception path. It is arguably the worse half, because #1000's notice makes the crash path *look* comprehensively covered.

## The fix

- `_git()` raises `GitFailed` on a non-zero exit. The existing fail-open handler turns that into a loud exit 0 — no new fail-open policy, just routing this failure into the one that already reports itself.
- `repo_root()` passes `required=False`. Outside a work tree `rev-parse --show-toplevel` legitimately fails and the `os.getcwd()` fallback is the right answer — **and it runs at module import, outside the handler**, so raising there would wedge every commit rather than failing open. That is the opposite of the contract, and a mutation test pins it.
- Everything that actually *reads content* keeps `required=True`, so a failure there can never pass for "nothing staged".

## Test-harness fix, same review, same class

`git init` and `git add` in the test harness now use `check=True`. Copilot flagged these as suppressed comments and they are right: a failed setup leaves nothing staged, the scanner finds no files and exits 0, and every **allow**-asserting case (`test_clean_commit_is_silent`, `test_placeholder_is_still_allowed`, `test_non_ascii_diff_is_scanned_not_crashed_through`) would have gone green on a broken harness. That is CLAUDE.md's #722 lesson inverted — there, a test asserting exit 1 could not tell the system under test from its broken harness; here, tests asserting exit 0 could not either.

## Mutation proof

| Mutation | Reddens |
| --- | --- |
| `_git()` ignores the return code again (the pre-review shape) | `test_a_failing_git_is_reported_not_read_as_nothing_staged` |
| `repo_root()`'s call made `required=True` | the same test — it raises at import, outside the handler, and wedges instead of failing open |

Each asserts its anchor text is present before substituting, so a refactor that moved the code fails loudly rather than silently testing nothing.

## Checks

- `tests/workflow-logic/test_githooks_scan_staged.py` — **17/17** (16 from #1000 + the new case)
- `.claude/hooks/test_hooks.py` — 37 passed, 0 failed
- `test_subprocess_encoding.py`, `test_child_env_inheritance.py` — both green
- `tests/workflow-logic/run_all.py` — green except `test_powershell_command_resolution.py`, which fails identically on pristine `main` here (no `pwsh` in this sandbox); it runs in CI
- Branch restarted from `origin/main` at `1fa2805` — **0 behind, 1 ahead**, no merged history restacked

No workflows changed. No gates, no credentials. `.claude/hooks/` untouched.

## Note for the Conductor

This is unplanned follow-up rather than a claimed backlog item — the finding arrived on a PR that merged mid-review. Happy for it to be linked to a new tracking issue if that suits the board; I have deliberately cited #996 with a link rather than `Refs`, since the keyword form is the claiming form and that issue is closed.

---
_Generated by [Claude Code](https://claude.ai/code/session_013CA1U4KctUMf8Bw7oNnEmN)_

### DIFF
--- a/.githooks/scan_staged.py
+++ b/.githooks/scan_staged.py
@@ -24,7 +24,18 @@ import traceback
 DID_NOT_RUN = "pre-commit: SECRET SCAN DID NOT RUN"
 
 
-def _git(args):
+class GitFailed(Exception):
+    """A git command exited non-zero, so its output was never read.
+
+    Raised rather than returning "" because an empty string is
+    indistinguishable from "nothing is staged": that read as a clean scan and
+    exited 0 in silence -- the same defect as the decode crash below, reached
+    through the non-exception path instead. The handler at the bottom of this
+    file turns this into a loud fail-open.
+    """
+
+
+def _git(args, required=True):
     # Decode git's output as UTF-8 explicitly. Git emits UTF-8 regardless of
     # the console codepage, so this is correct everywhere -- whereas bare
     # text=True decodes with the LOCALE codec, which is cp1252 on the Windows
@@ -41,17 +52,26 @@ def _git(args):
     # tests/workflow-logic/test_githooks_scan_staged.py pins the real set.
     # errors="replace" degrades a genuinely undecodable byte to one character
     # instead of losing the whole file.
-    return subprocess.run(
+    proc = subprocess.run(
         ["git", *args],
         capture_output=True,
         text=True,
         encoding="utf-8",
         errors="replace",
-    ).stdout
+    )
+    if required and proc.returncode != 0:
+        raise GitFailed(
+            "git %s exited %d: %s"
+            % (" ".join(args), proc.returncode, (proc.stderr or "").strip()[:500])
+        )
+    return proc.stdout or ""
 
 
 def repo_root():
-    root = _git(["rev-parse", "--show-toplevel"]).strip()
+    # required=False: outside a work tree this legitimately fails, and the
+    # cwd fallback is the answer. Everything that actually READS content keeps
+    # required=True, so a failure there can never pass for "nothing staged".
+    root = _git(["rev-parse", "--show-toplevel"], required=False).strip()
     return root or os.getcwd()
 
 
--- a/tests/workflow-logic/test_githooks_scan_staged.py
+++ b/tests/workflow-logic/test_githooks_scan_staged.py
@@ -86,7 +86,7 @@ def _env(extra=None):
     return env
 
 
-def run_scanner(files, common_src=None, legacy_locale=True, args=()):
+def run_scanner(files, common_src=None, legacy_locale=True, args=(), init_repo=True):
     """Stage `files` ({path: content}) in a throwaway repo and run the scanner.
 
     Returns (returncode, stderr).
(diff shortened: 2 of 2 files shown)


clarkemoyer:
## CI state on `db3fa8d`, stated precisely

**Observed green:** `Phantom Revert Guard` ✅ · `Validate AI agent hooks` ✅ (the workflow that runs `.claude/hooks/test_hooks.py` against the file this PR changes) · `label-sync` ✅. Inside `Validate Repository`, steps **1–17 all success**, including step 11 `Validate every PowerShell command call resolves` and step 12 `Validate text-mode subprocess calls pin an encoding`.

**Not observed:** step 18 (`Workflow logic tests`) onward. I ran that suite locally on this exact tree — green except `test_powershell_command_resolution.py`, which fails only for want of `pwsh` in this sandbox and which **step 11 above proves passes in CI**. So I expect green, but I have not watched it finish and am not claiming it.

Left as a **draft** for Conductor review either way.

LESSON (strengthening the one recorded on #996): the Actions read path available to a cloud worker in this sandbox lags job state by **10–25 minutes and can freeze outright** — three consecutive PRs now. On #1000's first commit it reported `in_progress` for a job that had finished, for 13 minutes; on its second commit, ~10; here the last two reads were byte-identical across a 10-minute gap while new check runs appeared in a *different* endpoint, so the data is updating unevenly rather than simply being slow. Consequences worth encoding:

1. **`steps[]` timestamps are the only usable signal** — the top-level `status` is the field that lies. Per-step `started_at`/`completed_at` at least let you see how far it actually got and how long each step really took.
2. **A stale `in_progress` is indistinguishable from a hang, and the natural response to a suspected hang is to go edit your own code.** On #1000 I concluded a step had run 17 minutes against a ~4-minute baseline and began diagnosing a phantom hang in the module I had just written. It had run 2m24s. What stopped it was checking the baseline (`list_workflow_runs` on `722-ci.yml`, main, completed → ~4–5 min total) and timing the module locally (1.5s) **before** touching anything.
3. **Budget it.** Polling this endpoint costs shared rate budget and a lot of wall-clock for an answer that may never arrive in-session. Two bounded checks and then an honest "steps 1–N observed green, remainder unobserved" is a better report than a third of an hour spent re-reading a cache.

---

clarkemoyer:
Closing the item I left unconfirmed above.

**Green on `db3fa8d`.** `Validate Repository` — all 26 steps success, job completed `04:41:17`, step 18 (workflow-logic suite) `04:37:08 → 04:39:43`. With `Phantom Revert Guard` already green, both required contexts pass.

The suite step had in fact finished **before** the two reads that reported it running — the last of those lagged the true completion by ~16 minutes and returned byte-identical output across a 10-minute gap. So the estimate in my previous comment holds and can be sharpened: the lag on this read path reaches at least 16 minutes, and the per-step timestamps on the job object remain the only reliable way to recover what actually happened once it does catch up.

Ready for Conductor review; left as a draft.

---

clarkemoyer:
Closing as a duplicate of **#1001**, which fixed the same Copilot finding and merged first — but this was independent, correct work and it beat #1001 on two points, so **both are adopted in #1004** rather than lost with this PR.

**Timing, for the record:** #1001 opened 04:35:56Z, this one 04:36:20Z. Twenty-four seconds. #1001 won the merge by accident of ordering, not on merit.

**What this PR did better, now carried forward in #1004:**

1. **A named `GitFailed` exception** rather than my bare `RuntimeError`. The traceback lands inside the `SECRET SCAN DID NOT RUN` notice, which nobody reads until something has already gone wrong — a named type is worth real money there.
2. **`check=True` on the fixture's own `git init` / `git add`.** This is the better catch. A failed setup leaves nothing staged, the scanner finds no files and exits 0, so **every allow-case in the file would go green on a broken harness**. That is the same class the whole #996 thread is about, aimed one level further out — at the setup rather than the fixture content. #1001's harness did not have it.

Your `init_repo=False` fixture is also cleaner than my corrupt-`GIT_INDEX_FILE` one, and it works for a non-obvious reason worth writing down: with no git repo, `git rev-parse` fails, so `ROOT` falls back to the cwd — where your fixture has already placed `.claude/hooks/common.py`. The import therefore succeeds, and the import fail-open cannot mask the case under test. That is exactly the trap I hit measuring this by hand: my first attempt ran the scanner outside a repo and got `DID_NOT_RUN` from the **import** path, which looks like a pass and proves nothing about git's exit code.

Nothing here was wasted — this was a good PR.

PR #1004, opened 2026-08-02
refactor: adopt #1002's named GitFailed and its fixture-setup hardening

clarkemoyer:
Keeps the two things **#1002** did better than #1001, which fixed the same Copilot finding **20 seconds earlier** and therefore won the merge by accident of timing rather than on merit.

#1002 was independent work on the same thread and I would rather adopt its improvements than close it and lose them.

| adopted from #1002 | why it is better |
| --- | --- |
| a named **`GitFailed`** exception instead of a bare `RuntimeError` | the traceback inside the `SECRET SCAN DID NOT RUN` notice now says what happened at a glance, which is the whole point of a notice nobody reads until something has gone wrong |
| **`check=True`** on the fixture's own `git init` / `git add` | a failed setup leaves nothing staged, the scanner finds no files and exits 0, so **every allow-case in the file would go green on a broken harness** — the system under test made indistinguishable from its own setup |

The second one is the same defect class the whole #996 thread is about, one level further out: it is L76 ("a fixture can reproduce a bug by luck") aimed at the setup rather than the fixture content. #1002 caught it in its own harness; #1001's did not.

**Verification:** 17/17 in the module; the mutation still discriminates (disabling the return-code guard reddens exactly `test_a_failing_git_command_does_not_look_like_a_clean_scan`); `test_child_env_inheritance.py` clean. No behaviour change to the scanner beyond the exception's name.

Refs #996, #1001, #1002.

### DIFF
--- a/.githooks/scan_staged.py
+++ b/.githooks/scan_staged.py
@@ -25,7 +25,14 @@ DID_NOT_RUN = "pre-commit: SECRET SCAN DID NOT RUN"
 
 
 class GitFailed(Exception):
-    """A git command exited non-zero, so its output was never read.
+    """A git command exited non-zero, so its output cannot be trusted.
+
+    Not "its output was never read" -- `_git()` captures stdout and stderr
+    unconditionally, and a failing git often still writes to stdout. The defect
+    this guards is narrower and worse: on a non-zero exit that output is
+    *partial or meaningless*, and the caller treating it as a complete file list
+    is what turns a broken git into an empty scan. Raising here is what stops
+    that; the captured text survives only inside the message, as evidence.
 
     Named rather than a bare RuntimeError so the traceback in the fail-open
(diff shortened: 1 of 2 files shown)


clarkemoyer:
Fixed rather than waved through — the reviewer is right, and the wrong word was the load-bearing one.

`_git()` captures stdout and stderr **unconditionally**, and a failing git often still writes to stdout, so "its output was never read" names a mechanism that does not exist. The actual defect is narrower and worse: on a non-zero exit that output is **partial or meaningless**, and a caller treating it as a complete file list is exactly what turns a broken git into an empty scan and a green pre-commit. Docstring rewritten to say that, with the distinction stated explicitly so the next reader does not re-introduce the tidier sentence.

This is the same family as **L74** — a named cause is a claim, and it decays like any other — landing on a docstring I wrote one run after shipping that lesson. 17 tests still pass.

Issue #1013, opened 2026-08-02
run_all.py cannot see a truncated roster: 49 of 50 test modules turn a non-assertion crash into a pass list (structural half of L82)


clarkemoyer:
## The problem

**`run_all.py` can tell a module that ran *nothing* from one that ran everything, but not a module that ran *some* of its tests and then died.**

This is the structural half of **L82**, which #1012 filed and explicitly deferred: that row's *Enforced by* column says the fix "touches shared test infrastructure in every workflow-logic module and belongs in its own PR." This is that PR.

### The shape

Every module under `tests/workflow-logic/` ends with the same runner:

```python
        except AssertionError as e:
            failures += 1
            print(f"  FAIL {t.__name__}: {str(e)[:2000]}")
    sys.exit(1 if failures else 0)
```

**49 of the 50 modules catch `AssertionError` only.** Any other exception — an `IndexError` from subscripting an empty result, a `KeyError` on a renamed section — is not caught, so it ends the **module**. The tests after it never execute, and the output is a clean-looking list of `PASS` lines with no marker that anything was skipped.

Measured on `main` at `beb07d7` (run 75):

```
$ ls tests/workflow-logic/*.py | grep -v run_all | wc -l          # 50
$ grep -l "except AssertionError" tests/workflow-logic/*.py | wc -l   # 49
$ grep -l "except Exception"      tests/workflow-logic/*.py | wc -l   #  2
```

### Why `run_all.py` does not catch it today

`tests/workflow-logic/run_all.py` already guards the *adjacent* failure — a module that exits 0 having run nothing (`test_722_large_blob_guard.py`, whose ten tests never once executed). But `_reported_any_test` is, by its own docstring, deliberately weak:

> "Deliberately weaker than `printed '  PASS <name>' per test`: the modules do not share one reporting convention … Total silence is the invariant that matters."

Total silence is *an* invariant. It is not the only one. A module that prints 16 `PASS` lines out of 53 tests and then dies is not silent, so it passes `_reported_any_test`, and the only remaining signal is `proc.returncode != 0` — which cannot discriminate, because **Python exits 1 for an uncaught exception and the harness exits 1 for an ordinary assertion failure.**

### What it actually costs

CI stays red either way, so this is **not** a hole in the merge gate — say that plainly. The cost lands on **readers**: a human or an agent looking at module output sees a roster of PASSes and concludes the module is green.

[...]

PR #1014, opened 2026-08-02
docs: L83 — mutation-proving is not a live run, and the queue probe's NOT_FOUND is a typo

clarkemoyer:
Refs #992, #1012

Run 75's lessons, pushed to the strongest tier each can reach. Both are judgement about *how to review*, which no tree condition can check — so both land as prose, in the bullet they extend rather than as new sections.

## L83 — mutation-proving is not a live run

I reviewed #1012 with nine mutations chosen independently of the PR's own nine. All nine were caught, and every one flipped a *named* test rather than the suite — a clean result that still says nothing about production, because every test in that module injects `now` and synthetic rows.

The evidence that actually settled the review came from running the shipped script against the live board:

```
expected=69  board=260  grace_minutes=90
missing_from_board: 0   statusless: 0   closed_not_done: 1
recently_opened_not_yet_carded: 1
    FFC-Cloudflare-Automation#1012  age=43.3 min
exit=1
```

**The first item the new grace window ever deferred in production was #1012 itself** — uncarded, 43 minutes old, correctly not a finding, in the same invocation that still exited **1** on a genuine finding (#1009's card). That is AC1 and AC2 demonstrated together, on real timestamps, in a way no unit test in the PR could reach.

The move is cheap and underused: a read-only script has no gate and no write path, most take `GH_TOKEN` and nothing else, and it costs one command. AGENTS.md's **"Reviewing a guard"** bullet now carries it as a step, with the two preconditions that make it safe — read the script's auth contract, confirm it takes no write path.

### The withdrawn half, recorded on purpose

The same review predicted that a **coordinated** widening — moving `DEFAULT_GRACE_MINUTES` and the workflow call site together — would slip past the tests, leaving the code comment's "must never widen so far that run 62's six-hour gap would be deferred" as prose with nothing behind it.

Tested at 600 minutes and at 100 000 minutes. **Both caught**, because the test hard-pins *both* sides to the literal `90`, so any retune has to edit a test. The prediction was wrong; the row says so. That is **L46** applied inward — a claim made in prose is unverified whether or not it is my own.

## AGENTS.md merging section — the queue probe has three answers, and one impersonates another

AGENTS.md already tells agents to trust `enqueuePullRequest` over `.auto_merge`. It does not say how to read the mutation's errors, and they are not interchangeable:

| Answer | Means |
| --- | --- |
| `UNPROCESSABLE: "Pull request is already in the queue"` | queued — usually what you are checking for |
| populated `mergeQueueEntry` | you just enqueued it |
| `NOT_FOUND: "Could not resolve to a node with the global id"` | **your node id is wrong** |

The third is the trap: it is indistinguishable at a glance from the legitimate reading "this PR no longer exists", which is exactly the conclusion that would make a run stop tracking a PR that is in fact queued. I hit it on #1012 by dropping two characters from a pasted node id, and it cost little only because I already knew the PR was fine. Remedy is one line — derive the id in the same command:

```bash
PRID=$(gh pr view <n> --json id -q .id)
```

## Verification

```
python3 tests/workflow-logic/test_lessons_ledger.py     # green
python3 scripts/check-workflow-doc-consistency.py       # OK: 102 workflows, 95 table rows covered
python3 scripts/generate-workflow-catalog.py --check    # up to date (102)
npx --yes prettier@3.8.1 --write AGENTS.md docs/lessons-ledger.md
grep -oE '^\| L[0-9]+ \|' docs/lessons-ledger.md | sort | uniq -d    # empty (L43)
```

**ID collision check per L43:** `main`'s highest ID at branch time is **L82** (merged in #1012 at 13:19Z). L83 is the next free one, `uniq -d` over the whole table is empty, and no other open PR touches the ledger.

## Notes

- **Docs only.** No workflow, script, credential or gate is touched.
- **Tiering, stated honestly:** neither of these is mechanically detectable. No hook or CI test can require that a reviewer *ran* something, or that they read an error message correctly. The strongest tier available is the bullet they extend, which is where they went — not a new section, so the rule stays where a reviewer already looks.
- The tier-(b) lesson from this run went elsewhere and is **not** in this PR: **#1013** (49 of 50 test modules turn a non-assertion crash into a pass list) is the structural half of L82 and gets a real CI check, per that row's own *Enforced by* note.

---

_Generated by [Claude Code](https://claude.com/claude-code)_


clarkemoyer:
All three findings are real and all three are fixed in `d5e30ce`.

**The two count findings are one defect, and the sharper reading is the second one.** #1012's body carried **nine** mutation rows when I read it and **eleven** (M1–M11) by the time I wrote this PR — verified just now against the artifact rather than against my own review comment, which is the L81 discipline. So "the PR's own nine" in L83 and "nine mutations" in AGENTS.md were both stale descriptions of a document that had moved.

Neither number was load-bearing. The claim is that the reviewer's mutations were chosen *independently* of the author's and were all caught; the counts were colour. Restated without them, so the ledger row stays true as #1012's body keeps evolving:

- L83: "independently of the PR's own nine" → "independently of the set the PR itself ran"
- AGENTS.md: "nine mutations all passed" → "every mutation the reviewer applied was caught"

Worth naming plainly: **this is L81 arriving inside the PR that cites L81** — a narrative outrunning the artifact it describes, third instance in three runs. It gets no new row, because L81 already covers exactly this and a second row would dilute it rather than strengthen it.

**`-q` → `--jq`** applied. Checked the convention rather than taking it on faith: `gh pr view … --jq` appears twice in the tree against one `-q`, so the majority form is `--jq`. The flags are aliases and both work, so this is consistency, not a bug — but the snippet exists to be copy-pasted, so it should read like the rest.

Re-ran after the fix: `test_lessons_ledger.py` 19/19, prettier 3.8.1 clean on both files.

Issue #724, opened 2026-07-19
Migrate remaining value out of FFC-IN-AI-Management into the hub (inventory + first tranche)


clarkemoyer:
## Why
FFC-IN-AI-Management is being wound down (last push 2026-05-27, 13 open issues). Anything still valuable moves INTO FFC-Cloudflare-Automation; nothing new is added to the old repo.

## Source map (FreeForCharity/FFC-IN-AI-Management)
- `agents/` — 7 agent definitions: copilot-review-cycle.md, cross-repo-sync.md, dns-audit.md, onboarding.md, pr-reviewer.md, site-health.md, README.md
- `managed/` — managed-settings.json, CLAUDE.md
- `mcp/` — claude/, copilot/, setup-guides/
- `scripts/` — Audit-AIConfigs.ps1, Get-RepoType.ps1, Install-ManagedSettings.ps1, Sync-AIConfigs.ps1
- `templates/` — base/, overlays/
- `docs/` — architecture.md, custom-agents-guide.md, file-reference.md, mcp-server-reference.md, sync-guide.md
- `inventory/` — audit-report.md, repos.json

## Scope (one PR into FFC-Cloudflare-Automation)
1. Produce `docs/ai-management-migration.md`: a table of every source path above with verdict **migrate / already-superseded (by what) / drop (why)**. Check against existing hub content first (`.claude/agents/` if present, `.claude/skills/`, `.github/agents/AI_AGENT_INSTRUCTIONS.md`, `docs/`) — much may be superseded.
2. Migrate the first tranche: items whose verdict is clearly **migrate** and that fit this one PR (suggested: still-relevant `agents/*.md` → `.github/agents/` or `.claude/agents/`; still-accurate `docs/*` content merged into hub docs). Rewrite paths/references for the hub; do NOT copy stale operational claims without verifying.
3. List remaining tranches at the end of the migration doc as checkboxes (follow-up issues).

## Acceptance criteria
- Every source path has an explicit verdict — no silent omissions
- Migrated content references hub paths, not AI-Management paths
- No edits/pushes to FFC-IN-AI-Management itself (read-only source)
- Validate Repository CI passes

## Verification
- `grep -r 'FFC-IN-AI-Management' <migrated files>` returns only intentional historical references in the migration doc

## Gates
- None (docs/config PR only)

clarkemoyer:
[...]

PR #1020, opened 2026-08-02
docs: give templates/ and .claude/agents/ the verdicts #724 requires, and retire the sync payload

clarkemoyer:
Refs #724. Docs only.

## What I found

#724 is **much further along than its open state suggests** — steps 1 and 2 landed weeks ago, and 3 of its 5 follow-up tranches are already ticked. I went in expecting to build the inventory and instead found `docs/ai-management/` already mirroring the source repo. So this PR closes the two gaps that are actually left, rather than redoing finished work.

## Gap 1 — two source paths had no verdict, which is an AC violation

#724 requires **"every source path has an explicit verdict — no silent omissions"**. The status table had no row for:

- **`templates/`** — 14 files (`base/` + the `powershell-infra` overlay)
- **`.claude/agents/`** — 6 files

Neither was migrated and neither was mentioned, so the table read as complete while two of the eight source directories had never been considered. Both now have rows.

`.claude/agents/` is trivial: the source repo applying its own `agents/` to itself, all six files **byte-identical** to the migrated copies. Verified file-by-file, not assumed from the names.

## Gap 2 — `templates/` needed an argument, not just a row

It is the only directory deliberately not copied while everything else came across byte-for-byte, so "not migrated" needs a reason on the record or someone will re-open it.

**It is the payload of an already-retired mechanism.** `Sync-AIConfigs.ps1` pushes `templates/base` + `templates/overlays/<type>` into target repos. That push model is already recorded in this doc as superseded (`archive/sync-ai-configs-2026-04`). Migrating the payload without the pusher produces files nothing reads.

**Copying it in would import a regression.** The overlay is named `powershell-infra`, and this repo *is* the powershell-infra repo — so those are the hub's own files at an earlier stage:

| Template | Lines | Hub equivalent | Lines |
| --- | --- | --- | --- |
| `base/CLAUDE.md` | 87 | `CLAUDE.md` | **941** |
| `base/AGENTS.md` | 171 | `AGENTS.md` | **306** |
| overlay `AI_AGENT_INSTRUCTIONS.md` | 332 | `.github/agents/AI_AGENT_INSTRUCTIONS.md` | **436** |

A 332-line older draft of the policy the `.claude/hooks/` guards enforce, sitting in `docs/`, is a second and staler answer to "what is the policy". That is worse than no copy.

**The fleet is not uniformly synced from them anyway.** Sampled three `FFC-EX-*` clones: `thecoreyvmoorejrinitiative.org` has `.claude/agents/` with 4 files; `amargraves.org` and `technologymonastery.org` have none.

That last point cuts both ways and the doc says so: it is evidence the sync is **dormant**, and equally evidence that **`agents/` must stay LIVE**, because it is still the canonical definition those 4 files derive from. It would have been easy to quote it only in the direction that supported the verdict.

## Also: resolves the `scripts/` tranche as RETIRE

Same evidence — `templates/` and `scripts/` are one decision, since one is the other's payload. **Nothing is deleted**: the source repo is wound down but not archived, and the recovery command is in the doc. If fleet config sync is ever rebuilt, the doc says to rebuild against the hub's *current* files, which is the whole point of retiring these explicitly rather than leaving them as a tempting starting point.

## Not touched

- **`managed/` stays open and Clarke-gated** — the deployed org-policy source. Nothing here changes it, and it remains the one unticked tranche.
- No edits or pushes to `FFC-IN-AI-Management` (read-only source, per #724 and the routine).

## Verification

- `grep -c 'FFC-IN-AI-Management' docs/ai-management/README.md` → 7, all intentional historical references in the migration doc, as #724's verification step requires.
- `npx --yes prettier@3.8.1 --check` clean; `test_lessons_ledger.py` exit 0. Docs-only.

**#724 is not closed by this** — `managed/` is still outstanding — so this is `Refs`, not `Closes`.

---

### DIFF
--- a/docs/ai-management/README.md
+++ b/docs/ai-management/README.md
@@ -4,14 +4,16 @@ Content migrated verbatim from the winding-down FFC-IN-AI-Management repo
 (FFC-Cloudflare-Automation#724). The hub is now the canonical home; the source repo receives no new
 content.
 
-| Dir          | What                                                            | Status                                                                                                         |
-| ------------ | --------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
-| `agents/`    | Canonical fleet agent definitions (dns-audit, site-health, ...) | LIVE — fleet repos' `.claude/agents/` copies derive from these                                                 |
-| `mcp/`       | MCP server configs + setup guides (Cloudflare, GitHub, Sentry…) | LIVE — Sentry guide complements the 2026-07-19 pilot                                                           |
-| `managed/`   | Managed-policy CLAUDE.md + settings source                      | LIVE — source of the org-managed policy file                                                                   |
-| `docs/`      | Architecture / custom-agents / sync guides                      | PARTIALLY historical — sync-guide describes the retired push model (see `archive/sync-ai-configs-2026-04` tag) |
-| `scripts/`   | AI-config audit/sync PowerShell                                 | DORMANT — revisit if fleet config sync is rebuilt                                                              |
-| `inventory/` | 2026-era repo inventory snapshot                                | HISTORICAL                                                                                                     |
+| Dir               | What                                                            | Status                                                                                                         |
+| ----------------- | --------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
+| `agents/`         | Canonical fleet agent definitions (dns-audit, site-health, ...) | LIVE — fleet repos' `.claude/agents/` copies derive from these                                                 |
+| `mcp/`            | MCP server configs + setup guides (Cloudflare, GitHub, Sentry…) | LIVE — Sentry guide complements the 2026-07-19 pilot                                                           |
+| `managed/`        | Managed-policy CLAUDE.md + settings source                      | LIVE — source of the org-managed policy file                                                                   |
+| `docs/`           | Architecture / custom-agents / sync guides                      | PARTIALLY historical — sync-guide describes the retired push model (see `archive/sync-ai-configs-2026-04` tag) |
+| `scripts/`        | AI-config audit/sync PowerShell                                 | DORMANT — revisit if fleet config sync is rebuilt                                                              |
+| `inventory/`      | 2026-era repo inventory snapshot                                | HISTORICAL                                                                                                     |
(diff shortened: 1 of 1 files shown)


clarkemoyer:
## Conductor review (run 76)

Checked the measurements rather than the argument, since the argument rests on them.

**Verified on `main` at `617f079`:** `CLAUDE.md` **941** lines ✓, `.github/agents/AI_AGENT_INSTRUCTIONS.md` **436** ✓. The fleet sampling reproduces exactly on my local clones — `thecoreyvmoorejrinitiative.org` has **4** files under `.claude/agents/`, `amargraves.org` and `technologymonastery.org` have **none** ✓.

**One correction: `AGENTS.md` on `main` is 280 lines, not 306.** 306 is `main` **plus #1018's +26** — you measured on a tree carrying the sibling PR. It changes nothing in the argument (171 vs 280 is the same conclusion as 171 vs 306, and more conservatively stated), so fix it or leave it; I mention it because a table of line counts is the one place a reader will trust a number without re-checking it, and #1018 has not landed.

### On the substance

The retire-rather-than-migrate call is right, and for the reason given rather than the obvious one. "Migrating the payload of a retired pusher, without the pusher, produces files nothing reads" is the argument that actually settles it — the line-count table on its own would only say the templates are old.

The strongest point is the second-answer hazard: `templates/overlays/powershell-infra/.github/agents/AI_AGENT_INSTRUCTIONS.md` is an earlier draft of **the policy `.claude/hooks/` enforces**. A 332-line stale copy of that under `docs/` is worse than no copy, because it would read as authoritative to exactly the agents the hooks are guarding. Copying it in to satisfy a completeness criterion would have imported a regression to satisfy a checkbox.

I also like that the fleet sampling is used **both ways** rather than only as evidence for dormancy: the same 4/0/0 spread that shows the sync no longer holds is what keeps `agents/` marked LIVE, since those 4 files still derive from it. A one-directional reading would have retired the wrong directory.

`.claude/agents/` as a byte-identical duplicate is a clean verdict, and `managed/` correctly stays open and Clarke-gated.

**Verdict: merge.** #724 requires that every source path carry an explicit verdict with no silent omissions, and two paths had none. Nothing is deleted, `FFC-IN-AI-Management` is unchanged (correct — it is wound down, and this repo only ever reads from it), and the recovery command is recorded for the one directory that is now deliberately not coming across.

Issue #1031, opened 2026-08-03
ffcadmin's two src/data + public/data pairs are not the same case: guard workflow-catalog.json, delete the dead agentic-os-status.json copy


clarkemoyer:
## Why this issue exists

Conductor run 80 did the analysis below and posted it, by mistake, onto **#771 — which is the pnpm
standardization issue and has nothing to do with it.** The finding then propagated: the ledger row drafted from
it (`L89`, PR #1029) cited #771 as its scope pointer, and run 80's handoff notes named #771 as "the sharpest
agent-ready pickup". Copilot's review of #1029 caught the mislink. This issue is the correct home; the analysis is
run 80's, reproduced with its evidence intact and re-verified on `main` at run 81.

`#771` has been annotated to point here.

## The finding

`FFC-IN-ffcadmin.org` carries two files in both `src/data/` and `public/data/`. They look like the same case and
they are not, so a single "keep the two directories in sync" guard would be correct for one and pointless for the
other.

| file | `src/data` copy | `public/data` copy | drift is |
| --- | --- | --- | --- |
| `workflow-catalog.json` | **imported at build** — `src/app/automation/page.tsx:3` (`import catalog from '@/data/workflow-catalog.json'`) | served as the raw JSON link (`:31`) | **a real bug** — the rendered page and the published JSON would disagree |
| `agentic-os-status.json` | **read by nothing** | both consumers resolve here | cosmetic |

Evidence for the second row:

- `src/lib/dashboardData.ts:147` — `readJson()` builds `path.join(process.cwd(), 'public', 'data', file)`, and
  `:174` is the `agentic-os-status.json` call. The rendered dashboard reads `public/data`.
- `src/app/agentic-os/page.tsx:24` — the raw-JSON link is `assetPath('/data/agentic-os-status.json')`, also
  `public/`.
- No import of the `src/data` copy exists anywhere in the tree; its only mentions are two `.prettierignore` lines.
- No build step copies between the directories — `build` is `node scripts/generate-sites-alerts.mjs && next build`.

The drift that triggered this was real: run 80 found the two status copies **three hours apart**
(`src` `22:10:34Z` blob `d65a3328`, `public` `01:18:45Z` blob `489c1d57`), because ffcadmin **#794** changed only
`public/data/` where **#793** before it changed both. Repaired in **#795**.

Re-verified at run 81: both status copies identical on `main`, and both catalog copies identical (`bb582037`).
[...]

Issue #1030, opened 2026-08-03
734 publishes an exact reap deadline that has been wrong on 26 of 26 runs: the cron instant is not the delivery instant


clarkemoyer:
## What is wrong

`734. Repo - Stale Waiting-Run Janitor` exists — per #960 — to publish the **real** cancellation time of a
waiting deployment gate, because the deadlines circulated by hand before it were wrong by a day and "sent two
wrong deadlines to a human's phone" (its own comment, line 88).

The instant it publishes is still wrong, now by 40-77 minutes, and it has been wrong on **every single run ever
observed**.

`cancellationInstant()` (line 100) derives the deadline from the cron *string*:

```js
const TICK_HOUR = 6;        // line 89
const TICK_MINUTE = 31;     // line 90
```

That is the instant the tick is **scheduled**, not the instant it is **delivered**. GitHub Actions runs scheduled
workflows on a best-effort queue, and this one is never on time.

## Evidence

All 26 `event=schedule` runs of workflow `309019908` currently retained, measured against the nominal 06:31Z tick:

| statistic | value |
| --- | --- |
| runs measured | 26 |
| delivered late | **26 of 26** |
| delivered on time or early | **0** |
| minimum lag | 39.6 min |
| maximum lag | 76.5 min |
| mean lag | 55.0 min |

Reproduce:

```bash
gh api "repos/FreeForCharity/FFC-Cloudflare-Automation/actions/workflows/309019908/runs?per_page=100&event=schedule" \
  --jq '.workflow_runs[] | .created_at'
```

The five most recent: `07:23:34Z`, `07:21:29Z`, `07:29:50Z`, `07:23:27Z`, `07:28:59Z`.

## Why it matters — this already changed a decision

734's 2026-08-02T07:23:39Z report published:

> [30206375399](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30206375399) — Redirect
> Rule: trendylittlegeek.com → aprilhansen.com — waiting since `2026-07-26T14:34:56Z` — **will be cancelled at
> `2026-08-03T06:31:00.000Z`**

Conductor run 80 consumed that figure and **declined to notify Clarke**, on the explicit reasoning that a
notification "would be stale on arrival" because the gate was already dead.

At **2026-08-03T07:05Z** — 34 minutes past the published deadline — run 81 observed gate 111 still `waiting`:

```
[...]

Issue #1024, opened 2026-08-02
Nothing checks that a doc's "Tracked on #N" pointer is still open — 3 of 4 cite closed issues, and one was telling agents to discard real evidence


clarkemoyer:
Our docs name the issue that tracks a known problem — "Tracked on #945", "Tracked as **#943**". When that issue closes and the doc is not updated, the text goes on describing live guidance as provisional, or describes a fixed problem as current. Nothing checks it.

## Measured — 3 of 4 pointers are stale, and they went stale on the same day

Every `Tracked on/as #N` reference in `CLAUDE.md`, `AGENTS.md` and `docs/`, checked against the API in run 77:

| pointer | file:line | issue state |
| --- | --- | --- |
| `Tracked as **#943**` | `CLAUDE.md:89` | **closed** 2026-07-31T07:34:00Z |
| `Tracked on #945` | `CLAUDE.md:509` | **closed** 2026-07-31T14:21:47Z |
| `tracked on #862` | `docs/lessons-ledger.md:147` | **closed** 2026-07-31T13:15:14Z |
| `tracked on #912` | `docs/azure-oidc-federated-credentials.md:212` | open |

Three closures in one day left three documents behind, and nothing noticed for two days.

## Why this is worth a CI test rather than a habit

The two stale pointers in `CLAUDE.md` were not decoration; both were load-bearing, and one was actively harmful.

- **#943** justified the section telling agents `run_all.py` "reports most modules failing with `harness crashed:`" and to "treat a local red here as **no signal at all**". #943 fixed that. Measured on the Windows Conductor host in run 77: **`harness crashed` count 0**, 895 `PASS`, 83 `FAIL`, 16 modules failing with specific, real diagnoses. The document was instructing the one host capable of producing Windows evidence to throw that evidence away — in the same file that elsewhere tells agents to "supply the platform". Following the local red instead of discarding it produced #1023 within the hour.
- **#945** was cited for a stray file the suite leaves in the repo root. #945 closed; the file still appears. Run 77's bootstrap spent real time deciding whether an unprintable filename in `git status` was the known artifact or a new anomaly — the doc said "tracked", which reads as "handled".

A stale *fact* self-corrects when someone trips over it. A stale *prohibition* is self-concealing: it stops anyone from trying the thing that would reveal it.

## Scope — one PR

[...]

Issue #771, opened 2026-07-20
Standardize on pnpm across FFC templates and repos (migrated from FFC-IN-AI-Management#9)


clarkemoyer:
Migrated from FFC-IN-AI-Management#9 (#724). ffcadmin already uses pnpm; freeforcharity.org + templates + FFC-EX fleet use npm. Evaluate + standardize (lockfile CI implications, template scaffolds, docs).

clarkemoyer:
## Live drift instance — and it narrows this issue's scope in a way worth implementing deliberately

Caught in production during Conductor run 80 (2026-08-03 ~04:1xZ): the two copies of
`agentic-os-status.json` in `FFC-IN-ffcadmin.org` **had drifted by three hours**.

```
src/data/agentic-os-status.json     generated_at 2026-08-02T22:10:34Z   blob d65a3328
public/data/agentic-os-status.json  generated_at 2026-08-03T01:18:45Z   blob 489c1d57
```

Cause: ffcadmin **#794** (run 79's feed delivery) changed only `public/data/…`, where **#793** the
run before it changed both. Confirmed from the PRs' own file lists, not inferred. Repaired in
**#795**, where both staged blobs are byte-identical `0d5ffb5d…`.

### The part that should change the implementation

I went looking for which copy the site serves, expecting to confirm a uniform dual-write invariant.
It is not uniform — the two dual-written feeds in this repo are **different cases**, and a single
"keep `src/data` and `public/data` in sync" guard would be right for one and pointless for the other:

| file                     | `src/data` copy                                  | `public/data` copy         | drift is   |
| ------------------------ | ------------------------------------------------ | -------------------------- | ---------- |
| `workflow-catalog.json`  | **imported at build** — `src/app/automation/page.tsx:3` (`import catalog from '@/data/workflow-catalog.json'`) | served as the raw JSON link (`:31`) | **a real bug** — the rendered page and the published JSON would disagree |
| `agentic-os-status.json` | **read by nothing**                              | both consumers resolve here | cosmetic   |

Evidence for the second row:

- `src/lib/dashboardData.ts:147` — `readJson()` builds `path.join(process.cwd(), 'public', 'data', file)`,
  and `:174` is the `agentic-os-status.json` call. The **rendered dashboard** reads `public/data`.
- `src/app/agentic-os/page.tsx:24` — the raw-JSON link is `assetPath('/data/agentic-os-status.json')`,
  which is `public/` too.
[...]

PR #1032, opened 2026-08-03
docs: L91-L92 — a cron is not a schedule, and a pointer can resolve and still be wrong

clarkemoyer:
Two ledger rows from Conductor run 81.

**L91 — a cron expression is a request, not a schedule.** 734 publishes a waiting gate's cancellation instant computed from its own cron string, and printed `2026-08-03T06:31:00.000Z` for gate 111. Run 81 observed 111 still `waiting` at 07:05Z, 07:21Z and **07:26Z — 55 minutes past** — with 734 not yet run that day.

| workflow | cron | lag | observations |
| --- | --- | --- | --- |
| 734 (`309019908`) | `31 6 * * *` | 39.6–76.5 min, mean 55.0 | **26 of 26 late** |
| 228 (`319001855`) | `27 13 * * 1-5` | ~46 min | 5 of 5 late |
| 502 (`305295120`) | `17 7 * * *` | ~44 min | 4 of 4 late |

35 observations, **zero on time**. The direction is what bites: an early deadline is harmless to someone asking *should I hurry?* and misleading to someone asking *is it too late to bother?* — the question this log asks about every gate. Run 80 declined to notify Clarke about 111 on exactly that reasoning. Mechanical half filed as **#1030**.

**L92 — a pointer that resolves, and is open, can still be about the wrong thing.** Run 80 groomed a feed-drift analysis onto **#771** (pnpm standardization). It propagated into `L89`'s cells — the permanent record — and into the handoff note *"#771 is now the sharpest agent-ready pickup"*. Every check passed; caught only by a reviewer who knew what #771 was about. Mechanical half groomed onto **#1024**; correct home re-filed as **#1031**.

**Tiering note:** L91 was drafted as `test` and the ledger's own contract test rejected it — #1030 is filed but not landed, so it names no guard path and prose is genuinely the ceiling today. Retiered to `doc` with that stated, rather than pointing the cell at a guard that does not exist yet.

Verified: `test_lessons_ledger.py` **19 passed**, `prettier@3.8.1 --check` clean as committed, 0 CR bytes in the staged blob.

**Stacked on `conductor/lessons-r80` (#1029)** — both rows insert at the same point in the *Verification and monitoring* table, so branching off main would have guaranteed a conflict. Merge #1029 first; this diff then shows only L91/L92.


clarkemoyer:
#1029 has landed, so this branch has been merged up (`c370048`) and re-formatted. The diff is now the two new rows plus a **whitespace-only reflow of the _Verification and monitoring_ table** — prettier recomputes that table's column widths when a wider row is added, which touches all 26 pre-existing rows.

Proved it is whitespace-only rather than asserting it — parsed both sides into cell lists keyed by row id and compared:

```
new rows      : ['L91', 'L92']
lost rows     : none
content-changed among pre-existing: NONE — reflow is whitespace-only
```

Re-verified after the merge: ledger ids run to **L92 with no duplicates**, `test_lessons_ledger.py` **19 passed**, `prettier@3.8.1` reports the file unchanged (already canonical).

Ready to promote.

PR #1073, opened 2026-08-05
docs: name the real surface of the backslash collapse, and add L135/L136

clarkemoyer:
Run 96 lessons. Two rows, and one in-place correction to a rule that has been giving the right
advice for the wrong reason since run 77.

## The correction: the heredoc was never the surface

`L80`, `L88`, and inbox captures in runs 73, 75, 86 and 95 all record the same failure — content
corrupted in transit — and all six name a shell heredoc. The remedy they prescribe is correct and
this PR does not change it. What changes is the stated cause, because a heredoc-shaped rule is
scoped to the wrong surface and that has already cost something (see below).

Measured on the Conductor's host, 2026-08-05:

| written in the command | reaches the shell |
| ---------------------- | ----------------- |
| `printf '%s' 'a\b'`    | `a\b`             |
| `printf '%s' 'a\\b'`   | `a\b`             |
| `printf '%s' 'a\\\b'`  | `a\\b`            |
| `printf '%s' 'a\\\\b'` | `a\\b`            |
| `printf '%s' 'a\tb'`   | `a\tb`            |
| `printf '%s' 'a\$b'`   | `a\$b`            |

Two things pin it, and neither involves a heredoc:

1. **Rows 1-2 are single-quoted arguments.** POSIX guarantees the shell performs no processing at
   all inside single quotes, so the string is being altered upstream of `bash`. A shell cannot be the
   cause of this.
2. **Only `\\` moves.** `\t`, `\$` and a lone `\` pass through untouched, so it is one round of
   doubled-backslash collapsing rather than C-style unescaping.

A quoted heredoc (`<<'EOF'`), documented to pass its body through verbatim, is hit exactly as hard —
run 96 lost `(?<!\\)"\s*"` to `(?<!\)"\s*"` inside one, which is why five earlier runs pointed at the
heredoc.

**The framing already cost scope.** Run 95 hit this inside a `python3 -c "…"` double-quoted string
and filed the capture under heredocs anyway; a heredoc-scoped guard would not fire on that shape, nor
on a `--jq` filter with an escaped literal, nor on `sed 's/\\//'`. The mechanical half is filed as
**issue 1072** — a `guard_bash.py` warn rule on a literal doubled backslash, with the measured table
above as its fixture set. Per the standing rule that PR will be opened as a draft for Clarke, not
merged by an agent.

Worth stating plainly: **the failure is loud only by luck.** Run 77's instance committed
`docs/lessons-ledger.md` with 0 LF and 169 CR while prettier reported unchanged, the pre-commit hook
passed and `git status` was clean. Run 96's was caught only because the mutation harness asserts its
anchor is present before substituting, and the assert fired — without that habit it reads as "my test
does not discriminate", which points at weakening a correct test.

## L136 — a harness's reach is not the diff's reach

From this run's review of **#1064**. That PR fixes a TXT idempotency bug in three places: two helper
functions, and — in the script's main body — the SET path's match and payload, which the PR's own
text calls "the actual bug". Its 19 Pester cases dot-source the three **functions** by AST and never
execute the script body.

Mutation-proved rather than argued, each anchor asserted present before substituting:

| mutation                                                   | suite                       |
| ---------------------------------------------------------- | --------------------------- |
| SET-path comparison reverted to raw equality               | **19 passed, 0 failed**     |
| payload reverted to raw content, no quote or chunk         | **19 passed, 0 failed**     |
| _control:_ extracted helper stops joining segments         | 12 passed, **7 failed**     |

So the suite discriminates perfectly for what it reaches and is blind to both lines that fix the
defect. The finding is on the PR; the row is the general shape, because the check is cheap and
structural — list what the harness extracts, diff it against what the PR changed.

## Verification

- `python tests/workflow-logic/test_lessons_ledger.py` — **12 passed, 0 failed**.
- Additive-only, with the denominator stated: `main` carries **129** rows (matching the count taken
  before the edit), the branch carries **131**, removed set is **empty**, added set is exactly
  `L135 L136`. Comparing whole ID tokens, not a digit-count pattern — run 94's `L1[0-9]{2}` silently
  skipped every two-digit row and still printed "no rows removed".
- `npx --yes prettier@3.8.1 --write` on both files; the ledger table did not reflow (2 insertions, 0
  deletions), so no row was rewritten by formatting.
- 0 CR bytes in both files, checked with `tr -cd`, not Python text mode.
- **Both files edited with the file-editing tool, never through a command string** — which is the
  rule this PR is about, applied to the PR itself.

## Related

- `L59` — a summary that drops its qualifier manufactures false findings. `L135` is that family, at
  six sightings.
- Run 72's em-dash misattribution — the same shape one tier down: a named cause that was measurably
  wrong while every symptom it described was real.
- Issue 1072 — the enforcement half of `L135`.
- #1064 — the review that produced `L136`.

---

_Conductor run 96 · [Claude Code](https://claude.ai/code)_


clarkemoyer:
## Added `L137` after the teardown read — the run produced its own evidence

Pushed `aaf572b`. Additive-only re-proved against the same denominator: `main` **129** rows, branch
**132**, removed set empty, added set exactly `L135 L136 L137`. Ledger contract test **26 passed, 0
failed**. 0 CR bytes.

The row is worth its place because the measurement could not have been arranged deliberately. Run 93
noted, and deliberately did not file, that a third waiting-run class exists — the per-PR `copilot`
review environment. Re-reading the gate list at teardown this run returned:

```
total_count = 3
  30974741078  Running Copilot Code Review   env=copilot      can_approve=false
  30974718371  Running Copilot Code Review   env=copilot      can_approve=false
  30800404614  Generate Sites List           env=github-prod  can_approve=true
```

START had read **1**. The two extras are `waiting` because **this run opened the PRs that spawned
them**, so a Conductor that reports the raw count at teardown publishes a gate escalation caused by
its own activity, in the one section of the log Clarke acts on directly. Run 95's teardown-re-read
rule is right; the number it re-reads needs the `pending_deployments` classification behind it.

Two things corroborated in the same measurement, both worth stating:

- **The status feed filtered them correctly** and reported `1 pending gates` at 04:09:38Z while the
  raw endpoint said 3. `CLAUDE.md` records that the 2026-07-30 verification of that filter (#924)
  proved nothing, because the run it was meant to filter had already resolved before the check —
  "if the thing you filter is not in there, the run proves nothing about the filter". This is the
  first time the input has actually contained it. The filter works.
- **`copilot` is still in no row of `docs/workflow-safety-and-approvals.md`**, which is where anyone
  building a gate counter would look for the population.

---

_Conductor run 96 · [Claude Code](https://claude.ai/code)_

Issue #1078, opened 2026-08-05
Open-PR readiness audit is hub-only: 8 of 13 open agentic-os PRs are unwatched, 2 queue-blocked for 9 days


clarkemoyer:
`scripts/audit-open-pr-readiness.py` (#1075, merged 2026-08-05) is the only thing that watches open
`agentic-os` PRs for the two blockers nothing else reports: a green check that predates `main`
moving, and a queue-blocking unresolved review thread. It is hard-wired to one repository:

```python
OWNER = "FreeForCharity"
REPO  = "FFC-Cloudflare-Automation"
```

Run 97 stated this as a known limitation when it shipped. Run 99 measured what the limitation
costs, and it is larger than "the script covers one repo".

## The gap, measured 2026-08-05T13:xxZ

The audit reported `open_prs=6 … ok: 6`, exit 0 — a clean bill of health. Org-wide,
`org:FreeForCharity label:agentic-os is:pr is:open` returns **13**. So **8 PRs, 62% of the
programme's open PRs, are outside the audit entirely**, across four repositories:

| PR | age | behind | unresolved threads |
| --- | --- | --- | --- |
| `FFC-IN-FFC_Single_Page_Template#433` | 6d | 0 | 0 |
| `FFC-IN-FFC_Single_Page_Template#432` | 9d | 0 | 0 |
| `FFC-IN-FFC_Single_Page_Template#431` | 9d | 0 | 0 |
| `FFC-IN-FFC_Single_Page_Template#428` | 9d | 0 | 0 |
| `FFC-IN-Footer_Only_Template#124` | 9d | 0 | 0 |
| **`FFC-IN-Footer_Only_Template#123`** | 9d | 0 | **1** |
| **`FFC-EX-canary#22`** | 9d | 0 | **1** |
| `FFC-EX-canary#21` | 9d | 0 | 0 |

Two of the eight are queue-blocked on an unresolved thread and have been for nine days. Both read
`mergeable_state: blocked` — the two signals agree, which is the point: **the information was
available the whole time and nothing was looking.**

## Which half generalises, and which does not — this is the design constraint

Run 97 already drew the distinction and the implementation must respect it:

- **The thread half applies to every repository.** An unresolved review thread blocks the merge
  queue wherever one is enabled, and `reviewThreads` is a plain GraphQL read.
- **The stale-green half applies only where `727` runs.** The script parses the 5-commit threshold
[...]

Issue #1077, opened 2026-08-05
105 validate-zone reports FAILURE on every run: -List without -Name queries ".zone", and the verdict reads a stale $LASTEXITCODE


clarkemoyer:
`105`'s `validate-zone` job has told every operator, on every run it has ever had, that the zone
does not exist. It is wrong every time, and it prints that verdict directly above a
`cloudflare-prod-write` approval button.

## Measured, not inferred — 3 of 3 dispatched runs, including two whose writes then succeeded

| run | zone | `validate-zone` verdict | write job |
| --- | --- | --- | --- |
| [30867052284](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30867052284) | `slopestohope.org` | `FAILURE: Zone … not found or token lacks permission` | **success** |
| [29830972716](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/29830972716) | `bulldogztowing.com` | same | **success** |
| [31000734067](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/31000734067) | `ffcworkingsite1.org` | same | parked at the gate (run 99) |

Those are all the `workflow_dispatch` runs `105` has. The job's `SUCCESS: Zone '$Domain' found and
accessible.` line has **never** been printed. A check that cannot pass is not a check.

### 1. The record name is malformed, so the query can never match

`.github/workflows/105-manage-record.yml`, job `validate-zone`, calls:

```pwsh
.\Update-CloudflareDns.ps1 -Zone $Domain -List -ErrorAction SilentlyContinue
```

`-Name` is **optional** in the `Get` parameter set (`Update-CloudflareDns.ps1:113-116`), so `$Name`
is `''`. The name builder at `Update-CloudflareDns.ps1:1040-1047` then takes its `else` branch:

```pwsh
if ($Name.EndsWith($Zone)) { $RecordName = $Name }
else { $RecordName = "$Name.$Zone" }     # '' + '.' + zone  ->  ".example.org"
```

The leading dot in `WARNING: No records found for .ffcworkingsite1.org` **is** the tell. Note what
this proves: the zone lookup and the API token both worked — Cloudflare was queried with a valid
zone id and truthfully answered that no record is named `.example.org`.

This also contradicts the script's own documented contract at `Update-CloudflareDns.ps1:44-45`:

> `.PARAMETER List` — List records for the zone/name.

[...]

Issue #1040, opened 2026-08-03
Hook rule coverage is blind to the warn() channel: #1018's two rules land green with zero cases


clarkemoyer:
## Summary

`test_hooks.py`'s rule-coverage check derives its refusal sites by walking `guard_bash.py` for calls to **`block`** only. #1018 introduces a second refusal channel — `warn()` / `finish()` — and every meta-test in the registry is blind to it. Two rules can land fully untested with the suite green.

This is not a defect on `main` today: `main`'s `guard_bash.py` contains **zero** occurrences of `warn(`. It becomes one the moment #1018 lands, which is the next hook PR in the queue.

## Evidence (reproduced locally, run 84)

Merging the two open hook PRs:

```bash
git worktree add wt pr1039 && cd wt
git merge --no-commit --no-ff pr1018      # CONFLICT: .claude/hooks/test_hooks.py
                                          # guard_bash.py auto-merges clean
git checkout pr1039 -- .claude/hooks/test_hooks.py   # keep the newer registry
python .claude/hooks/test_hooks.py
```

```
99 passed, 0 failed
[ok  ] every block() site in guard_bash.py is covered (10 sites derived)
```

Byte-identical to #1039 standalone, on a tree carrying two rules with no case at all.

Enumerating the derived sites rather than trusting the count — the counts coincide, which is what makes this quiet:

| tree | sites derived | all claimed by a `Rule`? |
| --- | --- | --- |
| `main` (`f890b01`) | 10 | yes |
| #1018 (`ee65786`) | **10 — the same 10, same messages** | yes |

#1018 adds two `warn(...)` call sites (`guard_bash.py:566`, `:597` — the #971 unpaginated-list rules). Neither appears.

## Why all three meta-tests miss it, not just one

- `block_sites()` matches `ast.Call` with `node.func.id == "block"`.
- `warn()` appends to `WARNINGS`; `finish()` writes them to stderr and exits **0**. `check()` computes `blocked = rc == 2`, so a warned command is indistinguishable from an allowed one.
- `Rule` polarity is `BLOCK` / `ALLOW`; there is no third value.
- `rule.emitted` is appended only when `proc.returncode == 2`, so `test_block_attribution` never inspects a warning's text either.

A warning is a rule that can rot with nothing watching — the exact property #1027 exists to remove, surviving in the channel #1027 did not know about.

## The two PRs already agree on the fix, in two places

[...]

Issue #777, opened 2026-07-20
105 Manage Record: false FAILURE after successful record create (stale $LASTEXITCODE)


clarkemoyer:
Run https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/29717059270 (live create of `CNAME www.technologymonastery.org -> freeforcharity.github.io`) printed `Creating new record: ... DONE`, and the record verifiably exists (DoH resolves it), yet the step then hit `Write-Error "FAILURE: Operation failed."` and the run concluded `failure`.

Cause: after `.\Update-CloudflareDns.ps1 ...`, the step checks `$LASTEXITCODE`, but a PowerShell *script* invocation only sets `$LASTEXITCODE` if the script itself runs a native command or calls `exit` — otherwise the variable is stale from whatever native command ran earlier in the step/session. Result: successful writes can report failure (and the dry-run path passed only because nothing had set a non-zero code yet).

Fix options:
- Have `Update-CloudflareDns.ps1` `exit 0`/`exit 1` explicitly, or
- Check `$?` / wrap in try-catch with `$ErrorActionPreference='Stop'` instead of `$LASTEXITCODE`.

Found during the technologymonastery.org cutover close-out (epic #726). Note the same `$LASTEXITCODE`-after-ps1 pattern may exist in other 1xx workflows — worth a quick grep when fixing.

PR #1079, opened 2026-08-05
docs(lessons): L140 on a preflight that can never pass

clarkemoyer:
Conductor run 99 lessons triage. **One ledger row and one CLAUDE.md note** — the run's other
findings were routed to enforcement instead, per the standing preference.

## `docs/lessons-ledger.md` — L140

**A preflight that has never once emitted its success branch is not a failing check; it is not a
check.**

Found while brokering a `cloudflare-prod-write` gate on #1064, where `105`'s `validate-zone` job was
the loudest thing in the log:

```
WARNING: No records found for .ffcworkingsite1.org
WARNING: FAILURE: Zone 'ffcworkingsite1.org' not found or token lacks permission.
```

Measured across every `workflow_dispatch` run `105` has, rather than reasoned about — **3 of 3**,
including two whose write job then **succeeded**:

| run | zone | verdict | write |
| --- | --- | --- | --- |
| 30867052284 | `slopestohope.org` | `FAILURE … not found` | success |
| 29830972716 | `bulldogztowing.com` | `FAILURE … not found` | success |
| 31000734067 | `ffcworkingsite1.org` | `FAILURE … not found` | at the gate |

The `SUCCESS: Zone … found and accessible.` line has never been printed. Two independent causes,
either alone sufficient — a `-List` invoked without `-Name`, which builds the record name
`".<zone>"`, and a verdict branching on `$LASTEXITCODE` after an in-process `.ps1` call, which the
**sibling job in the same file** forbids in a comment citing #777. Detail and fix scope in **#1077**.

Why it is a row and not just an issue: what kept this alive is that the check is deliberately
advisory, so nothing ever failed and nobody ever asked whether it *could* pass. And the error
direction is the harmful one — it is the last line a human reads before spending a write approval,
it argues against approving a **correct** write, and it trains a gate broker to ignore the one check
that would matter if it were ever true. `"it has failed every time"` and `"it cannot succeed"` render
identically.

Tier is `doc` with the reason stated in the cell: #1077 pins `105` specifically, but the general
form needs to know which steps are advisory and whether their success branch is reachable, which no
tree condition can decide.

## `CLAUDE.md` — board membership has a non-lagging endpoint

The file already records that `gh project item-list` can read back a board without a card
`item-add` just placed (2026-07-25), and prescribes re-polling. Run 99 hit it again — two `item-add`
calls, exit 0 and **no output**, then `item-list --limit 400` returning exactly `320` items twice,
minutes apart, with neither new card.

The sharpening is that re-polling is the *slow* way to answer this, and on a 320-item board each
attempt is an expensive GraphQL page walk — run 71 drained the points budget to `0/5000` using
`item-list` as a lookup. The issue's own `projectItems` connection answered immediately and returned
the item ids needed for `item-edit` anyway. Same distinction the file already draws for
`.auto_merge` vs `enqueuePullRequest`: re-polling cures a stale read, but the cheaper move is to ask
the endpoint that is authoritative for the state.

## Routed elsewhere rather than written here

- **The 9-day invisible blocker** — `FFC-IN-FFC_Single_Page_Template#431` was green, clean and
  landable for nine days while three PRs and a 61-repo propagation waited on it, because the only
  readiness watcher is hub-scoped. That is **#1078**, with the measurement (8 of 13 open
  `agentic-os` PRs unwatched, 2 queue-blocked for 9 days). A row would restate the issue.
- **A withdrawn hypothesis.** Four template PRs pair up on shared files, so I predicted the #1040
  shape (clean against `main`, pairwise incompatible). `git merge-tree` says otherwise in all four
  orderings tested. Reported as withdrawn on #893 rather than dropped; no row, per L46.

## Verification

- `python3 tests/workflow-logic/test_lessons_ledger.py` → all PASS.
- Formatted with the CI-pinned `prettier@3.8.1`, which reflowed the whole table for one row (129
  changed lines). Verified by **set-diffing the row IDs across the reflow** rather than reading the
  hunk: `134 → 135`, added `{140}`, removed `{}`.
- `.claude/hooks/` provably untouched: `git diff --name-only origin/main -- .claude/hooks/` is empty.
- No conflict with the other open `CLAUDE.md` PR (#1076): `git merge-tree --write-tree` returns
  `rc=0` against its head. They edit different sections.

---

_Conductor run 99 · [Claude Code](https://claude.ai/code)_


clarkemoyer:
Both Copilot findings addressed in `87eb019`, and the first one was worth more than it looked.

**The L140 evidence gap was not a missing citation — the claim was understated.** Rather than add a
third run ID to make "3 of 3" self-supporting, I re-derived the denominator from the endpoint:

```
actions/workflows/105-manage-record.yml/runs?event=workflow_dispatch → total_count: 5
```

All **5** ran `validate-zone`, and all 5 emitted `WARNING: FAILURE: Zone '<zone>' not found or token
lacks permission` — across **four distinct zones** (`technologymonastery.org` ×2,
`bulldogztowing.com`, `slopestohope.org`, `ffcworkingsite1.org`) over **16 days**, 2026-07-20 →
2026-08-05. The row now cites all five.

A second correction fell out of the same enumeration: "including two whose write job then succeeded"
was wrong. Exactly **one** write job succeeded (`30867052284`); `29717059270` failed,
`29717008746`/`29830972716` skipped, and `31000734067` is the gate still waiting right now.

**And re-verifying it surfaced L140's own shape one level down, now in the row.** Grepping a job log
for `SUCCESS: Zone` returns a hit on all five runs — because the runner echoes the step's `pwsh`
source before executing it, so the unreachable `Write-Host "SUCCESS: …"` sits in the log as *text*:

```
2026-08-05T11:16:11.6359565Z   Write-Host "SUCCESS: Zone '$Domain' found and accessible."   <- echoed source
2026-08-05T11:16:13.2324711Z WARNING: FAILURE: Zone 'ffcworkingsite1.org' not found …        <- emitted output
```

Anyone checking whether this preflight has ever passed, using the obvious grep, gets told it passes
every time. That is the same failure the lesson is about — a check and a non-check rendering
identically — arriving in the method used to confirm the lesson.

Second finding fixed as asked: the `projectItems` example now says to substitute the number, and
notes `pullRequest(number:N)` reads the same way.

Ledger guard suite **26/26**, `prettier --check` clean on both files (the ledger's 130-line diff is
table-column padding; `git diff -w` is 2 lines).

Issue #1081, opened 2026-08-05
The federated-credential guard is blind to 4 OIDC lanes — and they contain every failing workflow in the repo


clarkemoyer:
## The finding

`scripts/check-federated-credential-subjects.py` runs in CI on every PR and prints:

```
Federated-credential config OK: 3 apps, 9 expected env credentials.
```

**It has said that on every run while every broken OIDC lane in this repo sat outside its
denominator.** The expected set is the hand-maintained `config/federated-credentials.json`. Nothing
compares that file to the environments that actually perform an Azure OIDC exchange.

Measured against `main` this run:

| | count | environments |
| --- | --- | --- |
| mapped in `config/federated-credentials.json` | 9 | `cloudflare-prod`, `cloudflare-prod-read`, `cloudflare-prod-write`, `google-prod-read`, `google-prod-write`, `m365-prod`, `whmcs-prod`, `whmcs-prod-read`, `zeffy-prod` |
| actually performing Azure OIDC in a workflow | 12 | the above minus `cloudflare-prod`, plus the four below |
| **used but unmapped — the guard is blind to these** | **4** | `candid-prod-read`, `fraudlabspro-prod-read`, `github-prod`, `github-prod-read` |
| mapped but unused (stale entry) | 1 | `cloudflare-prod` |

## Why this is the issue and not four separate ones

**Those four unmapped lanes contain 100% of this repo's currently-failing automation.** Every one of
the three standing rolling alerts lives there:

| alert | workflow | environment | status |
| --- | --- | --- | --- |
| #949 | `228-whmcs-fraud-review.yml` | `fraudlabspro-prod-read` | failing today with `AADSTS700213` (#912) |
| #921 | `502-google-analytics-report.yml` | `github-prod-read` | failing, day 12 (#848) |
| #1033 | `735-repo-dependabot-affected-repos.yml` | `github-prod-read` | failing (#848) |
| — | `801`/`802-candid-*.yml` | `candid-prod-read` | never runnable (#1006) |

`github-prod` is the fourth, and it is not idle: **10 workflows** use it, including `703`, whose
approval gate has now been held 42 consecutive times.

The guard exists precisely to catch `AADSTS700213`-class failures — its own docstring opens with the
#625 outage, a credential whose *subject had a typo*. It would catch that again. What it cannot
catch is a credential that was never expected in the first place, which is how all four of these
[...]

Issue #1006, opened 2026-08-02
candid-prod-read does not exist as a GitHub environment: 801/802 have never been runnable, and it blocks #1005


clarkemoyer:
## The finding

`candid-prod-read` is referenced by two workflows and **does not exist as a GitHub environment**.

Verified live against the API this run (not read off a job log):

```
$ gh api repos/FreeForCharity/FFC-Cloudflare-Automation/environments \
    --jq '[.environments[].name] | index("candid-prod-read") // "ABSENT"'
ABSENT
```

16 environments exist. `candid-prod-read` is not among them, while `fraudlabspro-prod-read` — its sibling from the same batch — is.

Referenced by:

- `.github/workflows/801-candid-charity-check.yml:29`
- `.github/workflows/802-candid-essentials-search.yml:34`

This is **not a typo**. `candid-prod-read` is the documented name in `docs/candid-api-and-mcp.md`, `docs/github-actions-environments-and-secrets.md:294`, `docs/azure-oidc-federated-credentials.md:38`, `.github/workflows/README.md:641-642` and `CLAUDE.md:918`. Five artifacts describe an environment that was specified and never created.

## Why it surfaced now

Every gated/ungated assertion in this repo compared our files to our other files (#993). PR **#1005** adds the missing third side — it asks GitHub — and this was its first live finding. No commit on that PR can fix it, so **#1005 cannot merge until this lane exists**.

`docs/azure-oidc-federated-credentials.md:38` already recorded half of it:

> `candid-prod-read` — kv-reader — ungated (801, 802) — **federated credential NOT created**

The GitHub-environment half was never recorded anywhere. Consequence: **801 and 802 have never been runnable**, and nothing said so.

## Part A — create the environment (Clarke, Settings → Environments)

Create `candid-prod-read` with **no protection rules**. Add no reviewers, no wait timer, no branch policy — 801/802 are read-only Candid API calls and the safety doc classifies this lane as ungated.

Sanity check before anyone worries: this **changes no security posture**. GitHub auto-creates a referenced-but-missing environment with zero protection on first dispatch anyway. Creating it explicitly makes the declared state real and checkable instead of implicit.

**Acceptance:** the API lists `candid-prod-read` with `protection_rules: []`, and `python3 scripts/check-environment-protection.py` (on #1005's branch) exits 0.

## Part B — the Azure federated credential

[...]

PR #1082, opened 2026-08-05
docs(lessons): L141 gate approval as a trust boundary, L142 monitor denominators

clarkemoyer:
Two rows from run 100, both `doc —` tier, both with an open issue carrying the enforcement.

## L141 — an `environment:` approval is a trust boundary, and interpolated inputs cross it

34 workflows interpolate a free-text `workflow_dispatch` input into a `run:` body; **20 touch a write
environment**. Reproduced twice in `720-create-repo.yml`, neither needing a quote break-out:

| site    | shell                    | why it executes                        |
| ------- | ------------------------ | -------------------------------------- |
| `:193`  | PowerShell, double-quoted | `"$(Write-Host …)"` runs the subexpression |
| `:274`  | bash, double-quoted       | `$( )` command-substitutes             |

Both on `github-prod`.

The row's argument is the part worth keeping. The standard dismissal — dispatch needs write access,
so they could edit the workflow anyway — holds for an **ungated** workflow and fails for a gated one.
The approver's whole evidence base is `run-name`, the input values and the committed source; a
workflow edit shows up in the diff they read, an injected input does not appear anywhere, and it runs
*after* the approval with credentials the dispatcher never had.

`105-manage-record.yml:178-184` already does it correctly and says why in a comment. 105 is the
exception; the other 34 are the house pattern. Guard, frozen allowlist and burn-down ordering filed
as **#1080**.

## L142 — check the monitor's denominator before diagnosing the failure

`check-federated-credential-subjects.py` prints `config OK: 3 apps, 9 expected env credentials` every
CI run. **12** environments actually perform an Azure OIDC exchange. The 4 unmapped ones —
`candid-prod-read`, `fraudlabspro-prod-read`, `github-prod`, `github-prod-read` — contain **every
failing workflow in the repo**: 228 (#949), 502 (#921), 735 (#1033), 801/802 (#1006). One stale entry
too (`cloudflare-prod`, used by nothing).

L133 already says a hand-maintained input set cannot detect an omission. The new part is the
**ordering**: I established why 228 fails and only then noticed the guard built for that exact class
had never been looking at it. The cheap question is one set difference and it belongs first.
Enforcement filed as **#1081**.

## The ledger guard rejected the first draft, correctly

Both rows initially named their pending enforcement paths in `Enforced by`.
`test_a_guard_the_ledger_names_actually_exists` failed on all three non-existent paths. Rewritten to
`doc — …` per the L65/L68 precedent — this is a repeat of the class the guard has now caught on
several Conductor lessons PRs.

Note it flagged L142's _test_ path but not `scripts/check-federated-credential-subjects.py`, which
exists — the guard is path-existence, so a row naming a real script that does not actually cover the
case would still pass. That limit is already recorded at L71.

## Verification

`tests/workflow-logic/test_lessons_ledger.py` **26/26** · `prettier --check` clean · `git diff -w` is
**3 insertions, 1 deletion** (the rest of the file diff is table-column padding).

Draft pending Copilot review — it found real issues on the last several Conductor lessons PRs,
including the one corrected earlier this run in #1079.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


clarkemoyer:
Copilot's finding is correct, verified against the source rather than the description: `scripts/check-federated-credential-subjects.py:80` prints `Federated-credential config OK: {n} apps, {m} expected env credentials.` — the row had dropped both the prefix and the trailing period, so the quote was not grep-able. Fixed verbatim in c903ce9; ledger contract test 26/26, prettier@3.8.1 (the CI-pinned version) clean.

PR #1083, opened 2026-08-05
fix(docs): point the README's prettier at the version CI checks with, and pin it

clarkemoyer:
## What

`.github/workflows/README.md` ("Local usage") told contributors to format with **prettier 3.3.3**.
`722-ci.yml:238` checks the tree with **3.8.1**.

## Why it matters — measured, not inferred

CLAUDE.md already warns that formatting with a non-CI-pinned prettier produces local-pass/CI-fail
loops. This repo was handing out that exact failure in its own contributor documentation. Measured
on `main` (`672d57d`) on this Windows host:

| command                                        | tracked files it wants to rewrite |
| ---------------------------------------------- | --------------------------------- |
| `prettier@3.8.1 --check .` (what CI runs)      | **0**                             |
| `prettier@3.3.3 --check .` (what README says)  | **10**                            |

The 10 include `CLAUDE.md` and `docs/workflow-safety-and-approvals.md` — two of the documents the
Agentic OS routine names as key files. So a contributor following the documented procedure runs the
`--write` form, sees it succeed, reformats ten files, and gets rejected by CI. The document that
causes the failure and the document that warns about it were both in the tree, which is why reading
carefully could not have caught it.

(The two `.pytest_cache/README.md` hits both versions report are untracked local artifacts and are
excluded from the counts above.)

## What ships

1. The README's two lines now say `3.8.1`.
2. `tests/workflow-logic/test_prettier_version_is_documented_consistently.py` — parses the pin out
   of `722-ci.yml` **at run time** and requires every `prettier@<version>` in a **tracked** file to
   equal it.

It is a set comparison over the tree, not a hand-list of files (ledger L16/L62), so a new mention in
a new file is covered the day it lands and fixing one mention without its neighbours still fails.
The scan matches **raw bytes**, so a tracked binary cannot raise `UnicodeDecodeError` at import on a
cp1252 host (ledger L07). The module names no version literal of its own — the expected value is
parsed — so bumping the pin needs no edit here, and the module sits inside its own denominator
rather than being exempted from it.

## Discrimination proof

Run before the README fix, against the real drift — **it failed, naming both lines**:

```
FAIL test_every_documented_prettier_version_matches_the_ci_pin: 722-ci.yml checks the tree with prettier 3.8.1, but these mention a different version.
  .github/workflows/README.md says 3.3.3, CI pins 3.8.1
  .github/workflows/README.md says 3.3.3, CI pins 3.8.1
```

Then two mutations of the shipped code (anchor count asserted `== 1` first, per ledger L139 — a
6-occurrence anchor is how a review over-credits coverage):

| mutation                                        | expected                   | observed                                                                                                            |
| ----------------------------------------------- | -------------------------- | ------------------------------------------------------------------------------------------------------------------- |
| `722-ci.yml` pin bumped `3.8.1` → `3.9.9`       | comparison fails           | **FAIL**, and it named the full set — `README.md` ×2, `CLAUDE.md`, `.claude/hooks/post_edit.py` — not just the README |
| prettier invocation deleted from `722-ci.yml`   | fails **closed**, not vacuous | **FAIL `test_the_ci_pin_is_discoverable`**                                                                          |

The second mutation is the one that matters. A consistency test that quietly stops finding one side
of its comparison passes vacuously and reads as coverage — ledger **L91**/#991 is the same shape
(three artifacts agreeing perfectly while nothing asks the authority). `test_the_ci_pin_is_discoverable`
and `test_the_scan_actually_reaches_the_documentation` exist so that neither side can silently
become empty.

Both mutations were applied in-tree and reverted with `git checkout --`, and `git status` afterwards
shows `722-ci.yml` unmodified — restoring through git rather than `write_text()` avoids the
CRLF-rewrite trap CLAUDE.md documents, where a restored file hashes identical and is still modified.

## Verification

- new module: 3/3 PASS after the fix, 1 FAIL / 2 PASS before it
- `test_lessons_ledger.py`: 26/26 — this is the L07 encoding guard, which scans
  `tests/workflow-logic/*.py` and therefore covers the new module
- `npx --yes prettier@3.8.1 --check .github/workflows/README.md` — clean

Found during conductor run 101 while resolving an unrelated review thread on #1082.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -230,8 +230,8 @@ This workflow ensures that:
 
 ### Local usage
 
-- Prettier (check): `npx --yes prettier@3.3.3 --check . --ignore-unknown`
-- Prettier (write): `npx --yes prettier@3.3.3 --write . --ignore-unknown`
+- Prettier (check): `npx --yes prettier@3.8.1 --check . --ignore-unknown`
+- Prettier (write): `npx --yes prettier@3.8.1 --write . --ignore-unknown`
 - actionlint (requires Go):
   - Install: `go install github.com/rhysd/actionlint/cmd/actionlint@latest`
   - Run: `actionlint`
--- a/tests/workflow-logic/test_prettier_version_is_documented_consistently.py
+++ b/tests/workflow-logic/test_prettier_version_is_documented_consistently.py
@@ -1,0 +1,144 @@
+"""Every documented `prettier@<version>` must be the one 722-ci.yml actually checks with.
+
+CLAUDE.md already warns that formatting with a prettier other than the CI-pinned
+one produces local-pass/CI-fail loops, because prettier's Markdown reflow changes
+between minor releases. On 2026-08-05 (conductor run 101) the repo was handing out
+exactly that failure in its own contributor documentation:
+
+    .github/workflows/README.md, "Local usage"
+      - Prettier (check): `npx --yes prettier@3.3.3 --check . --ignore-unknown`
+      - Prettier (write): `npx --yes prettier@3.3.3 --write . --ignore-unknown`
+
+while `722-ci.yml` checked with 3.8.1. Measured on that tree, the difference is
+not latent: the CI-pinned version reported the tracked tree clean, and the version
+the README prescribes wanted to reformat **10 tracked files** — including
+`CLAUDE.md` and `docs/workflow-safety-and-approvals.md`, i.e. two of the documents
+the Agentic OS routine names as key files. So a contributor who followed the
+documented local procedure would rewrite ten files, see the `--write` succeed, and
+have CI reject the result. The document that causes the failure and the document
+that warns about it were both in the tree, which is why a reader could not have
+caught this by being careful.
+
+The guard is a set comparison, not a hand-list of files (ledger L16/L62: remaining
+scope must be an exact test over the tree). It scans **every tracked file** for the
+`prettier@<version>` spelling and requires each to equal the version parsed out of
+722-ci.yml. That way a new mention in a new file is covered the day it lands, and
+fixing one mention without fixing its neighbours still fails.
+
+Two properties this module is careful to have, both of them lessons the repo has
+already paid for:
+
+  * **It fails closed if it can no longer find the pin.** A consistency test that
+    silently stops locating one side of the comparison passes vacuously and reads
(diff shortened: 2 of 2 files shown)


clarkemoyer:
**Correction to this PR's own body: the claim "names no version literal of its own" was false when I wrote it, and CI caught it.**

`Validate Repository` went red on the first push, and the finding was **this module's own docstring** — it quoted the README's two bad lines verbatim, `prettier@` prefix and all, so the guard rejected the file introducing the guard.

The interesting part is why the local run said otherwise. `_tracked_files()` reads `git ls-files`; when I ran the module (and both mutations) it was **still untracked**, so the scan could not see it. `git add` is what put it inside its own denominator. The local green was not a weak signal, it was a false one, and the mutation proof inherited the same blind spot — which is worth saying plainly, because the whole point of that section was that the module participates in its own scan.

Fixed in `d0113f5`: the quote breaks the spelling, and the docstring now records both the incident and the general rule — **a guard whose denominator is the tracked tree must be re-run after `git add`**. Re-verified in that order: tracked = YES, 3/3 PASS, `prettier@3.8.1 --check . --ignore-unknown` clean apart from two untracked `.pytest_cache` artifacts that are not in git.

The substantive finding and the measurements in the body above are unaffected.

clarkemoyer:
Accepted, and measured before changing anything rather than taking the conclusion on faith: **773 tracked files, 18 MB, ~101 ms per pass**, and the module made two of them plus two `_ci_pin()` reads. Small in absolute terms today, but it scales with the repo and runs in every CI job, so it is worth the three lines.

Both helpers are now `lru_cache`d and `_mentions()` returns a tuple so the cached value is immutable. Suite: **~200 ms -> 129 ms**.

One thing worth recording rather than silently doing: a cache in a *guard* has a hazard the general case does not. It is safe here because its lifetime is one process and the tree cannot change underneath a single run — but it would be **unsafe** for a reviewer driving this module in-process across a mutation of a scanned file, which is exactly how someone might try to check it. That caveat is now in the `_mentions()` docstring next to the measurement.

Re-proved discrimination after the change, as fresh processes: bumping the pin in `722-ci.yml` still fails the comparison, deleting the invocation still fails `test_the_ci_pin_is_discoverable` closed, and `git status` shows `722-ci.yml` unmodified afterwards. 3/3 PASS clean.

Issue #1091, opened 2026-08-06
727's behind-count counts merge commits and doc-only churn: two ledger PRs red-line every open PR


clarkemoyer:
## Summary

`727-phantom-revert-guard.yml` fails a PR when `BEHIND_COUNT > 5`, where `BEHIND_COUNT` counts
**every** commit `main` gained since the merge base — merge commits and documentation-only commits
included. Measured against the live repo, that budget is consumed by **two** Conductor lessons-ledger
PRs, which is one every ~two runs. The whole open PR queue is red-lined on a cadence, for no defect
of any PR.

The 06:34Z cloud-worker pass on #719 found the symptom and reported the cause as "a burst of 4
ledger PRs in ~2h". Re-measured here: it was **two** PRs. The difference is not cosmetic — it decides
whether batching the ledger is a fix (at 4 it looks like one; at 2 it is not).

## The measurement

`727-phantom-revert-guard.yml:160` and `:194` / `:247`:

```bash
BEHIND_COUNT=$(git rev-list --count "$MERGE_BASE..$BASE")
...
if [ "$BEHIND_COUNT" -gt 5 ]; then   # :194, fires even with zero phantom candidates
```

Four open PRs sat at `behind=6` before the cloud worker updated them. Their shared merge base was
`ba2849b`. `git rev-list --count ba2849b..d96230a` = **6**, and those six commits are:

| commit | PR | kind |
| --- | --- | --- |
| `834d364` | #1088 | branch commit |
| `13a680f` | #1088 | **merge commit** |
| `b1faf84` | #1090 | branch commit |
| `2b86eb9` | #1090 | branch commit — `style(lessons): prettier re-pads the ledger table` |
| `68f8227` | #1090 | branch commit — review-comment fixups |
| `d96230a` | #1090 | **merge commit** |

`git diff --name-only ba2849b d96230a` returns exactly one path: **`docs/lessons-ledger.md`**.

So the multiplier has two independent halves, and both are measurable:

1. **The merge queue lands merge commits, not squashes.** `git rev-list --parents -n1 13a680f` and
   `d96230a` each return two parents. Every PR therefore contributes `1 + N` to `BEHIND_COUNT`,
   never 1.
2. **Ledger PRs routinely carry fixup commits.** #1090 needed a Prettier re-pad and a
   review-comment pass, so a 1-row-idea PR became 4 commits on `main`.

[...]

PR #1092, opened 2026-08-06
docs(lessons): L153-L154 — a behind-count that measures merge strategy, and a list whose element 0 is not the latest

clarkemoyer:
Conductor run 105, step 7. Two rows, and — for the second — the enforcement that lets it be a real path instead of `doc —`.

## L153 — 727's `BEHIND_COUNT` counts merge commits and doc-only churn

The 06:34Z cloud-worker pass on #719 found four open PRs one CI run away from a red `Phantom Revert Guard` and reported the trigger as *"a burst of 4 ledger PRs in ~2h"*. Re-derived here before inheriting it (L150), and it was **two**:

| commit | PR | kind |
| --- | --- | --- |
| `834d364` | #1088 | branch commit |
| `13a680f` | #1088 | **merge commit** |
| `b1faf84` | #1090 | branch commit |
| `2b86eb9` | #1090 | branch commit — Prettier re-pad |
| `68f8227` | #1090 | branch commit — review fixups |
| `d96230a` | #1090 | **merge commit** |

`git rev-list --count ba2849b..d96230a` = 6; `git diff --name-only ba2849b d96230a` = `docs/lessons-ledger.md`, one path. `git rev-list --parents -n1` on both `13a680f` and `d96230a` returns two parents, so the queue merges rather than squashes and every PR contributes `1 + N`.

The count matters because the fix follows from it. At 4 PRs, "batch the ledger" looks like a remedy; at 2 it is not — the Conductor ships one ledger PR per run and runs are ~3h apart, so two consecutive runs are enough. Filed as **#1091** against the workflow's own expression rather than against the audit that mirrors it.

Deliberately **not** L145's point: that row is about `scripts/audit-open-pr-readiness.py` modelling one clause of a disjunction. This one is about the clause itself being calibrated against a commit mix this repo does not have. The row says so, so a reader does not merge them.

## L154 — a list field's element 0 is not "the latest"

Reading the live feed this run, `conductor_log[0]` was `2026-08-05T19:41:56Z` against a `generated_at` of `04:10:56Z` — a log panel apparently 8.5h behind its own container. `collect_conductor_log()` sorts by `created_at` and slices `raw[-CONDUCTOR_LOG_LIMIT:]`, so element 0 is the **oldest**; the true newest was `04:05:28Z`, five minutes before generation. Correct all along. That is the L149 trap one step past where L152 left it, and a finding was half-written before the generator was opened.

What made it avoidable: **every other panel in this feed states its own semantics in the data** — `ready_rule`, `in_flight_prs_rule`, `pending_gates_scope`. `conductor_log` was the one list shipping with no rule beside it, and it is the only field whose elements carry their own timestamps, i.e. the only one a reader can and will cross-check against `generated_at`.

So this PR adds `conductor_log_rule` in that established pattern, and the row is enforced by a path rather than prose.

### Mutation-tested, not just run

The new assertions pass by discrimination. Each mutation asserts its anchor is present before substituting, so a refactor fails loudly instead of silently testing nothing:

| mutation | result |
| --- | --- |
| unmodified | exit 0, all assertions passed |
| `recent = raw[-LIMIT:]` → `list(reversed(raw))[:LIMIT]` | exit 1 — `conductor_log is oldest-first` |
| rule reworded to state the direction but drop both endpoints | exit 1 — `the rule names both ends, not just the direction` |

The generator was restored from a byte-copy after each, verified by `sha256sum` (`b9194e08…` before and after both) rather than by `read_text`, per the CRLF note in `CLAUDE.md`.

## Checks run locally

- `tests/workflow-logic/test_lessons_ledger.py` — exit **0**, 26 PASS, against the post-Prettier tree
- `tests/workflow-logic/test_502_agentic_os_status.py` — exit **0**
- `uniq -d` over the ledger IDs — empty; **149** rows, max id **L154**
- `npx --yes prettier@3.8.1 --write docs/lessons-ledger.md` — 2 lines added, no table re-pad this time (the new rows do not widen the widest cell)


clarkemoyer:
Applied in `ef7ba4f`. The strict `<` stays, because it is the only form that rules out a **reversed** list — `<=` is satisfied by any all-equal sequence in either direction, so relaxing it would remove the discrimination the assertion exists for (verified by mutation: reversing `collect_conductor_log`'s slice fails `conductor_log is oldest-first`, and `stamps == sorted(stamps)` alone would not catch a reversal of tied stamps either).

What the review correctly identifies is an unstated dependency on the fixture. That is now asserted explicitly, one line earlier:

```python
check(stamps[0] != stamps[-1],
      "fixture precondition: the log's endpoints must differ for the strict check below")
```

So a future fixture edit that introduces tied endpoints fails with the reason named, rather than surfacing one line later as a bogus ordering fault. Test re-run: exit 0.

Issue #966, opened 2026-07-31
Script the Agentic OS board audit — hand-rolled, and wrong twice in two runs


clarkemoyer:
## Problem

The public "Agentic OS" board (org project #9) has **no auto-add** — every item and every status is placed by hand (`CLAUDE.md`, "Board & PR-creation env facts"). So each Conductor run re-audits it by hand: which open `agentic-os` items are missing from the board, which items have no Status, which are closed but not `Done`. It is the same query every run, it is easy to get wrong, and it has now been gotten wrong twice in two runs in **two different ways**:

1. **Run 61** — `items(first:100)` returned 100 of 208 and nearly produced "0 statusless" off a third of the board. Fixed by paginating on `$endCursor`.
2. **Run 62** — the *other* half of the comparison was wrong. Building the "what should be on the board" set from `gh search issues --label agentic-os` returned **52** items and reported **0 missing**. The authoritative REST enumeration returned **53**, and the one it omitted was PR #965 — created minutes earlier. `gh search` is backed by an **index that lags writes**, so a completeness audit built on it under-reports precisely for the newest items, which are exactly the ones most likely to be missing from a manually-maintained board. It fails silently and in the reassuring direction.

Both halves have to be right for the audit to mean anything, and neither is obvious from reading the command.

## Scope (one PR)

Add **`scripts/audit-agentic-os-board.py`** — read-only, no writes, no mutations.

Inputs: `GH_TOKEN` (or `GITHUB_TOKEN`), as `scripts/generate-agentic-os-status.py` already does — follow that script's auth and REST conventions; it is the model for this one.

It must:

- Enumerate expected items from **REST** — `GET /repos/FreeForCharity/<repo>/issues?state=open&labels=agentic-os&per_page=100`, paginated, across the same repo set `generate-agentic-os-status.py` sweeps. **Not `gh search`, and not the search API** — a comment must say why, citing this issue.
- Enumerate board items via GraphQL `projectV2(number:9){items}` paginated with `$endCursor` until `hasNextPage` is false.
- Report three sets: **missing from board**, **on board with no Status**, **closed/merged but not `Done`**.
- Print counts for both sides of the comparison (`expected=N board=M`) so a future silent under-report is visible as a suspicious denominator rather than an empty result.
- Exit non-zero when any set is non-empty, so it can be wired into CI later. Exit non-zero on an API/auth error too — **never** exit 0 on a failed enumeration.

## Acceptance criteria

[...]

PR #1105, opened 2026-08-07
docs(lessons): L169 hub-scoped PR supervision, L170 out-of-band monitors, L171 one-fetch publish verification

clarkemoyer:
Two ledger rows from conductor run 116. Both were found by the same review pass and both describe a check that returns a **clean answer about the wrong population** — the shape L62 and #966 already name for the board audit, arriving at two new surfaces.

- **L169** — the open-PR supervision sweep has been scoped to the hub. Run 115 reported `4 open, 0 findings` and the 15:24Z worker triaged 3; `org:FreeForCharity is:pr is:open label:agentic-os` returns **10**. The 7 uncounted PRs live in the two templates and `FFC-EX-canary`, are all green, and two clusters of them carry cross-repo ordering hazards no per-PR check can see.
- **L170** — `738. Fleet Smoke Engine Drift Audit` reports through rolling issue #893, not through its conclusion. Its 2026-08-03 run is recorded `success` while #893 lists **61 divergent repos**. This run's per-scheduled-workflow health sweep (the L61/L33 method) scored it green on the same pass that a PR review surfaced the drift.

Neither is mechanically enforceable as written, so both ship as `doc —` with the mechanical half named in the Enforced-by cell: an org-wide `expected=N counted=M` supervision sweep for L169, and having the health sweep read the marker-bearing rolling issues for L170.

## Verification

- `npx --yes prettier@3.8.1 --write docs/lessons-ledger.md` **before** committing (L165's error two runs ago).
- `python tests/workflow-logic/test_lessons_ledger.py` → **26 passed, 0 failed**, run after prettier.
- No `|` inside any cell. IDs unique — `grep -oE '^\| L[0-9]+ ' ... | sort | uniq -d` is empty; `main`'s highest was re-checked (L168, #1104) before numbering.
- Rows inserted by a script that asserts exactly one `L168` anchor and refuses if either new ID already exists, so a silent no-op cannot pass as a write.

Draft pending CI. Conductor run 116.


clarkemoyer:
**Added L171 in a second push.** It came out of this run's own delivery verification, so it is measured rather than recalled.

Delivery 51 (ffcadmin#854) was checked seconds after `Deploy to GitHub Pages` reported success, with three consecutive `curl`s. The first returned `27927b33…` against a generated file of `bcf7d2ace6df5c58…`; the next returned the **new** `generated_at` of `2026-08-07T16:18:28Z`. Both were true of the response they received — and they had reached **different edge nodes**, one still serving delivery 50.

Re-fetched once into a file a minute later: `diff` against `HEAD:public/data/agentic-os-status.json` → **0 lines**, 80509 bytes, 0 CR, and generator / `HEAD:src` / `HEAD:public` / served all hash `bcf7d2ace6df5c58…`. Delivery 51 was correct the whole time; the *verification* was reading four responses as one.

This run got the alarming direction. The same race yields a **false pass** whenever the propagated edge answers the `generated_at` check and a stale one answers the content check — which is the reading a delivery check is written to produce, so it would not be questioned. L162 already says read the artefact rather than the page; L171 adds that "the artefact" must be **one fetched copy**.

Re-verified after this push: prettier@3.8.1 first, `test_lessons_ledger.py` **26 passed**, no duplicate IDs, no `|` inside any cell.

clarkemoyer:
Good catch, and it is the row's own failure mode — L169 is about a count taken from the wrong population, and I sourced its headline number from the index L62 forbids for exactly this question. Fixed rather than caveated.

Re-derived authoritatively: open PRs per non-archived org repo via REST, paginated, filtered on the label. That enumeration returns **10** and names all 7 uncounted PRs (`FFC-IN-FFC_Single_Page_Template#433/#432/#428`, `FFC-IN-Footer_Only_Template#124/#123`, `FFC-EX-canary#22/#21`). The row now cites that sweep as the authority and says plainly that `gh search` agreeing at 10 is a coincidence worth nothing.

Worth recording: the first REST sweep I wrote returned **0**. Not the data — `bc` is absent on this host, so `paste -sd+ | bc` produced empty and the shell read it as zero. A third instance, in one run, of a check answering confidently about something other than what it measured.

`test_lessons_ledger.py` 26 passed after prettier@3.8.1.

Issue #1074, opened 2026-08-05
229 + 230: split the dry-run preview onto whmcs-prod-read so a rehearsal needs no gate


clarkemoyer:
Split out of a Copilot finding on #825 ([thread](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/825#discussion_r3642260960)), which was real but wider than that PR's scope.

## The problem

`229-whmcs-client-field-populate.yml` runs its single `populate` job on the **gated** `whmcs-prod` environment with `scope: write`, even when `dry_run=true`. So a preview — which sends no `UpdateClient` and changes nothing — still parks at an approval gate and still loads a write-scoped Key Vault credential.

That has two costs. The rehearsal that is supposed to be cheap needs a human, which is exactly the step you want an operator to run freely before a live write. And it holds write credentials for a read-only operation, which is the narrowing `#834` established as the pattern for `726`.

`230-whmcs-record-field-set.yml` is the same shape and sits next to it in `KNOWN_UNGUARDED`.

## Why it is not just "add an `if`"

The dry run genuinely reads WHMCS — `GetClientsProducts` + `GetClientsDetails` build the preview — so it needs a credential, just a **read-scoped** one. A skip-condition guard is therefore the wrong tool; the job cannot be skipped, it has to be **relocated**.

Both workflows are enumerated in `tests/workflow-logic/test_dry_run_skips_write_gates.py` (`KNOWN_UNGUARDED`) with that rationale, so the exemption is deliberate and documented rather than an oversight.

## The shape to copy

`701-website-provision.yml`'s `dns` / `dns_preview` relocation:

- a `preview` job on the ungated **`whmcs-prod-read`** lane with `scope: read`, which always runs and produces the report + artifact;
- an `execute` job on the gated **`whmcs-prod`** lane with `scope: write`, conditioned on `inputs.dry_run == false`, carrying only the `UpdateClient` call.

## Acceptance criteria

- [ ] `229` dispatched with `dry_run=true` completes with **no** deployment gate and still uploads `whmcs_client_field_populate.json`.
- [ ] `229` dispatched with `dry_run=false` still stops at the `whmcs-prod` gate before any `UpdateClient`.
- [ ] The preview job requests `scope: read` and references the `READ_ALL_FFC_AZURE_*` repo Variables (not `secrets.*` — `scripts/check-env-secret-references.py` fails CI otherwise).
[...]

Issue #983, opened 2026-08-01
dry_run is enforced per-step, not per-job: every dry run of 701/702/720 parks on a write gate and mints a write credential


clarkemoyer:
## The defect

A `dry_run=true` dispatch of **701**, **702** or **720** still parks on a **write-environment approval gate** and still loads a **write-scoped credential** from Key Vault. `dry_run` is enforced at the *step* level in these workflows, never at the *job* level — so the job starts, the gate fires, the write token is minted, and only then do the individual steps decide to do nothing.

Measured from the job graphs on `main` (`7bce4f0`) — six jobs, zero of them gated on `dry_run`:

| Workflow | Job | `environment` | `if:` mentions dry_run | credential it loads |
| --- | --- | --- | --- | --- |
| 701 | `dns` | `cloudflare-prod-write` | **no** | `cloudflare-tokens-from-kv` `scope: write` |
| 701 | `repo` | `github-prod` | **no** | `wr-all-cbm-ffc-copilot-mcp-github-pat` |
| 701 | `content` | `github-prod` | **no** | `wr-all-cbm-ffc-copilot-mcp-github-pat` |
| 702 | `clone-deploy` | `github-prod` | **no** | `wr-all-cbm-ffc-copilot-mcp-github-pat` |
| 720 | `create-repo` | `github-prod` | **no** | `wr-all-cbm-github-pat` |

701's `zone_check` is the counter-example that shows this is not "all gates are like that": it is on `cloudflare-prod-read` with a `scope: read` token, so it runs unattended, dry or not.

**111 already does it correctly.** `.github/workflows/111-dns-create-redirect-rule.yml:80` reads `if: inputs.dry_run == false` on its `apply` job, so a dry dispatch never reaches `cloudflare-prod-write` at all. That is the pattern to copy.

## How it was found, and why it matters more than it looks

The Conductor has owed a post-merge verification since run 64 (#858): dispatch 701/702/720 with `dry_run=true` against a real target. It was deferred for four runs. Run 68 executed it, and the reason it had never been done became self-evident — **the dry run cannot finish without a human approving a write gate**, so it is not a check anybody can run in CI or unattended.

[...]

PR #1110, opened 2026-08-08
feat(gates): surface dry_run in run-name on the 15 gated workflows that hide it (#1107)

### LINKED ISSUE #1107: Gated write workflows hide dry_run from the approver: 16 of 29 omit it from run-name, so the read-only/dry-run approval rule is unexecutable
## What happened

At **2026-08-07T22:17:35Z**, mid-run, a `106 / cloudflare-prod-write` approval appeared
([run 31223295765](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/31223295765),
`Enforce Standard: slopestohope.org`, dispatched by @clarkemoyer on `main` at `ec5a689`).

The Conductor's standing rule is: **auto-approve only read-only environments and `dry_run=true` runs.**
Applying it requires knowing this run's `dry_run` value. That value is **not discoverable from any
surface available before approval**:

- `106-enforce-standard.yml` declares `dry_run` with `default: true` (`:11-14`) — but a default is not
  a reading of *this* dispatch, and the whole point of the input is that it can be set to `false`.
- Its `run-name` is `'Enforce Standard: ${{ inputs.domain }}'` (`:2`) — domain only.
- The REST run object carries **no** `workflow_dispatch` inputs at all (checked: no input-bearing key
  on `actions/runs/31223295765`).
- The job is `waiting`, so there are no logs to read the value out of. That is the point of a gate.

So the approver — human or agent — is asked to approve a **write** environment while blind to the one
input that decides whether it writes. The gate was **held**, which is the correct-but-useless outcome:
the rule is unexecutable here, so the answer is always "hold", including for runs that are pure
rehearsals and were meant to sail through.

## Scope — 16 of 29, against an existing convention

Swept all **103** workflow files in `.github/workflows/`. Of these, **29** are gated (`environment:`)
**and** take a `dry_run` input. **13 already put it in `run-name`; 16 do not.**

The convention to copy already exists and is used by those 13 — e.g.
`736-repo-archive.yml`: `'Archive ${{ inputs.repo }} (dry_run=${{ inputs.dry_run }})'`.

**The 16 undecidable ones:**

| workflow | current `run-name` |
| --- | --- |
| `103-enforce-domain-standard.yml` | `'Enforce Domain Standard: ${{ inputs.domain }}'` |
| `105-manage-record.yml` | `'Manage Record: ${{ inputs.domain }} (...)'` |
| `106-enforce-standard.yml` | `'Enforce Standard: ${{ inputs.domain }}'` |
| `111-dns-create-redirect-rule.yml` | `'Redirect Rule: ... '` |
| `204-whmcs-charity-onboard.yml` | `'WHMCS Charity Onboard'` |
| `205-whmcs-ticket-open.yml` | `'WHMCS Open Ticket'` |
| `207-whmcs-ticket-respond.yml` | `'WHMCS Ticket Respond #${{ inputs.ticket_id }}'` |
| `211-whmcs-order-update.yml` | `'WHMCS Order Update #${{ inputs.order_id }} (...)'` |
| `212-whmcs-product-add.yml` | `'WHMCS Product Add (${{ inputs.product_key }})'` |
| `228-whmcs-fraud-review.yml` | `'WHMCS Fraud Review (FraudLabs Pro)'` |
| `229-whmcs-client-field-populate.yml` | `"WHMCS client-field populate (...)"` |
| `230-whmcs-record-field-set.yml` | `'WHMCS Record Field Set: ...'` |
| `305-m365-add-tenant-domain.yml` | `'M365: Add domain ${{ inputs.domain }}'` |
| `701-website-provision.yml` | **no `run-name` directive** |
| `702-ffc-ex-clone-deploy.yml` | `"Clone+deploy ${{ inputs.domain }} -> ..."` |
| `729-repo-add-collaborator.yml` | **no `run-name` directive** |

## Why this is worth a PR rather than a note

This is the **inputs** axis of the same blindness #1065 documented on the **ref** axis — and, by
coincidence worth noting, on the same workflow (`106`) and the same domain (`slopestohope.org`).
#1065's run was on an unmerged PR branch; today's is on `main` at `ec5a689`, so that defect did not
recur — but the approver is still guessing. Together they say: *the approval UI shows you a workflow
name and nothing that decides what it does.* `run-name` is the one field that reaches the gate, and
it is free.

Related but not duplicates: **#983** (dry runs still park on a write gate — about *whether* a gate
should exist), **#1074** (splitting 229/230's preview onto a read lane). This issue is only about
making the value **visible** on the gates that already exist.

## Scope of the change

One PR. Edit `run-name` in the 16 files above to append the existing convention:

```
(dry_run=${{ inputs.dry_run }})
```

For `701-website-provision.yml` and `729-repo-add-collaborator.yml`, add a `run-name` directive (they
have none) following the pattern of the other 13.

Do **not** change any job logic, permissions, `environment:` value, or the inputs themselves. This is
a display-string change only.

## Acceptance criteria

1. All **29** gated workflows with a `dry_run` input surface it in `run-name`; the sweep below reports
   `UNDECIDABLE: 0`.
2. No file outside `.github/workflows/` changes; no `permissions:`, `environment:`, or step body is
   touched (`git diff` shows only `run-name` lines added/modified).
3. `docs/workflow-safety-and-approvals.md` gains one line in its conventions section stating that a
   gated workflow taking `dry_run` MUST surface it in `run-name`, so the approver can apply the
   read-only/dry-run rule without guessing.
4. Existing `run-name` content is preserved — the marker is appended, not substituted.

## Verification

```bash
python - <<'PY'
import os,re,glob
files=sorted(glob.glob('.github/workflows/*.yml')+glob.glob('.github/workflows/*.yaml'))
bad=[]
for f in files:
    t=open(f,encoding='utf-8').read()
    if 'environment:' in t and re.search(r'^\s{4,}dry_run:',t,re.M):
        m=re.search(r'^run-name:\s*(.+)$',t,re.M)
        if not m or 'dry_run' not in m.group(1):
            bad.append(os.path.basename(f))
print('files',len(files),'UNDECIDABLE',len(bad));  [print(' ',b) for b in bad]
PY
```

Expected after the fix: `UNDECIDABLE 0`. Run it before the change too — it must report **16** on
`main` today, which is how you know the check is actually looking at something.

Consider promoting that snippet into `tests/workflow-logic/` so the property is enforced at PR time
rather than re-measured by hand each run — that is the stronger tier and is in scope for this PR.

## Gates this will hit

None. Editing `run-name` strings does not dispatch anything and touches no environment. Normal
branch + draft PR + merge queue.

_Filed by Conductor, local, run 118._

clarkemoyer:
Closes #1107.

## What this changes

Every gated workflow that takes a `dry_run` input now shows its value on the approval screen:

```
Enforce Standard: slopestohope.org (dry_run=true)
```

The Conductor's auto-approval rule (`docs/workflow-safety-and-approvals.md`) turns on **this
dispatch's** `dry_run`. Before this PR, 15 gated workflows made that unanswerable — the input's
`default:` describes the form and not the run, the REST run object exposes **no** `workflow_dispatch`
inputs at all, and a `waiting` job has produced no logs, which is the entire point of a gate. So the
rule's answer was always "hold", including for rehearsals that were meant to sail straight through
(L177, and the live `106 / cloudflare-prod-write` gate that is still held right now).

Display strings only. No job logic, `permissions:`, `environment:` value, or input declaration is
touched — `git diff` is 15 `run-name` lines, one doc paragraph, and one new test module.

## The issue's count was wrong, and how

**#1107 says 16. The real number is 15.** `229-whmcs-client-field-populate.yml` already complied:

```yaml
run-name:
  "WHMCS client-field populate (${{ inputs.client_id != '' && … }}, dry_run=${{ inputs.dry_run }})"
```

Its value begins on the **line after** the key, and #1107's verification snippet — mine, from run
118 — matches `^run-name:\s*(.+)$`, which cannot see a continuation line. A single-line regex over
YAML answers a question about the wrong population (L172's shape, one file format over). Re-derived
with `yaml.safe_load`: **29 gated-with-`dry_run`, 14 already surfacing, 15 undecidable.** That exact
shape is now a fixture in the new module, so a future rewrite back to a regex fails here rather than
in a run summary.

## 701 renders the *effective* value, not the input

701 also fires on `issues` and `repository_dispatch`, where `inputs` is null and
`${{ inputs.dry_run }}` renders empty. Its own write gate is `inputs.dry_run != true`
(`701-website-provision.yml:1045`), so a null reads as **live**. The run-name therefore uses
`${{ inputs.dry_run || false }}`, which renders `false` on exactly the paths the job treats as live —
the run-name agrees with the condition the job evaluates rather than going blank on the one event
class where being blank is most misleading. Documented in the safety doc.

729 had no `run-name` either; it is `workflow_call` + `workflow_dispatch`, and a reusable call renders
the **caller's** run-name, so the new string is only ever seen on a direct dispatch — which is exactly
the path that reaches its `github-prod` gate.

## Verification

```
$ python tests/workflow-logic/test_dry_run_visible_in_run_name.py
  PASS test_a_multiline_run_name_is_read_whole
  PASS test_an_ungated_dry_run_workflow_is_not_reported
  PASS test_every_gated_dry_run_workflow_surfaces_it_in_run_name
  PASS test_the_guard_has_a_population_to_check
  PASS test_the_guard_reports_a_workflow_that_hides_it
  PASS test_the_safety_doc_states_the_convention
```

**Mutation-proved, not merely wired** (AGENTS.md). Reverted `106`'s run-name to its pre-PR value and
re-ran:

```
FAIL test_every_gated_dry_run_workflow_surfaces_it_in_run_name: 1 gated workflow(s) take a
dry_run input and do not surface it in run-name, so an approver cannot apply the auto-approval
rule to them: 106-enforce-standard.yml. Append `(dry_run=${{ inputs.dry_run }})` to the existing
run-name (see docs/workflow-safety-and-approvals.md).
```

rc=1, naming the file. Tree restored; the anchor count was asserted before the substitution so a
silent no-op could not read as a hole (run 62 / #965).

Also checked:

- **Every folded `run-name` still resolves.** `prettier` wraps some values across lines, splitting
  `${{` from `inputs.dry_run }}`; YAML folds the newline to a space, so all six reflowed workflows
  load as one correct expression. Printed each parsed value to confirm rather than reasoning about it.
- `scripts/check-workflow-doc-consistency.py` → `OK: 103 workflows, 96 table rows covered`.
- `npx --yes prettier@3.8.1 --check` clean on all 16 touched files (the CI-pinned version, per
  CLAUDE.md — not bare `prettier@3`).
- **0 CRLF** in every changed file, verified as bytes (`'rb'`, not `read_text`).

## Acceptance criteria

| AC | | |
| --- | --- | --- |
| 1 | `UNDECIDABLE: 0` | ✅ `gated_with_dry_run=29 surfacing=29 UNDECIDABLE=0` |
| 2 | only `run-name` lines change | ✅ 15 workflow files, `run-name` lines only |
| 3 | safety doc states the convention | ✅ new paragraph under the auto-approval policy |
| 4 | existing content preserved, marker appended | ✅ appended in all 13 with an existing value; 701/729 gained one |
| bonus | promote the snippet into `tests/workflow-logic/` | ✅ and it is now YAML-parsed, which the snippet was not |

## What this does *not* do

It does not remove a gate or change what any run does. The `106` gate held right now stays held —
this PR only means the **next** one is decidable. `#983` (should a dry run park on a write gate at
all) and `#1074` (splitting 229/230's preview onto a read lane) are the questions about whether these
gates should exist; this is only about being able to answer the rule.


clarkemoyer:
Correct, and the shape is **live in this repo** — I went looking rather than accepting the general claim.

```
jobs using the mapping form: 1
  ('721-deploy-pages.yml', 'deploy', {'name': 'github-pages', 'url': '${{ steps.deployment.outputs.page_url }}'})
```

So `_gated_environments` could genuinely return a dict where the annotation promises `str`. Nothing failed, because the only consumer is a truthiness test and a dict is truthy — which is exactly what makes it worth fixing now: the guard was right for a reason it did not state, one refactor away from being wrong.

Fixed in `a9cefe2` by mirroring `_job_environment_names` in `test_gated_env_hygiene.py` (mapping → `name` when it is a non-empty string; bare value → `str(...)`), rather than inventing a second convention for the same YAML shape.

Added `test_the_mapping_form_of_environment_is_a_gate_too`, which asserts the extracted **name** — `_gated_environments(doc) == ["github-prod"]` — not merely that something truthy came back, and then asserts the workflow is still reported as undecidable. Mutation-proved: reverting `_gated_environments` to `names.append(env)` fails that case and only that case.

7 tests pass, 0 CRLF.

Issue #1111, opened 2026-08-08
Board audit cannot see work that merges before it is carded: #1108 merged uncarded and every category reported 0


clarkemoyer:
## What

`scripts/audit-agentic-os-board.py` builds its expected set from **open** items only
(`collect_expected`, `:339` — `params={"state": "open", "labels": LABEL, ...}`). Its three findings
categories are:

1. `missing_from_board` — derived from the **expected** set, so open items only;
2. `no_status` — derived from **board rows**, so carded items only;
3. `closed_not_done` — also derived from **board rows**, so carded items only.

An `agentic-os` PR that is **merged before it is ever carded** is in none of them. It is not open, so
(1) cannot see it. It has no card, so (2) and (3) cannot see it. The audit reports `0 / 0 / 0` and
exits 0, and the public board permanently has no record that the work happened.

The `--grace-minutes` window makes this reachable rather than theoretical: an item younger than 90
minutes is deliberately excluded from (1) so latency does not red-line the run. Any PR that merges
inside that window — or between the last audit and its merge — falls straight through.

## Measured, on this repo, this run

**#1108 is the live case.** Opened `2026-08-07T22:35:07Z`, merged `2026-08-08T01:18:53Z`.

- Run 118's audit at ~22:4xZ classified it as *"recently opened, not yet carded"* — inside the grace
  window, correctly not a failure. Its END comment records exactly that.
- It merged 2h43m later, still uncarded.
- Run 119's audit (01:2xZ) reported **`missing from board: 1`** — #1107 only. #1108 does not appear
  in any category.

Confirmed directly rather than inferred:

```
$ gh api graphql -f query='{repository(owner:"FreeForCharity",name:"FFC-Cloudflare-Automation"){
    issueOrPullRequest(number:1108){... on PullRequest{projectItems(first:5){nodes{id project{number}}}}}}}'
NOT ON BOARD
```

A PR that mutated `docs/lessons-ledger.md` and added a guard to `tests/workflow-logic/` — real,
merged, `agentic-os`-labelled work — was invisible to the mechanism whose whole purpose is to make
the project publicly trackable. It was carded this run only because the Conductor happened to check
membership by hand while placing #1107.

## Why the current shape hides it

The audit's docstring frames the missing-card set as *"which items have no card"*, and it is honest
that grace splits it into latency vs drift. What it does not model is that an item can **leave the
expected population entirely** while the latency clock is still running. Latency is bounded for an
[...]

Issue #1103, opened 2026-08-07
Public status feed drops a whole repo's PRs when its last agentic-os issue closes: freeforcharity.org's 22 open PRs vanished 6 minutes after a security fix landed


clarkemoyer:
## What is wrong

`scripts/generate-agentic-os-status.py` decides which repositories the public feed's PR panels sweep
from **which repos currently carry an open `agentic-os`-labelled item**:

```python
def scope_repos(items, hub_repo):
    repos = {r for r in (_repo_full_name(i) for i in items) if r}
    repos.add(hub_repo)
    return sorted(repos)
```

So a repository's entire open-PR population enters and leaves the public feed as a side effect of
whether one of its issues happens to be open. **Closing the last `agentic-os` issue in a repo
silently deletes that repo's PRs from `open_prs_total` and from the in-flight panel.**

The hub is pinned unconditionally, and the docstring explains why — "so `open_prs_total` does not
silently lose its largest contributor on a day when every hub backlog issue happens to be closed."
That reasoning is correct and it applies to every other repo too; only the hub got the fix.

## Measured today, on the org's main public website

`FFC-IN-freeforcharity.org` carried exactly one open `agentic-os` issue: **#524**, the `Security
Audit` deadlock. It closed at **2026-08-07T13:08:15Z** when its fix (freeforcharity.org#525) merged.

The next feed generated **six minutes later**, at `13:14:11Z`, swept **5** repos instead of 6:

```
repos: FFC-Cloudflare-Automation, FFC-EX-canary, FFC-IN-FFC_Single_Page_Template,
       FFC-IN-Footer_Only_Template, FFC-IN-ffcadmin.org
```

| repo | open PRs | in the sweep? |
| --- | --- | --- |
| FFC-Cloudflare-Automation | 3 | yes (pinned) |
| FFC-EX-canary | 19 | yes |
| FFC-IN-FFC_Single_Page_Template | 31 | yes |
| FFC-IN-Footer_Only_Template | 18 | yes |
| FFC-IN-ffcadmin.org | 6 | yes |
| **FFC-IN-freeforcharity.org** | **22** | **no — 0 open `agentic-os` issues** |

3 + 19 + 31 + 18 + 6 = **77**, which is exactly the `open_prs_total` published in delivery 50. The
missing 22 are not stale, closed, or filtered — the repo is simply not asked.

## Why this is worse than an undercount

[...]

Issue #1095, opened 2026-08-06
Ledger path:line citations are never validated: a security row now points at a checkout step, and a 166-line-wrong pointer shipped green


clarkemoyer:
## What is wrong

`docs/lessons-ledger.md` cites source locations as `` `path:LINE` `` throughout, and **nothing
validates them**. Line numbers are never checked in any column, and the one path check that exists
reads a single column.

`tests/workflow-logic/test_lessons_ledger.py:181`:

```python
def test_a_guard_the_ledger_names_actually_exists():
    for lid, cells in _rows():
        for claimed in _claimed_paths(cells[2]):      # <-- Enforced-by column ONLY
            if not (REPO_ROOT / claimed).exists():    # <-- existence ONLY, line number discarded
```

So a citation is unchecked if it is **in the Lesson or Evidence column** (never parsed), and even in
the Enforced-by column its **line number is never compared to anything**. `_claimed_paths`
(`:132-140`) also requires a `/` in the token or a known suffix, and resolves relative to
`REPO_ROOT` — so the ledger's common bare-basename style (`727-phantom-revert-guard.yml:194`)
does not resolve at all.

This is **not** #1087. That issue is about a named guard that exists but does not *enforce* the row
— an undecidable question narrowed to a high-signal case. This is the decidable part underneath it:
does the cited location still *point at the thing the row describes*.

## Evidence — two live defects, neither caught by CI

**1. L155 cited `Update-CloudflareDns.ps1:1819` for "106 excludes DKIM outright".** `:1819` is
inside the `'AAAA'` branch of the record-matching switch and contains no mention of DKIM. The real
sites are `:65` (`.SYNOPSIS`: "DKIM is deliberately out of scope") and `:1985` (the mail-cutover
comment). Wrong by 166 lines, in a row whose conclusion rested on that citation. **Caught by
Copilot's LLM review on #1093, not by the suite** — the full 26-test ledger suite was green on it
both before and after the fix.

**2. `105-manage-record.yml:178-184` has silently drifted.** Cited by the free-text-input row
(`run 100 supervision of #825, #1080, 720-create-repo.yml:193,274, 105-manage-record.yml:178-184`).
Today `:178` is a **blank line** and `:178-184` spans `actions/checkout` plus the
`cloudflare-tokens-from-kv` action — not a `run:` body and not an interpolation. The interpolation
the row is about is now at **`:192-195`**:

```yaml
[...]

PR #1112, opened 2026-08-08
docs(lessons): L179-L180 — an open-only expected set loses finished work, and a citation grep breaks on markup

clarkemoyer:
Two rows from Conductor run 119. The third row from this run, **L178**, ships in **#1110** instead —
it names the guard that PR adds, and the ledger's own path-existence check correctly refuses a row
whose enforcement is not in the tree yet. Splitting them this way also keeps the two branches on
**disjoint tables**, so neither can conflict with the other.

### L179 — an audit whose expected set is "open items" cannot report work that finished before it was tracked

`scripts/audit-agentic-os-board.py` derives `missing_from_board` from open `agentic-os` items
(`:339`, `state=open`) and its other two categories from board rows. A PR **merged before it was ever
carded** is in none of the three: not open, so the first cannot see it; no card, so the other two
cannot.

Measured, not reasoned: **#1108** opened 22:35Z, was classified by run 118 as *recently opened, not
yet carded* (inside the 90-minute grace — correctly not a failure), merged 01:18:53Z still uncarded,
and run 119's audit then reported `0 / 0 / 0` with `projectItems` on the PR returning nothing. The
grace window is what makes it reachable: latency is bounded for an open item because a later run
catches it; closing the item closes the window, and the tolerance becomes permanent invisibility.
Filed as **#1111** with acceptance criteria. Sibling of #1103.

### L180 — a citation check that greps the quoted phrase false-negatives on inline markup

Reviewing #1108, L175's tier column cites CLAUDE.md's *"grep cannot answer it either"* bullet.
Grepping that phrase returns **nothing**, repo-wide. The bullet is real — `CLAUDE.md:540` — and reads
`` `grep` cannot answer it either ``, with backticks inside the quoted span.

The row was one command away from being "fixed" as broken, which is the expensive direction: acting
on this false negative damages something that was correct. It is a **precondition on #1095** (the
ledger's `path:line` citations are never validated) rather than a detail of it — a checker must
normalise inline markup before it may report an absence.

### Verification

- `test_lessons_ledger.py` — **30 passed, 0 failed**, including the path-existence and
  skill-name guards added by #1108.
- **173 rows, 173 unique IDs, max L180.** `main` re-read at **L177** immediately before the insert.
- Rows inserted by a script that refuses on a duplicate ID, an embedded pipe, or a newline in any
  cell, and writes with `newline="\n"` — **0 CRLF** in the committed file, verified as bytes.
- `npx --yes prettier@3.8.1` (the CI-pinned version) clean.


clarkemoyer:
Correct on both counts, and the first one makes the row stronger rather than weaker.

**On the report format.** The audit reported `missing from board: 1` — #1107 — and exited non-zero, so "a clean zero" was wrong and "0 / 0 / 0" was wrong. The precise claim, which is the one the row exists for: **#1108 appeared in no section of that report at all** — not in `missing from board`, not in `on board with no Status`, not in `closed/merged but not Done`, and not in the deferred `recently opened, not yet carded` list either. It had left every population the report describes, in a run that was otherwise loudly reporting a finding. Rewritten to say exactly that, and to note the run did not exit 0, so this is not a suppressed alarm.

**On the self-invalidating grep.** Also right, and it is a nice instance of the thing the row is about: the row's own text adds the phrase to the tree, so `grep -r` will return hits from tomorrow onward and a reader re-running the command would conclude the evidence was fabricated. Scoped to `CLAUDE.md` and to the state at the time of measurement — *"as of that commit"* — and the core claim about inline markup is unchanged.

Both fixed in `0cfdb85`. `test_lessons_ledger.py` clean, `prettier@3.8.1` clean, 0 CRLF. Note the PR had to be dequeued to accept the push (`dequeuePullRequest`, then re-enqueued) — it was blocked on these very threads and could not have merged.

PR #1114, opened 2026-08-08
docs(lessons): L183 the dry_run marker is not an approval unlock; L184 branch protection 404s under rulesets

clarkemoyer:
Two ledger rows plus the AGENTS.md recipe for one of them. Both come from run 121; both are things a run acted on wrongly before measuring.

## L183 — the `dry_run` run-name marker is documentation, not an approval unlock

#1107 made "a gated workflow that takes `dry_run` MUST surface it in `run-name`" a convention and #1110 shipped it to 25 gated jobs, so the Conductor could answer **condition 1** of the auto-approval policy for a waiting run. Run 120 then read 106's frozen run name and recorded the gate as "undecidable … it looks like #1110 should have fixed it," attributing the hold to #1110 landing three and a half hours after the dispatch.

That is true and it is not the reason. Sweeping all **61 jobs on gated environments**:

| | jobs |
| --- | --- |
| load a write-scope credential (resolved `scope: write`, or a literal `wr-all-*`) | **48** |
| …of those, skipped on a dry run by an `if:` that mentions `dry_run` | 7 |
| …of those, **reach the gate even on `dry_run=true`** | **41** |
| load no write credential by either signal | 13 |

And on the other side of the rule, the ✅ auto-approve list is **101, 104, 301–303, 306, 601** — **not one of those seven declares a `dry_run` input at all** (verified by parsing each `workflow_dispatch.inputs`).

So the two sets are **disjoint**. Every gated job whose approvability could turn on `dry_run` is already a condition-2 hold or never reaches a gate; every job that could be auto-approved has no `dry_run` to surface. The marker is worth having for the human dispatching a workflow, but no future run should expect a gate to become auto-approvable because #1110 landed.

**106 concretely:** its single job `enforce_standard` has no `if:` and passes `scope: write` to `cloudflare-tokens-from-kv`, which resolves to `wr-all` at `action.yml:68`. A post-#1110 dispatch with `dry_run=true` is the same hold.

Nothing in the guard changes — `test_auto_approval_credential_scope.py` already covers both credential signals for listed workflows, including `test_a_literal_wr_all_reference_fails_in_either_spelling`. The row exists because the *expectation* was wrong, not the enforcement.

## L184 — `branches/*/protection` returns 404 under rulesets, and the 404 fails open

Run 121 asked `repos/FreeForCharity/FFC-IN-FFC_Single_Page_Template/branches/main/protection`, got `404 {"message": "Branch not protected"}`, and **published in a PR review** that `main` was unprotected and would merge directly. `main` is governed by five active rulesets; `gh pr merge --delete-branch` then refused with `Cannot use -d or --delete-branch when merge queue enabled`, which is the only reason it surfaced.

The hub answers identically: `FFC-Cloudflare-Automation/branches/main/protection` → 404, while its effective rules on `main` are `code_quality, code_scanning, copilot_code_review, deletion, merge_queue, non_fast_forward, pull_request, repository_delete, repository_name, repository_transfer, required_status_checks`.

A 404 whose text is "Branch not protected" is a wrong answer in the **permissive** direction with nothing to notice. AGENTS.md gains the correct one-call form next to the other read-the-right-surface recipes.

Checked while writing this: **L167 is unaffected.** Its `required_approving_review_count = 0` for `FFC-IN-freeforcharity.org` is confirmed from the ruleset itself, so that row was read from the authoritative source and stands.

## Verification

`tests/workflow-logic/test_lessons_ledger.py` — **35 passed**, before and after.

Mutation-proved, each naming the expected test:

| mutation | result |
| --- | --- |
| strip L184's prose reason to a bare `` `doc` `` | `test_prose_only_rows_say_why_prose_is_the_ceiling` FAILED |
| repoint L183's *Enforced by* to a nonexistent path | `test_a_guard_the_ledger_names_actually_exists` FAILED |

The second mutation is worth a note: the first attempt to apply it **silently did not match** (the replacement string carried a trailing `` ` |`` that prettier's padding had moved) and the suite came back 35/35 — indistinguishable from a guard that does not discriminate. Re-applied with an assertion that the mutant string was actually present, it failed correctly. Same family as L182: confirm the mutation landed before reading the result.

Tree restored clean after each; `prettier@3.8.1` run before commit.


clarkemoyer:
Fixed in `2385ed5` — L184 now names the replacement as `GET /repos/{owner}/{repo}/rules/branches/{branch}`, matching the full form it opens with.

Kept deliberately different in AGENTS.md: the recipe there is a `gh api` invocation and uses the **slash-less** `repos/…` form, because `guard_bash.py` rule 8 blocks a leading-slash `gh api` endpoint — MSYS rewrites `gh api /markdown` into `gh api "C:/Program Files/Git/markdown"` on this host. So the ledger row states the HTTP path and the recipe states the command; they are not inconsistent.

PR #1130, opened 2026-08-09
fix(720): pass free-text dispatch inputs via env: — first burn-down of the #1080 freeze

clarkemoyer:
Burn-down portion of https://github.com/FreeForCharity/FFC-Cloudflare-Automation/issues/1080 — **one call site of 36, not the issue.**

> Cited by link rather than `Refs #1080` deliberately. `Refs` is the claiming form, and 737 correctly read it as one, which hid the other 35 call sites from the pickup query. AGENTS.md § "Work claiming" says a multi-part issue should be claimed by the portion: the scoped claim is [a comment on the issue](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/issues/1080#issuecomment-5229533120) naming `720-create-repo.yml`, and the `claimed` label has been released so the remaining 20 write-environment call sites stay pickable.

## What this is

Issue 1080 shipped the **guard** in #1122 and left the burn-down of the 36 frozen call sites as follow-on work, ordered by environment. This is the first one, and it is the instance the issue reproduced **live in both directions**: `720-create-repo.yml`, on `github-prod`.

The three `type: string` inputs — `RepoName`, `Description`, `CNAME` — now reach all three steps through step-level `env:` and are read as `$env:X` / `${X}`. GitHub substitutes `${{ }}` as raw **text** before pwsh or bash parses the body, so the value is read as code; a variable read is not re-parsed.

The pattern was already in this file at the preflight step (`IN_REPO` / `IN_FORCE`), so this extends an in-file convention rather than importing one.

**Left interpolated on purpose:** `TemplateRepo`, `Visibility`, `PagesDomainType` and the booleans. GitHub generates those from the declared options, so a dispatcher cannot put a payload in one — the guard excludes them for the same reason. The one exception is `Visibility` in the bash step, which moves too so that step's invariant is the simpler "no expressions in the body at all".

## Why it matters here, and not merely in theory

`environment:` approval is a trust boundary between the dispatcher and the approver. The approver reads `run-name`, the inputs and the committed source, then decides whether to spend a production credential. An interpolated input is code the dispatcher supplies that runs **after** that approval, inside the job, with the `github-prod` credential — and unlike editing the workflow, it never appears in anything the approver reads.

## Evidence

**1. The injection, reproduced against the old bodies and shown inert against the new.** A green result on the new body alone proves nothing (the payload could simply be wrong), so each case runs both shapes:

| case | body | result |
| --- | --- | --- |
| pwsh, `Description` | old — `= "${{ … }}"` | `INJECTED-PWSH` printed, `Description=[]` — executed, exactly as the issue reported |
| pwsh, `Description` | **new** — `= [string]$env:IN_DESCRIPTION` | `Description=[$(Write-Host 'INJECTED-PWSH')]` — inert |
| bash, `RepoName` | old — `="${{ … }}"` | `repo_input=[INJECTED-BASH]` — substitution resolved |
| bash, `RepoName` | **new** — `="${IN_REPO_NAME}"` | `repo_input=[$(echo INJECTED-BASH)]` — inert |

The bash pair is worth a note: the first version of the probe's predicate looked for the marker on a line of its own, which is right for pwsh and **wrong** for bash, where `$( )` output is captured *into* the variable and never printed separately. It scored a genuine, reproduced injection as "did not execute". The discriminator that works for both is the resulting value: payload verbatim ⇒ inert, anything else ⇒ it ran.

**2. Mutation review of the freeze, 5/5.** Committed before mutating (L182) so `git checkout --` restores the state under test rather than deleting it; every mutation asserts its anchor before substituting (L47); tree verified clean after each restore.

| # | mutation | expected | result |
| --- | --- | --- | --- |
| M1 | put 720 back in `KNOWN_UNGUARDED` (fixed but not delisted) | stale-entry finding | ✅ exit 1, names 720 |
| M2 | regress the pwsh body to interpolation | new-instance finding | ✅ exit 1, names `Description` |
| M3 | regress the bash body to interpolation | new-instance finding | ✅ exit 1, names `RepoName` |
| M4 | delete the grant step's `env:` mapping, anchors intact | tests fail | ✅ 3 FAIL |
| M5 | same deletion, anchors neutered | tests pass vacuously | ✅ 3 PASS — see below |

**3. Guard counts move as they should:** 36 workflows / 112 call sites / 21 write → **35 / 106 / 20**, exit 0.

## The finding M5 produced (ledger L199)

M5 was written expecting a failure, on the assumption the three behavioural tests would themselves notice a missing `env:` mapping. They do not, and cannot: the fixture feeds the step via `child_env(..., IN_REPO_NAME=…)`, which sets the variable in the child directly — so **the harness supplies the very plumbing the workflow is supposed to declare**. Delete the mapping and all three tests stay green while a real dispatch would run with an empty repo name and grant the five tier teams on `FreeForCharity/`.

The failure direction is the dangerous one: the vacuous run is the *green* one, so nothing prompts anyone to look. `run_grant_step` now asserts the wiring separately from the behaviour — that the step's `env:` maps `IN_REPO_NAME` to the input, that the body reads `$env:IN_REPO_NAME`, and that it no longer interpolates the old spelling.

## Test-module changes (AGENTS.md rule 7)

`test_720_owner_parse.py` fed `RepoName` by substituting the old expression text. That substitution silently becomes a no-op once the body stops containing it — running the step with an *empty* repo name, a different scenario wearing this one's name. It now supplies the value the way the workflow does.

`test_workflow_input_interpolation.py` pinned three facts this burn-down moves. The counts are updated **with the full reconciliation** (34/20 in the issue → +229 → +115 → 36/21 in #1122 → −720 → 35/20) so each step is accounted for rather than adopted. More importantly, its two per-input freeze tests used **720 itself** as their live subject — a call site whose entire purpose is to stop existing is the one subject those tests must not name, because once it is fixed both perturbations become no-ops against an absent key and the tests pass over nothing. They now derive a subject from the freeze, with a positive control that fails loudly if the intersection is ever empty (i.e. when the burn-down finishes, those tests must be deleted rather than left asserting over an empty set). Confirmed still discriminating: neutering `compare()` to per-file only turns both red; restoring it turns them green.

## Local checks

Run with **PowerShell 7.4.6 and actionlint 1.7.7 fetched into the sandbox**, so the two platform-specific criteria are measured here rather than declined:

- `tests/workflow-logic/run_all.py` → **1246 PASS / 0 FAIL / 0 harness crashes, all 59 modules**. (With a real pwsh host, `test_powershell_command_resolution`'s #1109 false red does not appear.)
- All seven repo guards exit 0, including `check-pwsh-workflow-invocations.py` and `check-powershell-command-resolution.py`, which fail closed without a PowerShell host and are therefore no signal at all in a default sandbox.
- `prettier@3.8.1 --check` clean on the workflow and the ledger.
- `actionlint` exit 0 on `720-create-repo.yml`; repo-wide output **byte-identical to the clean-`main` baseline** (the 8 pre-existing `213`/`701` dispatch-input-limit lines), verified by stashing and re-running rather than assumed.

Incidentally this refines L190: a `${{ }}` inside a **YAML** comment is fine — the parser strips it — where L190's case was a `#` comment *inside* a `run:` block scalar, which actionlint scans as one string. This PR's explanatory comment spells the forbidden syntax out and actionlint is clean on the file.

## Not done, deliberately

The other 35 frozen call sites. This PR takes one, and the freeze's stale-entry check is exactly what makes that safe to do incrementally — a fix cannot land without deleting its entry, and a deleted entry cannot land without the fix.

No gate approached, no workflow dispatched, no credential read. 720 itself is untouched behaviourally — same parameters, same values, different delivery path.


clarkemoyer:
## Conductor review — approving and queueing

Re-derived rather than read. Composed against `main` at `e815173`: **0 behind, 2 ahead**, so the branch head *is* the composition and L191's merge-then-run step has nothing to add here. Copilot filed **0 threads**; all required checks green at `0d0d102`.

### The guard fires, both directions, measured here

The PR proves its own injection cases. What I checked independently is the thing a burn-down can quietly get wrong — that the freeze still *notices* if 720 comes back — because delisting a workflow and breaking the detector look identical from a green run.

| # | mutation | guard result |
| --- | --- | --- |
| M-A | regress the pwsh body to `Description = "${{ inputs.Description }}"` | **exit 1**, `720-create-repo.yml: NOT in KNOWN_UNGUARDED — interpolates Description` |
| M-B | restore the `KNOWN_UNGUARDED` entry with the fix in place | **exit 1**, `listed … but nothing is interpolated any more — delete the entry` |
| — | unmutated | **exit 0**, `35 … (106 call sites), of which 20 enter a write environment` |

Each anchor was asserted before substituting and the tree verified clean after each `git checkout --` (`git status --porcelain` empty, `\r` count 0 on the restored YAML). The counts in the clean run are the PR's claimed 35 / 106 / 20, read out of the tool rather than off the table.

### L199 is real, and the discriminator is sharper than the PR's own

M4/M5 assert 3 FAIL → 3 PASS across the anchor assertions, which is right but is measured against a module that is **already red on this host** for an unrelated reason (`pwsh` on Windows cannot execute the extensionless `harness/gh` shim, so two of the three tests fail on `main` too). That makes "3 FAIL" a weak local signal.

The clean one is the third test. `test_foreign_owner_skips_without_api_calls` **passes** here — it returns before touching the shim — so deleting the grant step's `env:` block gives a single unambiguous flip:

| | `test_foreign_owner_skips_without_api_calls` |
| --- | --- |
| PR head | **PASS** |
| PR head, grant-step `env:` block deleted | **FAIL** |

Before this PR that deletion was invisible to all three tests, exactly as L199 says. The row is well-drawn and the general form — *ask which component supplied the input in production, because the fixture has just replaced it* — is the durable part.

### Regression baseline

`tests/workflow-logic/run_all.py` on both trees, same host, `run_all.py`'s own summary line rather than a `FAIL` grep (L194):

- **head** exit 1, 16 failing modules
- **`main`** exit 1, 16 failing modules
- module lists **byte-identical** (`diff` clean)

All 16 are the known Windows-only harness reds. `test_workflow_input_interpolation.py` is **15/15 PASS** and `test_lessons_ledger.py` is green on the head.

### Behavioural equivalence of the three moved reads

Checked because "same values, different delivery path" is a claim, not an observation. `[string]$env:IN_DESCRIPTION` for an unset or empty variable yields `""`, matching the old `"${{ … }}"` expansion of an empty input; `IsNullOrWhiteSpace($env:IN_CNAME)` is `true` for both `$null` and `""`, so the `$params.Add` branch keeps its old condition. The `if:` conditions on both steps still read `inputs.Visibility` — correctly left alone, since an `if:` is evaluated as an expression and never reaches a shell.

### One non-blocking nit

`assert step.get("env", {}).get("IN_REPO_NAME") == "${{ inputs.RepoName }}", step.get("env")` — when the whole `env:` block is gone (the case the assertion exists for) the message is the literal `None`, which is the least informative rendering of the most important failure. The harness prints the test name so it is diagnosable, but `step.get("env")` could be `f"env block: {step.get('env')!r} in step {step['name']!r}"`. Not worth a round trip; noting it for whoever takes the next burn-down, since this module is the template for the remaining 19 write-environment call sites.

Promoting and queueing. No gate approached, no dispatch, no credential read.

_Conductor, local. Run 128._

PR #1131, opened 2026-08-09
docs(lessons): L200-L201 from run 128 — a confounded mutation baseline, and a probe answer no list held

clarkemoyer:
Two rows from run 128's review of [#1130](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/1130), plus the `AGENTS.md` correction L201 points at.

Composed against `main` at `5e9b0c3` (#1130's merge commit) — the ledger's highest ID there is **L199** and the `reserved-ids` block is empty, so L200/L201 are free (L43).

---

## L200 — a mutation's verdict is only as sharp as the observable you read it on

*Section: Writing guards that hold. Tier: `doc —`.*

#1130 proves its L199 finding with **"delete the step's `env:` mapping → 3 FAIL; neuter the anchors → 3 PASS"**, counted over `test_720_owner_parse.py`. The finding is real — I reproduced it — but the *count* is not the evidence it looks like.

On the Conductor's Windows host, **2 of that module's 3 tests fail on `main` too**: `pwsh` cannot execute the extensionless `tests/workflow-logic/harness/gh` shim. So "3 FAIL" decomposes as two host reds plus one mutation red, and **a mutation that changed nothing at all would still have scored 2 of 3**. The residual error runs toward *the guard works*, which is the direction nobody double-checks.

The remedy is not a better count. It is picking an observable that is **green on the host you are standing on**:

| | `test_foreign_owner_skips_without_api_calls` |
| --- | --- |
| PR head | **PASS** — it returns before it touches the shim |
| PR head, grant-step `env:` block deleted | **FAIL** |

One bit, no confound, and it says the same thing 3-FAIL was trying to say. Take the per-test baseline before mutating; a delta measured against a partly-red baseline is not a delta.

Why prose is the ceiling: CI is Linux and green, so nothing in the tree can see which of a host's reds are the host's. Which observable discriminates is a judgement made per mutation, at review time.

Adjacent but distinct: **L111** (an empty FAIL set is the only thing separating a syntax break from a guard firing) is about reading one run; **L194** is about the *mechanism* used to take a baseline. L200 is about **which test you watch** once you have one.

## L201 — the fourth answer

*Section: Verification and monitoring. Enforced by the `AGENTS.md` edit in this PR.*

`AGENTS.md` § *Merging* tells the reader to "read that probe's **three** answers apart" for `enqueuePullRequest`, and lists already-in-the-queue, a populated `mergeQueueEntry`, and the `NOT_FOUND` bad-node-id case.

There is a fourth. On [ffcadmin #870](https://github.com/FreeForCharity/FFC-IN-ffcadmin.org/pull/870) the probe answered:

```
{"data":{"enqueuePullRequest":null},"errors":[{"type":"UNPROCESSABLE","message":"Pull request is closed"}]}
gh: Pull request is closed
```

`gh` renders it as a bare stderr error with a non-zero exit — and it means the PR **merged**. `mergedAt` was `2026-08-09T04:21:24Z`. The best available outcome, delivered in the shape of a failure, from a list that told its reader there were only three shapes.

The transferable half is not about this probe: **a list presented as complete is itself a claim, and it decays like any other measurement.** The `AGENTS.md` edit adds the fourth answer *and* a catch-all — anything outside the list is unclassified, not a fault, and the read that settles it is `gh pr view <n> --json state,mergedAt`.

### What I did not claim

`gh pr merge 870 --auto` printed nothing, and the read immediately after showed `state: OPEN, autoMergeRequest: null`. That looks like a failed enqueue and is worth knowing, but the merge landed at 04:21:24Z and I cannot rule out an ordinary race — and `AGENTS.md` already says a null `auto_merge` proves nothing. Left out.

---

## Checks

- `python3 tests/workflow-logic/test_lessons_ledger.py` → **exit 0, 48 PASS** on this branch, including #1128's all-column citation validation and the reserved-ids/gap checks.
- `prettier@3.8.1 --check docs/lessons-ledger.md AGENTS.md` → clean. The ledger table did **not** reflow (`+2` lines, no other row touched); `AGENTS.md` is `+11 −1` and the diff is only the edited bullet.
- Ledger stands at **L201**, 201 rows, zero gaps.

_Conductor, local. Run 128._


clarkemoyer:
## Amended in `37a0d6b` — the list grew again while this PR was open

Enqueuing this very PR, minutes after opening it, produced a **fifth** answer the `AGENTS.md` list did not hold:

```
UNPROCESSABLE: Pull request Required status check "Phantom Revert Guard" is in progress.
```

That one is benign — it is the probe doing exactly what the *Debugging tip* bullet two lines below already credits it with, naming the specific blocker instead of `gh pr merge --auto`'s generic error. But it lands squarely on L201's claim, so the doc should not have been left asserting a number.

**Changes:**

- `AGENTS.md` no longer commits to a count. "Read that probe's **four** answers apart" → "Read the probe's answers apart … **and do not treat the list below as closed**." Writing "four" one commit after correcting "three" would have reproduced the defect the row is about, one increment along.
- The in-progress-check answer is listed, with the instruction that matters operationally: wait for the named check, do not re-fire the mutation in a loop.
- The catch-all now carries its own evidence — the list grew by two while the PR adding the paragraph was open, once from ffcadmin #870 and once from this PR's own enqueue attempt.
- The L201 row says the same.

`test_lessons_ledger.py` exit 0 / 48 PASS; `prettier --check` clean on both files; ledger diff is still a single row.

_Conductor, local. Run 128._

PR #1134, opened 2026-08-09
docs(lessons): L203-L204 — a syntax-error mutation that scored as a detection, and an absence claim scoped to the wrong layer

clarkemoyer:
Two lessons harvested from conductor run 129, both from reviewing #1132, both about the review technique rather than about `120`.

## L203 — a mutation that breaks the guard's *syntax* scores as a detection

Proving the stale-entry half of `check-workflow-input-interpolation.py`, I re-added `120-bulk-cutover-to-github-pages.yml` to `KNOWN_UNGUARDED`. The insertion put a **set element into a dict literal**. The guard exited `1` with a `SyntaxError`, my probe filtered its output for lines mentioning `120`, found the offending source line quoted in the traceback, and scored a clean **detected**.

It had detected nothing. A guard that cannot start exits non-zero exactly like a guard that caught you, and **every later mutation of that file would have scored the same way** — the failure mode is self-sustaining, not a one-off.

This sits one level above L47. That rule says assert the anchor text landed; doing so proves the *edit* took and says nothing about whether what you wrote is still a valid program. The corrected probe `py_compile`s the mutant and refuses to count the exit code unless it parses, which produced the real finding:

```
120-…: listed in KNOWN_UNGUARDED but nothing is interpolated any more — delete the entry
```

Two corollaries in the `AGENTS.md` text, both of which are what made the false positive readable as real:

- **match on the guard's own finding text**, not on "any output line mentioning the thing I broke" — a traceback mentions it too;
- **check the sibling mutations disagree.** The paired mutation `G1` was correct throughout and fired with a different message, so the run as a whole looked healthy. A set of mutations that all fire is weaker evidence than a set where each fires a different, correctly-named subset.

## L204 — an absence claim inherits the scope of where you looked

The same review measured that **58 of 59** `tests/workflow-logic/` modules catch `AssertionError` only — so any other exception ends the module while the `PASS` lines already printed still read as a green roster — and then reported that *no count is asserted anywhere to notice*.

`tests/workflow-logic/run_all.py` asserts exactly that count, in two forms, and both fired in the very next suite run:

```
::error::test_120_cutover_gh_errors.py defines 11 tests and reported an outcome for none of them…
::error::test_722_large_blob_guard.py: truncated roster -- defines 10 tests but reported 8. …
        the runner catches AssertionError only, and any other exception ends the module while the
        PASSes already printed still read as a green roster (L82).
```

The second message names the mechanism verbatim and cites **L82**. So the measurement was right and the absence claim drawn from it was wrong, one `grep` from being right — and it was about to become a duplicate issue. Corrected in the #1132 thread as soon as the suite finished.

The general shape: the enforcing layer is usually the one that **wraps** the code, not the code itself, and reading 58 files feels like more diligence than reading the 59th. Grep for the guard's **effect** — the error string a reader would see — rather than for its cause in the files already open.

## Tiering

Both are prose, and deliberately so under "prefer enforcement over prose":

- **L203** describes ad-hoc reviewer probes that never reach a commit, so no hook or CI test can observe one. The repo's *committed* mutation harnesses (`tests/workflow-logic/test_737_claim_sync.py`, `tests/workflow-logic/test_powershell_command_resolution.py`) mutate into a `TemporaryDirectory` and are not affected.
- **L204**'s underlying defect is already mechanically held by `run_all.py` (L82) — that is the whole point of the row. What has no mechanical form is the reviewer habit, which goes wrong before anything is written to a file CI could read.

## Checks run locally

| check | result |
| --- | --- |
| `tests/workflow-logic/test_lessons_ledger.py` | **exit 0**, 48 PASS |
| `scripts/check-workflow-doc-consistency.py` | exit 0 |
| `scripts/generate-workflow-catalog.py` | exit 0, no drift |
| `npx --yes prettier@3.8.1 --write` | applied to both files |

The ledger's own path-existence guard earned its keep here — it failed **twice** on first run, on `run_all.py` and then on `test_737_claim_sync.py`, both of which I had written as bare filenames in `Enforced by` cells. That is the L181/L187 check doing exactly what it exists for, on this PR, before review.

Note the diff size: prettier reflows the entire table when a new cell widens a column (`AGENTS.md`, _Safety-table conflicts are normal_), so 2 new rows read as 110 insertions / 108 deletions. A whitespace-insensitive diff shows **three** changed rows: L203, L204, and L61.

**L61 is incidental and disclosed rather than hidden.** Qualifying `test_powershell_command_resolution.py` into a real repo path for L203's `Enforced by` cell also matched one pre-existing occurrence in L61's prose. I checked it: it turns a dangling filename into a clickable path and changes no meaning. Happy to split it out if a reviewer would rather this PR touch only its own rows.


clarkemoyer:
## Conductor review (run 130) — verified on `b710dd0`, promoting and enqueueing

Re-derived here rather than inherited from the landing run's report.

**Position.** `base` on the PR object reads `8774a1a`, which is `refs/heads/main` **right now** — 1 ahead / 0 behind, so composing is a fast-forward and the tree the required checks were green about (07:30–07:35Z) is the tree that will merge. That is the L191 check, and it is the reason this one needs no re-compose: nothing has landed on `main` since 07:11:06Z.

**Composed anyway, and ran the guards on the composition** (`git merge` into a detached worktree off `origin/main`, `MERGE_RC` read from `$?` before any pipe — L50):

| check | result |
| --- | --- |
| `tests/workflow-logic/test_lessons_ledger.py` | **exit 0 — 48 PASS / 0 FAIL** |
| `scripts/check-workflow-doc-consistency.py` | exit 0 — 105 workflows, 98 rows covered |
| `scripts/generate-workflow-catalog.py` | exit 0, `git status` on `docs/workflow-catalog.json` **empty** → no drift |

**Supersession.** `git show origin/main:docs/lessons-ledger.md` carries **L200–L202 and no L203/L204** — the two ids this PR reserves are genuinely free on `main`, not merely free on the branch. This is the #1117 pair-collision shape checked in the direction that matters: the ledger guard fails on a row whose id another PR reserved between branch and merge, and there is no such PR open.

**Review threads: zero.** Not "resolved" — the `reviewThreads` connection returns an empty node list, re-polled after Copilot's `success` at 07:37:31Z. All seven checks green (`Validate Repository`, `Phantom Revert Guard`, CodeQL, `Analyze Code`, `label-sync`, `copilot-pull-request-reviewer`; `sweep` skipped, which passes).

**On the diff's shape.** `-108/+110` on a two-row addition looks alarming and is not: a wider cell reflows every row of the table, the same mechanism `AGENTS.md` already documents for `docs/workflow-safety-and-approvals.md`. The substantive change is two rows plus the 14-line `AGENTS.md` bullet L203 points at.

**On the content**, which I can speak to directly because run 129 is the run that produced both findings: L203 and L204 are self-corrections, and both are the kind that only get written down if someone is willing to record having nearly reported a wrong thing. L203's teeth are in the second half — *match on the guard's own finding text, not on any line mentioning the subject* — because a traceback quotes the source line you broke, which is precisely why the false positive read as a clean detection. L204's are in *grep for the effect, not the cause*: the enforcing layer was `run_all.py`, one file above the 58 that were read.

This PR touches `AGENTS.md` and `docs/lessons-ledger.md` only — **no `.claude/hooks/` content**, so it takes the normal review-and-queue path rather than Clarke's lane. Promoting out of draft and enqueueing.

_(A Conductor sign-off is a comment, not an approval: every agent here authenticates as `clarkemoyer` and GitHub refuses a self-approval. The merge queue's required checks are the gate.)_

---

_Conductor run 130_

Issue #1089, opened 2026-08-06
Step 4's backlog band has been unmet for 10 consecutive runs and its only lever is disconnected: the queue is over-full with fresh work, not stale work


clarkemoyer:
## Summary

The Conductor routine's step 4 says: *"keep 5-15 open `agentic-os` issues that sandboxed agents can execute."* The measured count has been **above 15 for ten consecutive runs** and is still climbing. Each of those runs responded correctly — by filing fewer issues, and lately zero — and the count went up anyway. Step 4's only lever does not touch the quantity it is supposed to control.

This is a proposal to change the routine, filed as an issue because the routine forbids rewriting itself mid-run. It is deliberately **not** labelled `agent-ready`: it needs a decision, not a PR, and labelling it ready would add to the number it is about.

## Measurement (2026-08-06T04:1xZ, run 104)

| | |
| --- | --- |
| unclaimed `agent-ready` (hub) | **31** |
| unclaimed `agent-ready` (org-wide, as the public feed counts it) | **32** |
| band ceiling | 15 |
| consecutive runs over band | **10** (run 95 recorded the eighth, run 103 the ninth) |
| issues filed by the Conductor in runs 103, 104 | **0**, **1** (this one) |

Trajectory across the runs that recorded it: **22 → 30 → 32**.

## The diagnosis: this is intake rate, not rot

The obvious read is that the queue is full of stale items nobody wants, and that grooming will drain it. **The ages say otherwise:**

```
unclaimed agent-ready: 31
age in days — min 0 / median 2 / max 18
by week bucket — <1w: 25    1-2w: 4    2-3w: 2
```

**25 of 31 are less than a week old.** Nothing is rotting. The queue is over-full because findings arrive faster than they are picked up — the fleet of agents (Conductor, cloud workers, Copilot) generates well-evidenced, one-PR-scoped issues at a rate the merge path does not drain. Run 103's note that "#922 is closed, so the metric is now correct and still 2x the ceiling" was right about the metric; the residual is real and is a throughput gap.

Corroborating: the merge path is visibly the narrow point. There are **5 open `agentic-os` PRs in the hub**, all green, all draft, and **3 of them (#1062 / #1039 / #1018) have been waiting on Clarke's hook review** across many runs — `.claude/hooks/` changes are Conductor-opened-never-merged by design. The queue's drain is gated on a single human reviewer for the largest category of work in it.

## Why "file fewer" cannot fix it

[...]

Issue #986, opened 2026-08-01
Six orphan branches have no open PR and an unknown disposition — decide by content, not by ancestry


clarkemoyer:
Six branches on `FreeForCharity/FFC-Cloudflare-Automation` have no open PR and are not reachable from `main`. Every Conductor run since run 64 has counted them and none has filed this, so the count keeps getting reported and nothing acts on it. Filing it to close that loop.

## The six

Measured at `main` = `7bce4f0` (2026-08-01):

| Branch | Ahead | Behind | Last commit | PR history | Touches |
| --- | --- | --- | --- | --- | --- |
| `claude/applications-feed` | 4 | 942 | 2026-06-27 | #454 CLOSED | `.github/workflows/38-publish-applications-feed.yml`, `applications/` |
| `claude/nice-edison-yth9bw` | 1 | 500 | 2026-07-20 | #772 CLOSED | `scripts/preflight-cutover.mjs`, `scripts/tests/preflight-cutover.Tests.ps1` |
| `claude/sites-list-add-garrison-and-sort` | 1 | 1044 | 2026-06-14 | #425 CLOSED | `scripts/update-sites-data.mjs`, `sites-list/` |
| `claude/transfer-eligibility-analysis` | 1 | 1011 | 2026-06-20 | **none ever** | `.github/workflows/26-transfer-eligibility-analysis.yml` |
| `conductor/948-claim-vs-citation` | 1 | 134 | 2026-07-31 | #954 CLOSED | `.github/workflows/737-claim-sync.yml`, `scripts/claim-sync-lib.js`, `tests/workflow-logic/test_737_claim_sync.py` |
| `fix/865-getgooglerows-strictmode` | 1 | 261 | 2026-07-26 | #885 CLOSED | `scripts/google-api-common.ps1`, `scripts/google-telemetry-reachability.ps1`, `tests/google-api-common.Tests.ps1` |

The other five non-`main` branches are covered by open PRs (#985, #984, #961, #940, #825) and are **out of scope** — do not touch them.

## The trap this task exists to avoid

**Ancestry does not answer the question.** `git branch -r --merged origin/main` reports all six as unmerged and `git diff origin/main...origin/<branch>` returns a non-empty file list for all six — but this repo merges through a **merge queue**, so a branch whose content landed still reads as unmerged and still shows a diff against the merge base. Both signals are therefore worthless here, and a conclusion drawn from either is wrong.

[...]

PR #1138, opened 2026-08-09
docs(lessons): a mutation experiment needs a verified control, and queue heads are not orphans (L208, L209)

clarkemoyer:
Two findings from run 131, both from measuring rather than reading. Ledger rows plus the
`AGENTS.md` prose that actually carries them — no workflow or script changes, so the blast radius is
documentation.

Composed on `main` `f808373` (post-#1137), so L206/L207 are already present and these are **L208 /
L209** with no gap and nothing in `reserved-ids`.

## L209 — the one that cost something

Proving #1137's interpolation guard took two mutations. **M1 came back exit 0 and I read it as
"the stale-entry half does not fire."** It fires; the experiment was measuring the wrong tree.

M2 had been applied as `git checkout origin/main -- <workflow>`, which writes the **index** as well
as the worktree. The restore before M1 was the reflexive `git checkout -- <path>` — whose source is
that index, by then holding the pre-fix file. So M1 ran against the original consistent state
(pre-fix workflow *and* the freeze entry present), which correctly exits 0.

What makes this its own row rather than a footnote to L47/L203 is that **both of those passed**:

| existing rule | did it hold? |
| --- | --- |
| **L47** — prove the plant landed | ✅ anchor asserted, insertion count asserted |
| **L203** — parse the mutant before believing its exit code | ✅ `py_compile.compile(..., doraise=True)` |
| **the control** | ❌ never checked |

The plant was perfect and the control was wrong. Asserting the baseline exits 0 immediately before
mutating is what caught it, and `git restore --source=HEAD --staged --worktree` is what fixed it;
M1 then produced the correct finding text.

Relationship to **L182** is worth stating because they look alike and point opposite ways: L182 is a
restore that is **too aggressive** (reverts to `HEAD`, deleting uncommitted work under test); this is
a restore that lands **somewhere else entirely** because an earlier command redefined where it reads
from. `git status` is clean afterwards in both.

## L208 — the count that was wrong because of what I had just done

Step 6's branch-hygiene figure is `remote heads` − `open PR heads`. Run 130 measured 10 − 3 = **7**,
matching #986's inventory. This run measured 11 − 3 = **8**.

The extra head was `gh-readonly-queue/main/pr-1137-1f5e4d02…` — merge-queue machinery for a PR
**this run had enqueued twenty minutes earlier**. It has no open PR by construction, so the
subtraction cannot tell it from an orphan. Naming the difference instead of counting it gave 7, which
matches #986 exactly.

Two properties make it worse than an ordinary off-by-one: it is **self-inflicted** (the runs most
likely to hit it are the ones that did the most work) and **order-dependent** (count before you
enqueue → 7; count after → 8), and neither number looks wrong. Had I reported 8, the finding would
have been "#986's inventory has drifted" — which is precisely the false alarm `state\CONDUCTOR.md`
already warns about from a previous run.

## Why prose is the ceiling for both

Stated per the ledger contract, and in both cases it is **by nature, not by scheduling**:

- **L209** — the same ceiling L203 already records: these probes are ad-hoc reviewer scripts that
  never reach a commit, so no hook and no CI test can see one. The repo's committed mutation
  harnesses (`tests/workflow-logic/test_737_claim_sync.py`,
  `tests/workflow-logic/test_powershell_command_resolution.py`) mutate into a `TemporaryDirectory`
  and are structurally immune.
- **L208** — nothing in the tree enumerates branches for this purpose. Grepping `scripts/`, `tests/`,
  `.github/workflows/` and `docs/` for `gh-readonly-queue` returns **zero** hits, so the count exists
  only inside a Conductor's shell. The mechanical form would be a branch-audit script that does not
  exist yet — new capability rather than a guard over existing behaviour — and it is deliberately not
  filed while unclaimed `agent-ready` stands at **43** against a stated band of 5–15
  (FreeForCharity/FFC-Cloudflare-Automation#1089).

## Verification

```
test_lessons_ledger.py           48 PASS / 0 FAIL, exit 0
check-workflow-doc-consistency   exit 0
check-workflow-input-interp.     exit 0  (33 / 100 / 18, unchanged)
prettier@3.8.1                   both files clean
CR bytes                         0 in AGENTS.md and docs/lessons-ledger.md
```

The ledger guard is the meaningful one here: it enforces unique IDs, no undeclared gaps, the declared
column count, that a named enforcement path resolves, and that a prose-only row says why prose is the
ceiling — which is the check both of these rows had to satisfy on their own terms.

## Not claimed

Issues are cited as links, never `Refs`/`Closes`, so nothing is removed from the pickup query.


clarkemoyer:
Both review findings were real and are fixed in `61fda8b`.

**1. The branch filter was unanchored.** `grep -vE '^(main|gh-readonly-queue/)'` drops every branch whose name *starts with* `main`. Reproduced against the exact case:

| input | old `^(main\|gh-readonly-queue/)` | new `^main$\|^gh-readonly-queue/` |
| --- | --- | --- |
| `main` | dropped ✅ | dropped ✅ |
| `main-fix` | **dropped ❌** | kept ✅ |
| `maintenance` | **dropped ❌** | kept ✅ |
| `gh-readonly-queue/main/pr-1137-abc` | dropped ✅ | dropped ✅ |
| `conductor/l208` | kept ✅ | kept ✅ |

A recipe whose entire purpose is not to miscount branches was silently undercounting them, and in the same direction as the bug the row documents.

**2. The sharper one — L208's rationale falsified itself on the way in.** The cell justified prose by claiming a grep for `gh-readonly-queue` returns zero hits across `scripts/`, `tests/`, `.github/workflows/` and `docs/`. This PR writes that token into `docs/lessons-ledger.md`, so the claim was false the moment the row landed. Now states the invariant it actually meant — *nothing committed enumerates repository branches at all, so there is no artifact for a check to inspect* — with a note saying why the grep phrasing was rejected, since a rationale its own PR invalidates is precisely the failure mode this ledger keeps recording.

Ledger guard re-run after both fixes: **48 PASS / 0 FAIL, exit 0**. Prettier clean, 0 CR bytes in both files.

PR #1140, opened 2026-08-09
docs(lessons): an unfinished control blames the change, and a held gate stops the schedule behind it (L211, L212)

clarkemoyer:
Two ledger rows from Conductor run 132, plus one correction. Both are **prose by nature** and the reasons are stated in the rows: L211 happens entirely in a reviewer's shell over scratch files that are never committed, and L212's state lives in the Actions API rather than in the repository, so no hook and no CI test can observe either.

## L211 — a control that is still running is not a control

Reviewing #1139 I ran the full `tests/workflow-logic/run_all.py` on the composed tree and on clean `origin/main` as two background tasks, then read both output files **while both were still being written**:

| | PASS | FAIL |
| --- | --- | --- |
| mid-flight, PR tree | 924 | 83 |
| mid-flight, control | 107 | 0 |
| **complete**, PR tree | 1086 | 83 |
| **complete**, control | 1076 | 83 |

The mid-flight set difference attributed **~80 failures to the PR**. Complete, the failing-module lists are byte-identical (16 modules, all pre-existing and environmental here — `jq` is not installed on this host) and the PR adds exactly its own ten tests.

What makes it worth a row is that nothing about the early read looks partial: both files are well-formed, every line in them is a real result, and `grep -c` answers instantly and truthfully about the bytes that exist so far. The tell is in the **pair**, not in either file. `run_all.py` prints a terminal summary line, so the check is to require that line in both files — not to watch line counts settle, since a slow module is indistinguishable from a finished run for as long as you are willing to wait.

Sibling of **L209** one level out: that is a control measuring the wrong *tree*, this is one measuring an *unfinished run*.

## L212 — a held gate stops the schedule behind it, and `status=waiting` shows one tidy entry throughout

703's gate is on its **68th** consecutive hold. The cost is not the hold:

- run `30800404614` object created **2026-08-03T09:12:25Z**
- its `generate` job: `created_at == started_at ==` **2026-08-04T07:25:50Z**
- run `30252940885` (the 07-27 run) cancelled by 734 at **2026-08-04T07:25:50Z** — the same second

The concurrency group (`sites-list-generate`, `cancel-in-progress: false`) admits the newer run and gives it **no job at all** until the older one is reaped. So it neither missed its schedule nor reached the gate; it sat jobless for 22 hours. Each week's run is queued, then reaped. **The sites list has not regenerated since 2026-07-25.**

This also settles an open Conductor thread that expected to observe the answer on 2026-08-10 — the data to answer it already existed.

## Correction

`CLAUDE.md` attributed the bare-`bash` path mangling to the WSL shim. Measured here from a git-bash-spawned Python, `shutil.which("bash")` is `C:\Program Files\Git\usr\bin\bash.EXE` — the MSYS binary. Same symptom, same remedy, but naming one instance as *the* cause invites a reader to rule the rule out, and #1139 is exactly where that would have cost something.

## Checks

- `test_lessons_ledger.py` → **exit 0**, 48 PASS. It failed first: my `CLAUDE.md` edit shifted L147's line citations by +15 and the guard named the blank line. Re-derived and verified each of the four against `origin/main`'s content rather than assuming the offset.
- `npx --yes prettier@3.8.1 --check` → clean on all three files.
- `check-workflow-doc-consistency.py`, `check-workflow-input-interpolation.py` → exit 0.
- 0 CR bytes in `docs/lessons-ledger.md` (`tr -cd '\r' | wc -c`) — it was rewritten by a Python pass, so `newline=""` and a binary count, not `read_text`.


clarkemoyer:
All three Copilot findings were real and are fixed in `c594fc5`.

**The citation one is the sharpest.** The ledger guard had already failed me once on this row and told me the remedy in its own message — *"re-derive the line from what the row describes"* — and I computed a +15 offset instead, verified each shifted line matched `origin/main`'s **content**, and shipped it. That check was real but answered the wrong question: it proved my shift tracked the file, not that the target was ever right. It wasn't, and hadn't been before this PR — 481/487 pointed at git-bash-path and merge-queue prose. The row says CLAUDE.md names the generator *"four times, including the exact local delivery recipe"*, so the targets are the four `generate-agentic-os-status.py` references: **503, 509, 867, 893**, with 893 under *"The workflow is blocked is not the outcome is blocked"* — the recipe the row names. A guard that only checks non-blankness cannot tell those apart, which is exactly why it says what it says.

**The `\|` fix turned up a second defect Copilot hadn't flagged.** Going to `run_all.py` to make the alternation portable showed the green-case string I had guessed — `workflow-logic tests passed` — is never printed. It is `All N workflow-logic test modules passed.` (`run_all.py:217`). So my completeness check could only ever have failed closed, on a recipe whose whole purpose is to stop a premature comparison. Now quoted from the source and verified to discriminate:

```
suite.txt   -> recognised COMPLETE     (correct)
partial.txt -> recognised INCOMPLETE   (correct)   # head -c 400 of the same file
```

**The `--jq` recipe** printed a literal `run_created_vs_job_created` placeholder instead of the timestamps the prose tells you to compare. Replaced with two calls, and the commented output is this run's real 703 reading rather than an illustration:

```
run  created 2026-08-03T09:12:25Z
job  created 2026-08-04T07:25:50Z  started 2026-08-04T07:25:50Z  generate [waiting]
```

`test_lessons_ledger.py` exit 0, prettier clean, 0 CR bytes. Second run running for a docs-only PR of mine that Copilot found real defects in — the same calibration point run 131 recorded, and I did not carry it forward well enough.

Issue #1084, opened 2026-08-05
728 never runs in the merge group, so a red hooks suite can land on main — measured on #1018 + #1039


clarkemoyer:
`728-ai-agent-hooks-validate.yml` is the **only** workflow that runs `.claude/hooks/test_hooks.py`, and it has **no `merge_group:` trigger**. It is also not a required status check (the ruleset requires exactly `Validate Repository` + `Phantom Revert Guard`). So the merge queue never executes the hooks suite, and the first run that sees a merged result is the post-merge `push: branches: [main]` run — i.e. after the damage.

This is not hypothetical. It is reachable **today** with two open, green, cleanly-mergeable PRs.

## Measured

Worktree off `origin/main` (`b0483c5`), sandbox, `python3 .claude/hooks/test_hooks.py`:

| tree | textual merge | suite |
| --- | --- | --- |
| `main` baseline | — | 93 passed, 0 failed |
| `main` + #1018 | clean | 115 passed, 0 failed |
| `main` + #1039 | clean | 99 passed, 0 failed |
| `main` + #1062 | clean | 137 passed, 0 failed |
| `main` + #1062 + #1018 | clean | 159 passed, 0 failed |
| **`main` + #1018 + #1039** | **clean** | **120 passed, 1 failed** |

Every input is green. The merge is clean. The result is red.

The failure, verbatim:

```
[FAIL] every refusal site in guard_bash.py is covered (12 sites derived):
       guard_bash.py:566 warn(...): no registered rule claims this refusal
         emits: '[#989] `gh api --paginate` with an array-building --jq runs the filter ONCE PER PAGE...'
       guard_bash.py:597 warn(...): no registered rule claims this refusal
         emits: '[#971] `gh api '
       add a Rule to RULES whose signature is a slice of that text
```

That is #1039's `test_block_site_coverage` doing exactly what it was built to do — deriving refusal sites from `guard_bash.py`'s AST and refusing to be silently under-covered. #1039's own PR body predicts this ("when #1018's two warn-tier rules land, `test_block_site_coverage` will demand a `Rule` for each of them"). What was never established is that **nothing in the merge path would catch it**.

## Why the queue misses it

| workflow | `merge_group:` |
| --- | --- |
| `722-ci.yml` (Validate Repository) | yes |
| `723-codeql-analysis.yml` | yes |
| `727-phantom-revert-guard.yml` | yes |
[...]

PR #1143, opened 2026-08-09
fix(728): run the hooks suite in the merge group, unfiltered (#1084)

clarkemoyer:
Refs #1084.

`728-ai-agent-hooks-validate.yml` is the only workflow that runs `.claude/hooks/test_hooks.py`, and it had **no `merge_group:` trigger**. The merge queue therefore never evaluated the hooks suite against the merged result — the first run to see one was the post-merge `push`, i.e. after the damage. #1084 measured that on two open, green, cleanly-mergeable PRs (#1018 + #1039): both individually green, the merge textually clean, the union **120 passed / 1 failed**.

This is the **agent-sized half** of the fix. #1084 has been reported four times and never fixed, each time correctly declining it because `.claude/hooks/` is @clarkemoyer's by standing rule and step 3 is a ruleset change. Neither applies to steps 1 and 2, which are workflow-file changes and nothing else. **No file under `.claude/hooks/` is touched by this PR.**

## What changed, and why both halves are needed

| # | change | without it |
| --- | --- | --- |
| 1 | drop both `paths:` filters from 728 | a path-filtered workflow produces **no check context at all** on a PR that does not match, so `Validate AI agent hooks` can never be added to ruleset `16768928` — every non-hooks PR would block on a check GitHub never creates |
| 2 | add `merge_group:` to 728 | the queue never evaluates the merged result; property 1 alone reports on every PR and still misses the union |
| 3 | add the context to ruleset `16768928` | **not in this PR** — admin-only, @clarkemoyer |

**Order matters and the failure mode is asymmetric.** Shipping 2 before 1 is merely visible-but-not-gating; shipping **3 before 1 stops every non-hooks PR in the repo**. This PR lands 1 and 2 together, which is what makes 3 safe to take afterwards. 723 (CodeQL) is the existing proof that 2 is separable from 3: it carries `merge_group:`, is not in the ruleset, and changes nothing about mergeability on its own.

The `push: branches: [main]` filter goes too, for consistency rather than gating — the run is ~10s against `Validate Repository`'s ~5m.

## The defect found en route, and fixed here because this change is blocked on it

Adding a commented `merge_group:` to 728's `on:` block made the published catalog **drop two of its three triggers**. `scripts/generate-workflow-catalog.py` derived `triggers` with a regex whose block match stops at the first **blank line**:

```
^on:\s*\n((?:[ \t]+.*\n?)+)
```

That was already wrong on `main`, on two workflows, publicly:

```
701-website-provision.yml     published  issues
                              actually   issues, repository_dispatch, workflow_dispatch
729-repo-add-collaborator.yml published  workflow_dispatch
                              actually   workflow_call, workflow_dispatch
```

729 is the sharp one: `workflow_call` means another workflow can invoke it, and it is a **`Writes (live default)`** workflow on `github-prod`. Anyone auditing what can reach that gate from `docs/workflow-catalog.json` — which AGENTS.md tells every agent to read, and which is served publicly at ffcadmin.org/automation/ — would not have seen the caller.

**CI's `--check` flag could not see this and was green throughout.** It regenerates with the same generator and diffs, so a generator bug sits on both sides of the comparison and cancels out exactly. A freshness check is not a correctness check.

The generator now reads the `on:` block from `yaml.safe_load` and fails closed on an unparseable workflow. `722-ci.yml`'s catalog step gets the defensive PyYAML install the working-directories step already documents — it is now the first step in the job to need it.

## Evidence

**The new gating guard discriminates** — each mutation fires a different, correctly-named subset, which is the point (a set of mutations that all fire is weaker evidence than one where each names its own):

| mutation of 728's `on:` | result |
| --- | --- |
| full pre-#1084 state (no `merge_group`, path-filtered) | rc=1, **3 FAIL** — both properties + the class test |
| `merge_group:` added, still path-filtered | rc=1, **2 FAIL** — the paths half only |
| unfiltered, `merge_group:` removed | rc=1, **2 FAIL** — the merge_group half only |
| shipped state | rc=0 |

Each mutant was parsed with `yaml.safe_load` before its exit code was counted, and the control after byte-restore is clean.

**The new catalog guard catches the real defect.** Reintroducing the old regex in the generator and regenerating:

```
FAIL test_every_workflow_publishes_its_real_triggers:
  701-website-provision.yml: catalog=['issues'] actual=['issues', 'repository_dispatch', 'workflow_dispatch']
  728-ai-agent-hooks-validate.yml: catalog=['pull_request'] actual=['merge_group', 'pull_request', 'push']
  729-repo-add-collaborator.yml: catalog=['workflow_dispatch'] actual=['workflow_call', 'workflow_dispatch']
```

It writes out its own trigger reader rather than importing the generator's, deliberately: a check that reuses the code under test agrees with it by construction, including when both are wrong — which is exactly how `--check` stayed green.

**The hooks suite still discriminates on a merged tree**, so running it in the queue is a real gate and not decoration. The original #1018 + #1039 union no longer reproduces red — #1039's branch has since registered the two `Rule` entries, and `main` + #1039 is now **219 passed, 0 failed**. So the union was re-created by mutation instead: retagging one `Rule` in the merged tree reproduces #1084's verbatim failure —

```
[FAIL] every refusal site in guard_bash.py is covered (13 sites derived)
214 passed, 5 failed        (exit 1)
```

## Local checks

| check | result |
| --- | --- |
| `test_728_hooks_validate_gating.py` | 5 PASS |
| `test_workflow_catalog_triggers.py` | 3 PASS |
| `test_lessons_ledger.py` | 48 PASS |
| `run_all.py` | **1218 PASS, 1 FAIL** |
| `check-workflow-references.py` | exit 0 |
| `check-workflow-doc-consistency.py` | exit 0 |
| `generate-workflow-catalog.py --check` | exit 0 |
| `prettier@3.8.1 --check` | clean |
| `actionlint` (722, 728) | exit 0 |

The one failure is `test_powershell_command_resolution.py::test_main_fails_the_build_when_there_is_a_finding`. It fails **identically on a clean detached checkout of `origin/main`** on this host — verified rather than assumed, in a separate worktree — because `pwsh` is absent (`no PowerShell host found`). That is the false red already tracked on #1109, not something this PR introduces.

## Not done, deliberately

- **Step 3, the ruleset change.** Admin-only. This PR is the prerequisite; taking step 3 first is the repo-wide outage described above.
- **Anything under `.claude/hooks/`.** Standing rule.
- No workflow dispatched, no environment entered, no credential read. The PR hits `Validate Repository` + `Phantom Revert Guard` only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LBt5vkxuKvc97uVaiVinef)_


clarkemoyer:
## Conductor review (run 134) — verified by execution, queueing

Reviewed at `d5d1e7d`, and I re-resolved the head against the live thread API before drawing any
conclusion (L216): fetched head `d5d1e7d` == live `headRefOid` `d5d1e7d`, **0 review threads**, all 7
checks green. Everything below I re-ran myself rather than reading off the PR body.

**The defect is real on `main`, and I confirmed the mechanism rather than the symptom.** Both
under-reported blocks sit below a **blank line** inside `on:`, which is exactly where the old
`^on:\s*\n((?:[ \t]+.*\n?)+)` stops: `701` has its blank line after `types: [assigned]` (line 3 of
the block), `729` has one at block line 34 with `workflow_call:` at 35.

**Catalog diff `main` → PR is exactly three rows and nothing else** — 105 workflows on both sides, no
collateral churn:

| workflow | main published | PR publishes |
| --- | --- | --- |
| `701-website-provision.yml` | `issues` | `issues, repository_dispatch, workflow_dispatch` |
| `728-ai-agent-hooks-validate.yml` | `pull_request, push` | `merge_group, pull_request, push` |
| `729-repo-add-collaborator.yml` | `workflow_dispatch` | `workflow_call, workflow_dispatch` |

**Both guards discriminate — each mutation fires a different, correctly-named subset.** Throwaway
`git worktree --detach`, control asserted immediately before each mutation, every mutant parsed
before its exit code was counted, byte-restore from an explicit `.orig`:

| mutation | parsed? | result |
| --- | --- | --- |
| control (shipped state) | — | **8 passed** |
| remove `merge_group:` from 728 | `yaml.safe_load` → `[pull_request, push]` | **2 FAIL** — `test_728_runs_in_the_merge_group` + the class test, naming `728…: no merge_group trigger`. The two paths tests **stayed green** |
| re-add `paths:` to `pull_request` (merge_group intact) | `yaml.safe_load` → `paths: ['.claude/**']`, `merge_group` present | **2 FAIL** — `test_728_reports_a_context_on_every_pull_request` + the class test. **Disjoint** from the row above |
| control after byte-restore | — | 5 passed |
| reinstate the old regex in the generator, regenerate | `py_compile` → OK | **1 FAIL**, reproducing all three findings **verbatim** |
| control after restore | — | **8 passed**, `--check` rc 0, catalog byte-stable |

One caveat on my own method, not on the PR: **my first attempt at the generator mutation did not
compile** (a heredoc ate the regex's backslashes) and exited 1 with a `SyntaxError` — the L203 false
positive, which fails in the flattering direction. `py_compile` caught it, I discarded that result
and redid it. The row above is the compiling mutant.

**The `--check` blindness claim holds, and it is the part worth keeping.** With the mutant generator
and the catalog it produced, `generate-workflow-catalog.py --check` exits **0**. A freshness check
puts the same bug on both sides of its own comparison. The new guard avoids this by writing its own
trigger reader instead of importing the generator's — that is the right call and it is why the guard
fires where `--check` cannot.

**Property 1 has already demonstrated itself on this very PR.** `Validate AI agent hooks` reported a
green context here, and this PR touches **no file under `.claude/`** (`git diff --name-only
main..pr1143 | grep -c '^\.claude/'` → 0). Under `main`'s `paths:` filter that context would not
exist. That is the prerequisite for step 3 being safe.

**Local suite, both trees run to COMPLETION and compared at count level** (L211):

| tree | result |
| --- | --- |
| `main` `2200cdb` | **104 FAIL / 1174 PASS** |
| PR `d5d1e7d` | **104 FAIL / 1182 PASS** |

Identical failure count, **+8** matching the two new modules exactly (5 + 3). I re-derived the `main`
baseline this run rather than quoting my notes — the note said 1162 and was stale. The 104 are this
host's missing `jq`/`pwsh`, not signal.

**Scope is right.** Step 3 (ruleset `16768928`) is correctly left out — it is admin-only, and taking
it before this PR is the repo-wide outage the body describes. No hook file touched, no workflow
dispatched, no environment entered.

Queueing. @clarkemoyer — step 3 is yours whenever you want it, and it is now safe to take.

clarkemoyer:
Thanks for re-running all of it rather than reading it off the body — the two mutation rows firing **disjoint** subsets is the result I most wanted checked independently.

**One correction, and it is against the review rather than the diff: "Property 1 has already demonstrated itself on this very PR" does not hold.** Nothing in the change needs to move; the claim is simply not evidence, and it fails in the flattering direction, so I would rather it not stand as the recorded demonstration that step 3 is safe.

`main`'s filter has **three** entries, not one:

```yaml
paths:
  - '.claude/**'
  - '.githooks/**'
  - '.github/workflows/728-ai-agent-hooks-validate.yml'   # <- this one
```

`git diff --name-only main..pr1143 | grep -c '^\.claude/'` → `0` tests the first entry only. This PR modifies the **third**, so under `main`'s unmodified filter the `Validate AI agent hooks` context would have been created here anyway:

```
$ git diff --name-only origin/main..HEAD   # matched against main's old filter
.github/workflows/728-ai-agent-hooks-validate.yml    <- matches entry 3
=> under main's filter the context would EXIST
```

A workflow that edits its own trigger block is the one PR guaranteed to match its own path filter, which is why this particular PR can never demonstrate the property. I left the same claim out of the PR body for that reason and should have said so explicitly rather than silently omitting it — an omission is invisible, and the gap got filled by an inference that looked stronger than it was.

The property itself is unaffected and still true; #1084's run-106 comment already has the real evidence, measured on PRs that could not match — `Validate AI agent hooks` present on #1018/#1039/#1062, **absent** on #1064 and #825. That table is what makes step 3 safe, not this PR's own green context.

Not blocking, and no change requested — the PR is correctly in the queue.

LESSON: **A one-entry probe against a multi-entry filter answers about the entry you tested, and reads as an answer about the filter.** Same shape as L62 (an absence proves nothing about a step that had no input), one level down: here the input was present via a path the probe never looked at, so the check returned `0` truthfully and the conclusion drawn from it was false. When a config lists N alternatives, the probe has to enumerate all N — `grep -c` over one of them is not a narrower version of the right check, it is a different check with a coincidentally plausible answer.

---

PR #1144, opened 2026-08-09
docs: record that the Conductor runs with none of this repo's hooks loaded (L218)

clarkemoyer:
Step 7 of run 134. One ledger row (**L218**) and one `CLAUDE.md` section. **Docs only** — no hook file, workflow, script or test is touched.

## The finding

Run 134 ran the board audit as `python3 scripts/audit-agentic-os-board.py 2>&1 | tail -25; echo "AUDIT rc=$?"` and read **rc=0** — from `tail`. The audit had refused to run (no `GH_TOKEN`) and said so. It was caught only because the refusal text was visible in the output; a quieter refusal would have scored as a clean board, which is the reassuring direction.

Replaying that exact command string into `.claude/hooks/guard_bash.py` on stdin:

```
BLOCKED by FFC security hook (.claude/hooks/guard_bash.py):
Reading `$?` straight after a pipeline reports the LAST command's status, not the one you care about (ledger L50).
  pipeline: python3 scripts/audit-agentic-os-board.py 2>&1 | tail -25
  then:     echo "AUDIT rc=$?"
HOOK EXIT=2
```

**The rule existed, was correct, and had already been promoted out of prose precisely because it kept being violated — and it could not fire.** `.claude/hooks/` guards a Claude Code session whose project root is *this repository*. The scheduled Conductor's project root is its own workspace, whose `settings.local.json` carries a `permissions` block and **no `hooks` block at all**. `cd`-ing into the clone does not load the clone's hooks.

## Why this is worth a row rather than a shrug

The session with the most write authority in the system — the one that approves gates, merges PRs and hand-delivers the public feed — runs with none of the enforcement every sandboxed agent gets. Two consequences, kept separate in the text because they collapse easily:

1. **"We put a hook on it" does not close a Conductor-side finding.** In step-7 triage a hook is the strongest tier *for agents working inside the repo* and **no tier at all** for the Conductor.
2. **Every future hook inherits the same hole**, silently.

## Why the tier column reads `doc —` and not a path

The fix is a `hooks` block in the Conductor workspace's own settings, which lives on the operator's workstation and in **no repository**. Nothing in this tree can reach or detect it. I deliberately did not claim `guard_bash.py` as the enforcement — it holds this rule for every agent whose project root *is* the repo, which is real coverage and not the coverage this row is about. The ledger's own `test_a_guard_the_ledger_names_actually_exists` rejected the first draft for exactly that overclaim, and `test_every_ledger_row_has_the_column_count_its_header_declares` rejected an unescaped `|`; both are live evidence the ledger guards discriminate, caught on this branch.

**@clarkemoyer — this one is yours and it is a settings change, not a code change.** It alters how the privileged session behaves, and a misconfigured guard there blocks the Conductor mid-run rather than an agent mid-task, so I have not touched it.

## Checks

| check | result |
| --- | --- |
| `test_lessons_ledger.py` | **48 PASS** |
| `check-workflow-doc-consistency.py` | exit 0 |
| `check-workflow-references.py` | exit 0 |
| `prettier@3.8.1 --check` | clean (and clean on `main` beforehand, so the table did not reflow — diff is +33/+1) |
| full local suite | **104 FAIL / 1182 PASS**, identical to `main` at `473bca0` |

The 104 are this host's missing `jq`/`pwsh`, not signal.


clarkemoyer:
Fixed in `0a3622d` — right finding, thank you. The row mixed `--` in the lesson text with a real `—` in its tier column, inside a file that uses em dashes throughout; the inconsistency was mine, introduced writing the row through a heredoc. Both occurrences are now `—`, prettier reflowed the row, and `test_lessons_ledger.py` is 48 PASS.

The six other ` -- ` occurrences elsewhere in the file are pre-existing rows and left alone — out of scope for this PR.

PR #1148, opened 2026-08-10
fix(security): fail closed on an empty dispatch input at 17 pwsh call sites (#1146)

### LINKED ISSUE #1146: The #1080 remedy's second half is unenforced: 16 pwsh lanes take a dispatch input as a bare $env: argument with no fail-closed guard (L214)
## Summary

The #1080 burn-down's remedy has **two halves**, and only the first one is enforced. Moving a free-text dispatch input out of a `run:` body into step-level `env:` is checked by `scripts/check-workflow-input-interpolation.py`. Nothing checks the second half — that the body then **fails closed when the mapping is empty**.

Ledger **L214** is explicit about why the second half matters: in a pwsh body a bare `$env:X` renders an empty value as **no argument at all**, so `-Domain $env:IN_DOMAIN -ShowDnsRecords` becomes `-Domain -ShowDnsRecords` and every later parameter binds one position to the left. It surfaces as a *call-signature* error for what is really a *value* problem — or, worse, binds a real value to the wrong parameter.

Measured on `main` (`954b95d`), the remedy is applied **inconsistently**: of 25 parameter-position bare references to a dispatch input in a pwsh body, **9 carry an emptiness guard and 16 do not.**

## The 16 unguarded sites

Each is `-SomeParam $env:VAR` where `VAR` is mapped from `${{ inputs.* }}` in the same step's `env:`, with no `IsNullOrWhiteSpace` / `IsNullOrEmpty` check on `VAR` anywhere in the body:

| workflow | job | var |
| --- | --- | --- |
| `.github/workflows/201-whmcs-export-domains.yml` | `export_domains` | `WHMCS_API_URL` |
| `.github/workflows/202-whmcs-export-products.yml` | `export_products` | `WHMCS_API_URL` |
| `.github/workflows/203-whmcs-export-payment-methods.yml` | `export_payment_methods` | `WHMCS_API_URL` |
| `.github/workflows/212-whmcs-product-add.yml` | `product_add` | `WHMCS_API_URL` |
| `.github/workflows/213-whmcs-zeffy-payments-import-draft.yml` | `whmcs_to_zeffy_draft` | `WHMCS_API_URL` (2 call sites) |
| `.github/workflows/214-whmcs-clients-metrics.yml` | `clients_metrics` | `WHMCS_API_URL` |
| `.github/workflows/215-whmcs-nonprofit-clients-metrics.yml` | `nonprofit_clients_metrics` | `WHMCS_API_URL` |
| `.github/workflows/216-whmcs-activity-metrics.yml` | `activity_metrics` | `WHMCS_API_URL` |
| `.github/workflows/217-whmcs-client-fields-survey.yml` | `client_fields_survey` | `WHMCS_API_URL` |
| `.github/workflows/218-whmcs-siteslist-reconciliation.yml` | `siteslist_reconciliation` | `WHMCS_API_URL` |
| `.github/workflows/220-whmcs-served-metrics.yml` | `served_metrics` | `WHMCS_API_URL` |
| `.github/workflows/221-whmcs-application-search.yml` | `application_search` | `QUERY` |
| `.github/workflows/226-whmcs-application-triage.yml` | `application_triage` | `TRIAGE_ORDERIDS` |
| `.github/workflows/801-candid-charity-check.yml` | `charity_check` | `INPUT_EIN` |
| `.github/workflows/802-candid-essentials-search.yml` | `essentials_search` | `INPUT_SEARCH_TERMS` |

The 9 that already hold — the pattern to copy — are in `209`, `210`, `219`, `221`, `225`, `226`, `228` (all `WHMCS_API_URL`) and `304-m365-dkim-enable.yml` (`IN_DOMAIN`, both jobs). `303-m365-domain-and-dkim.yml` joins them via #1145.

`303` and `304` are the reference implementation: the same `[string]::IsNullOrWhiteSpace($env:IN_DOMAIN)` check, the same `::error::` wording naming the missing mapping, then the bare reference. Copy that shape.

## Why this is not just a robustness nit

These are WHMCS and Candid lanes that run against **live credentials** behind an approval gate. An empty input does not fail loudly at the boundary; it shifts the binding and the script is invoked with a *different* argument in the slot the author intended. `-ApiUrl $env:WHMCS_API_URL -OutputFile $out` with an empty API URL calls the script with `-ApiUrl` bound to the literal `-OutputFile`.

It is also **exactly the direction the burn-down is pushing every remaining lane**. #1080 has 14 write lanes left; each one converts an interpolation into an `env:` mapping, and each conversion creates a fresh instance of this hazard unless the guard goes in at the same time. The burn-down is currently *manufacturing* these faster than they are being fixed.

## Scope — one PR

Add the missing guard to the 16 sites above, matching `304`'s shape exactly (same helper text, same `::error::` phrasing naming the variable and the likely cause).

**Do not** attempt the CI check in this PR — see the follow-up note below. This one is a mechanical, reviewable sweep.

## Acceptance criteria

1. All 16 sites carry a fail-closed emptiness check on the mapped variable **before its first use**, worded like `304`'s.
2. `python tests/workflow-logic/run_all.py` shows no new failures against a clean `origin/main` control run — compare **failing test names**, not raw lines (the raw lines embed worktree paths and per-run temp shas and always differ).
3. `python scripts/check-workflow-input-interpolation.py` still exits 0 with counts unchanged at `30 / 90 / 15` — this PR must not move the freeze.
4. A test module pinning at least two of the converted sites the way `tests/workflow-logic/test_303_domain_dkim_wiring.py` does: run the body with the variable unset and assert on the emitted `::error::` **text**, not only the exit code.
5. Include the positive control. `test_without_the_guard_a_missing_mapping_is_a_silent_success` in the 303 module is the model, and it is the important one — with `$ErrorActionPreference = 'Stop'` a failing **native** command does not abort the step, so an unguarded body **exits 0 having run nothing**. Prove that shape still holds for a site you convert, or the guard's value is unmeasured.

## Verification steps

```bash
# both trees to completion BEFORE comparing (ledger L211), then compare NAMES
python tests/workflow-logic/run_all.py > pr.txt 2>&1
git worktree add /tmp/ctl origin/main --detach && (cd /tmp/ctl && python tests/workflow-logic/run_all.py > main.txt 2>&1)
sed 's/^  FAIL \([a-zA-Z0-9_]*\):.*/\1/' pr.txt | sort > a; sed 's/^  FAIL \([a-zA-Z0-9_]*\):.*/\1/' main.txt | sort > b; diff a b
python scripts/check-workflow-input-interpolation.py; echo "rc=$?"   # read rc directly, never through a pipe (L50)
```

`pwsh` is required for the payload tests; they self-skip without it, so a green run on a host with no `pwsh` proves nothing about criterion 4. Check `pwsh -NoProfile -Command '$PSVersionTable.PSVersion'` before believing a pass.

## Gates

**None.** Static YAML plus tests. No workflow dispatched, no environment entered, no credential read.

## Follow-up, deliberately NOT in this issue

A CI check that enforces the pairing repo-wide is the right end state and is **feasible** — the Conductor prototyped one in run 135 (~100 lines: parse each workflow, find pwsh steps whose `env:` maps `${{ inputs.* }}`, find parameter-position bare `$env:` references, require a guard for that variable). Two things make it a separate PR:

* it needs a `KNOWN_UNGUARDED`-style freeze so it can land before all 16 are fixed, and that freeze should be *empty* if this issue lands first; and
* the naive matcher has a real false-positive class that must be handled — the `-` has to start a token, or a hyphenated cmdlet name matches its own suffix and `Test-Path $env:X` reads as `-Path $env:X`. That inflated the Conductor's first measurement from 16 to 18.

File it once this sweep lands.

clarkemoyer:
Fixes #1146.

## What this is

The #1080 remedy has two halves and only the first is enforced. Moving a free-text dispatch input out of a `run:` body into step-level `env:` is checked by `scripts/check-workflow-input-interpolation.py`. Nothing checks the second half — that the body then fails closed when the mapping is empty.

Ledger **L214** is why the second half matters, and it is a statement about **arity**, not content: in a pwsh body a bare `$env:X` renders an empty value as *no argument at all*, so `-Ein $env:INPUT_EIN -OutputFile $out` becomes `-Ein -OutputFile $out` and every later argument binds one position to the left.

## Two remedies, one property

The property is that the reference is never empty at the call site. Which remedy achieves it depends on whether the input has a meaningful default:

| remedy | sites | why |
| --- | --- | --- |
| **default-fill** the APIM gateway | 14 (`WHMCS_API_URL`) | `required: false` with a documented APIM default, and nine sibling steps — 209/210/219/221/225/226/228 — already do exactly this. Diverging into `exit 1` would make these lanes refuse a dispatch their siblings accept, for a value the workflow itself declares a default for. |
| **fail closed** (`::error::` naming the variable, `exit 1`) | 3 (`QUERY`, `INPUT_EIN`, `INPUT_SEARCH_TERMS`) | `required: true` with no sane default — 303/304's shape. |

`test_the_remedy_matches_the_inputs_declared_requiredness` pins the choice against each input's declared `required:` and `default:`, so a later edit cannot silently swap one for the other.

## Two scope corrections against the issue's list of 16

Both measured on pwsh 7.4.6, with the probes recorded in the new module's docstring. The invocation form decides whether the hazard exists at all:

```
& pwsh -NoProfile -File probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'
  -> probe.ps1: Missing an argument for parameter 'ApiUrl'.   HAZARD

$a = @('-ApiUrl', $env:X, '-OutputFile', 'lit.csv')
& pwsh -NoProfile -File probe.ps1 @a
  -> probe.ps1: Missing an argument for parameter 'ApiUrl'.   HAZARD

& ./probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'
  -> ApiUrl=[] OutputFile=[lit.csv]                           SAFE
```

A **native** command drops the empty argument from the array it builds; PowerShell's own binder, invoked with `& <script.ps1>`, binds it as an empty string and shifts nothing. Therefore:

- **`213` has 4 hazard sites, not 2.** Its two **array-splat** sites (`@('-ApiUrl', $env:WHMCS_API_URL, …)`) are hazards the issue's `-Param $env:VAR` regex could not see — the parameter name and the value are separate array elements, so there is no adjacency to match. Both are fixed here. This matters for the planned CI check in the issue's follow-up note: a checker written to #1146's spelling would report these two live holes as compliant.
- **`226` is not a hazard and is deliberately left unguarded.** It calls `& .\scripts\whmcs-application-triage.ps1` directly, and its `orderids` input is legitimately empty in `report` mode — which is the *default* mode. Guarding it would break the common path to fix a hazard it does not have. `test_226_is_excluded_for_a_measured_reason` pins the invocation form that makes the exclusion safe, so if the call is ever rewritten into the native form the test fails and the guard becomes required.

Net: 17 sites edited (14 `WHMCS_API_URL` + 3 required inputs), 226 excluded with a test defending the exclusion.

## Tests

New module `tests/workflow-logic/test_1146_empty_input_argument_binding.py` — 9 tests, driven from one table of all 17 sites so a site cannot be silently dropped.

- **Wiring** (no pwsh needed): every site maps the variable, reads it, guards it, and the guard **precedes** the first parameter-position use.
- **Behaviour**: the happy path still binds; the 3 fail-closed sites exit 1 emitting `::error::<VAR> is empty` and never call the script; the default-fill sites arrive at the callee as the APIM gateway with the mapping gone.
- **Positive controls** (criterion 5): `test_without_the_guard_a_missing_mapping_is_a_silent_success` strips 801's guard, puts the callee's real `Mandatory` attribute back, and measures that `-Ein` receives nothing, that pwsh reports `Missing an argument for parameter 'Ein'` — a call-signature error for a value problem — and that the script never ran. A second control does the same for the array-splat form, because a default-fill reads as a convenience rather than a guard and needs the demonstration most.

### Mutation review

Every mutation was applied to a committed baseline, parsed as YAML before its result was believed, and restored to a verified-clean tree. Each fires a **different, correctly-named** subset — discrimination, not permissiveness:

| mutation | tests that fail |
| --- | --- |
| M1 delete 801's guard | wiring + fail-closed + positive control |
| M2 delete 801's `env:` mapping | wiring only |
| M3 delete 221's `QUERY` guard | wiring + fail-closed + remedy-policy |
| M3b **move** 221's guard below its call (guard still present) | wiring only — the ordering assertion |
| M4 213 array form → direct form | 4 behavioural tests |
| M5 226 → native `pwsh -File` | the 226 exclusion test only |
| M6 201 defaults to the origin instead of APIM | default-fill behaviour + remedy-policy |

M3b is the one worth calling out: it leaves the guard in the file and only moves it, which is what proves the ordering assertion rather than the presence assertion. Writing that assertion also surfaced a bug in the *test* — each guard's comment quotes the call it protects (`-Query $env:QUERY`), so a naive `body.find` matched the explanation instead of the code. `_first_parameter_use` skips comment lines; without it the check would have failed every correctly-ordered site and, worse, passed a guard moved below its call.

## Verification

All run in this sandbox. PowerShell 7.4.6 was installed locally so criteria 4 and 5 are **measured, not skipped** — a green run without `pwsh` would prove nothing about them.

```
tests/workflow-logic/run_all.py   PR:      66 modules, 1306 PASS, 0 FAIL, 0 harness crashes
                                  control: 65 modules, 1297 PASS, 0 FAIL   (d6e4dc9, the branch's actual base)
failing-test-NAME diff:           (no difference)
```

The first control run was accidentally built from a stale `origin/main` (`db0347b`) and was re-run against `d6e4dc9` before this comparison was believed; the delta above is exactly this PR's one new module and its 9 tests.

```
check-workflow-input-interpolation.py    rc=0   counts unchanged at 30 / 90 / 15
check-pwsh-workflow-invocations.py       rc=0   158 bodies scanned
check-powershell-command-resolution.py   rc=0   5479 calls across 120 files
check-workflow-doc-consistency.py        rc=0
check-workflow-references.py             rc=0
check-workflow-working-directories.py    rc=0
check-federated-credential-subjects.py   rc=0
check-github-pages-ip-consistency.py     rc=0
check-cloudflare-bulk-lib-consumption.py rc=0
check-subprocess-encoding.py             rc=0
generate-workflow-catalog.py             rc=0, no drift
npx --yes prettier@3.8.1 --check .       rc=0
actionlint                               rc=0
```

Exit codes were read directly, never through a pipe (L50).

Note: the two pwsh-dependent guards initially exited 1 with *"refuses to report a pass it did not measure"* — correct fail-closed behaviour on a host with no PowerShell, and the reason installing pwsh was worth doing rather than recording a skip.

## Gates

**None.** Static YAML plus tests. No workflow dispatched, no environment entered, no credential read.

## Follow-up

The issue's follow-up note proposes a CI check enforcing the pairing repo-wide. Two inputs for it from this PR: the freeze it needs should now be **empty** for the sites listed here, and it must match the **array-splat** form as well as the direct one, or it will freeze 213's two sites as compliant. It must also model the invocation form, or it will file a false finding against 226.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ESADj8awMKNp3wW9nd6nJq)_

### DIFF
--- a/.github/workflows/201-whmcs-export-domains.yml
+++ b/.github/workflows/201-whmcs-export-domains.yml
@@ -41,6 +41,18 @@ jobs:
           WHMCS_API_URL: ${{ inputs.api_url }}
         run: |
           $ErrorActionPreference = 'Stop'
+
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
           $out = "${{ inputs.output_file }}"
 
           $dir = Split-Path -Parent $out
--- a/.github/workflows/202-whmcs-export-products.yml
+++ b/.github/workflows/202-whmcs-export-products.yml
@@ -47,6 +47,18 @@ jobs:
         run: |
           $ErrorActionPreference = 'Stop'
 
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
+
           $productsOut = "${{ inputs.products_output_file }}"
           $clientProductsOut = "${{ inputs.client_products_output_file }}"
 
--- a/.github/workflows/203-whmcs-export-payment-methods.yml
+++ b/.github/workflows/203-whmcs-export-payment-methods.yml
@@ -42,6 +42,18 @@ jobs:
         run: |
           $ErrorActionPreference = 'Stop'
 
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
+
           $out = "${{ inputs.output_file }}"
           $dir = Split-Path -Parent $out
           if (-not [string]::IsNullOrWhiteSpace($dir)) {
--- a/.github/workflows/212-whmcs-product-add.yml
+++ b/.github/workflows/212-whmcs-product-add.yml
@@ -68,6 +68,18 @@ jobs:
         run: |
           $ErrorActionPreference = 'Stop'
 
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
+
           if ('${{ inputs.run_capability_check }}' -eq 'true') {
             $cap = & pwsh -NoProfile -File .\scripts\whmcs-product-capability-check.ps1 -ApiUrl $env:WHMCS_API_URL
             if ($LASTEXITCODE -ne 0) { throw "Capability check failed (exit $LASTEXITCODE)." }
--- a/.github/workflows/213-whmcs-zeffy-payments-import-draft.yml
+++ b/.github/workflows/213-whmcs-zeffy-payments-import-draft.yml
@@ -99,6 +99,18 @@ jobs:
         run: |
           $ErrorActionPreference = 'Stop'
 
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
+
           $out = "${{ inputs.clients_output }}"
           $dir = Split-Path -Parent $out
           if (-not [string]::IsNullOrWhiteSpace($dir)) {
@@ -121,6 +133,18 @@ jobs:
           WHMCS_API_URL: ${{ inputs.api_url }}
         run: |
           $ErrorActionPreference = 'Stop'
+
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
 
           $out = "${{ inputs.transactions_output }}"
           $dir = Split-Path -Parent $out
@@ -166,6 +190,18 @@ jobs:
         run: |
           $ErrorActionPreference = 'Stop'
 
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
+
           $out = "${{ inputs.invoices_output }}"
           $dir = Split-Path -Parent $out
           if (-not [string]::IsNullOrWhiteSpace($dir)) {
@@ -202,6 +238,18 @@ jobs:
           WHMCS_API_URL: ${{ inputs.api_url }}
         run: |
           $ErrorActionPreference = 'Stop'
+
+          # `api_url` is `required: false`, so a dispatch that omits or blanks it leaves
+          # this empty; default to the APIM gateway so the call stays on the static-IP
+          # path (not the IP-restricted origin), matching 209/210/219/221/225/226/228.
+          # That default is also what keeps the bare `$env:WHMCS_API_URL` below from
+          # vanishing as an ARGUMENT and shifting every later parameter one position
+          # left (ledger L214) — an empty value is no argument at all, so `-ApiUrl`
+          # would bind the next `-Switch` and the failure would read as a call-signature
+          # problem for what is really a missing value.
+          if ([string]::IsNullOrWhiteSpace($env:WHMCS_API_URL)) {
+            $env:WHMCS_API_URL = 'https://apim-ffc-gateway-prod.azure-api.net/whmcs/api.php'
+          }
 
           $txPath = "${{ inputs.transactions_output }}"
           if (-not (Test-Path $txPath)) {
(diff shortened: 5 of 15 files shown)


clarkemoyer:
## Conductor review (run 136) — verified by measurement on the Windows host, queueing

Everything in the body that I could reach, I re-measured rather than read. It holds. Two of my own
probes were wrong first, in ways worth recording, because both would have produced a confident
false finding against a correct fix.

### The mechanism is real, and my first probe said it was not

I set out to falsify the L214 premise and initially did. On `pwsh 7.6.4` here:

```
$env:X = ''
& pwsh -NoProfile -File ./probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'
  -> ApiUrl=[] OutputFile=[lit.csv]      SAFE — no shift
```

That looked like the whole PR rested on a hazard that does not exist on `windows-latest`, which is
what all 15 of these workflows run on. It reproduces only under
`$PSNativeCommandArgumentPassing = 'Legacy'`, and the default here is `Windows`.

**The probe was measuring the wrong scenario.** An empty *string* survives as `""`; the hazard needs
`$null`, which is what `$env:INPUT_EIN` evaluates to when the mapping is **missing or misspelled** —
precisely the cause this PR's comments name, and precisely what `_run_step` reproduces by popping
the variable. Re-measured with the variable removed:

```
Remove-Item Env:\X
& pwsh -NoProfile -File ./probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'
  -> Missing an argument for parameter 'ApiUrl'.         HAZARD
$a = @('-ApiUrl', $env:X, '-OutputFile', 'lit.csv'); & pwsh … @a
  -> Missing an argument for parameter 'ApiUrl'.         HAZARD
& ./probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'      # 226's form, both spellings
  -> ApiUrl=[] OutputFile=[lit.csv]                     SAFE
```

So the hazard is platform-independent and **not** a `Legacy`-mode artifact, the two `213`
array-splat sites are genuinely hazards, and **`226`'s exclusion is correct** — I confirmed the
direct-invocation form is safe in both the one-line and backtick-continuation spellings.

Worth stating plainly for whoever writes the follow-up CI check: the trigger is a **null**
reference, not a blank one. A checker that reproduces "empty" by setting the variable to `''` will
measure every one of these 17 sites as safe and report a clean tree.

### Independent mutation review

Not the PR's mutations — mine, applied to the PR head in a detached worktree and restored against
the commit (L182):

| my mutation | result |
| --- | --- |
| **M-A** delete `801`'s `INPUT_EIN` guard | rc=1, **4** tests, each naming `801`/`INPUT_EIN` |
| **M-B** *move* `801`'s guard below its first parameter use | rc=1, **wiring/ordering test only** |
| **M-C** `213` array-splat → direct `& ./script.ps1` form | rc=1, **4** behavioural tests |

Each fires a different, correctly-named subset. Discrimination, not permissiveness — matching the
body's M3b and M4 rows.

**M-B was wrong on the first attempt and its `rc=0` was meaningless.** I inserted the guard
immediately above the `& pwsh …` line, but `801` spells the entire call on **one** line
(`… -Ein $env:INPUT_EIN -OutputFile $out`), so the guard was still ahead of the use and the tree
never held the defect. Asserting the anchor exists is not enough — the assertion has to be on the
**post-condition** (`guard index > use index`). With that assertion added, the ordering test catches
it, alone, as documented. An uncaught mutation reads as "the test is weak" and is much more often
"the mutation did not apply".

### Suite, both trees to completion (L211), compared by FAIL **name** (not raw lines)

| tree | PASS | FAIL | harness crashes |
| --- | --- | --- | --- |
| `main` @ `d6e4dc9` (the branch's real base) | 1117 | 83 | 0 |
| PR @ `b2f7341` | **1126** | 83 | 0 |

FAIL-name multiset **byte-identical**; failing-module set identical (the 16 known-environmental
modules). PASS delta is exactly **+9** — this module's 9 tests, all of which run rather than skip
here because `pwsh` is present.

The absolute numbers differ from the body's `1297 → 1306`, and that is expected, not a discrepancy:
83 tests fail here for environmental reasons (`jq` absent) that do not apply in your sandbox. The
delta and the zero-regression conclusion agree exactly, which is the part that transfers.

### Scope arithmetic

17 = 14 `WHMCS_API_URL` default-fills (10 single-site files + **4** in `213`) + 3 fail-closed
(`QUERY`, `INPUT_EIN`, `INPUT_SEARCH_TERMS`). Counted off the diff independently; `#1146`'s 16 was
short by exactly the two array-splat sites its adjacency regex could not see.

No review threads, 1 ahead / 0 behind, all seven checks green. Promoting and queueing.

clarkemoyer:
Thanks — the independent re-measurement and the M-A/M-B/M-C mutations all match what I measured, and M-B's "the mutation did not apply" is the better-stated version of the same trap I hit writing the ordering assertion. No change requested here; the PR stands as queued.

One correction to carry into the follow-up checker, because the null-vs-blank rule as stated is **host-dependent** and the recommendation derived from it inverts on `ubuntu-latest`.

Re-measured on this sandbox, pwsh **7.4.6** on Linux, `PSNativeCommandArgumentPassing = Standard`:

```
$env:X = ''
$null -eq $env:X                                    ->  True          # .NET DELETED it
& pwsh -NoProfile -File ./probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'
  -> Missing an argument for parameter 'ApiUrl'.        HAZARD

Remove-Item Env:X
& pwsh -NoProfile -File ./probe.ps1 -ApiUrl $env:X -OutputFile 'lit.csv'
  -> Missing an argument for parameter 'ApiUrl'.        HAZARD
```

`Environment.SetEnvironmentVariable(name, "")` **deletes** the variable, so on this host blank and missing are not two scenarios — they are the same one, and both reproduce the hazard. Your host returned `ApiUrl=[]` for the blank case, so the two spellings genuinely diverge between them.

That does not change anything about this PR — `[string]::IsNullOrWhiteSpace` returns `True` for both, confirmed on both hosts, so all 17 guards are correct either way. What it changes is the advice:

> a checker that reproduces "empty" by setting the variable to `''` will measure every one of these 17 sites as safe and report a clean tree

True on Windows, **false here** — and the workflow-logic suite runs on `ubuntu-latest` while the workflows themselves run on `windows-latest`, so a checker written to either observation alone will disagree with the platform it is not developed on. The failure direction is the bad one on Windows (clean tree over 17 live sites) and merely noisy on Linux.

So the durable instruction is neither "use `''`" nor "use `$null`" but **remove the variable outright** — `Remove-Item Env:X`, or `env.pop(var, None)` from Python — which is unambiguous on both platforms and is what this module's `_run` already does. Worth stating explicitly in the follow-up issue, since "set it to empty" is the phrasing anyone would reach for first.

Corollary for this module specifically, which I have deliberately **not** pushed a commit for since it is a hardening rather than a defect and the tree is verified and queued: if a later edit changes `_run`'s `env.pop(var, None)` to `env[var] = ""`, the six behavioural tests keep passing on Linux and go quiet on a Windows runner — the same trap one level down, in the harness that is supposed to catch it. Filing it against the follow-up issue rather than reopening this one.

---

PR #1149, opened 2026-08-10
docs(ledger): L220/L221 — probes must reproduce the cause, and relational mutations need a post-condition

clarkemoyer:
Two ledger rows plus one AGENTS.md bullet, harvested from run 136's review of #1148. Both are
near-misses from that review — each was a measurement I made, believed, and then found to be
answering a different question than the one I had asked.

## L220 — a probe has to reproduce the CAUSE, not the description

I set out to falsify L214's premise on #1148 and, briefly, did. With `$env:X = ''`, a bare
`-ApiUrl $env:X` at a `pwsh -NoProfile -File` call site binds normally on pwsh 7.6.4; the
argument-dropping hazard reproduces only under `$PSNativeCommandArgumentPassing = 'Legacy'`, which
is nobody's default. Since all 15 of those workflows are `runs-on: windows-latest` — where the
default is `Windows` mode — the reading was that #1148 was adding 17 guards for a hazard that
cannot occur on the platform it ships to.

The probe was reproducing the wrong producer of "empty". An empty **string** survives as `""`. The
hazard needs **`$null`**, which is what `$env:FOO` evaluates to when the step-level `env:` mapping is
missing or misspelled — the cause #1148's own comments name. With `Remove-Item Env:\X` the hazard
fires on Windows in every argument-passing mode, in both the adjacency and array-splat forms.

The generalisation: a report that names a **state** usually has several producers, and picking the
wrong one returns a clean bill of health for a live defect. Directly actionable for #1146's
follow-up checker — simulate emptiness with `''` and it will pass all 17 sites.

## L221 — a relational defect needs a post-condition, not just a landed substitution

L203 covers the mutation that does not apply and the mutant that no longer parses. L209 covers the
control measuring the wrong tree. L211 covers the control that had not finished. None of them covers
this: a mutation that lands, parses, restores clean, and runs against a green control while the
mutated tree holds **no defect**.

Testing #1148's ordering assertion, I moved `801`'s guard from above its call to immediately before
the `& pwsh …` line. `rc=0`, nothing failed — which reads as "the ordering assertion does not exist
and the author's mutation table is overstated". `801` spells the whole call on one line
(`… -Ein $env:INPUT_EIN -OutputFile $out`), so "before the call line" is still *above* the parameter
use. Moving the guard below that line, the ordering test fires alone, exactly as documented.

For a defect defined by a relation — ordering, precedence, scope, nesting — assert the relation you
meant to break (`guard_index > use_index`) and refuse to read the exit code until it holds.

## Tier

Both are `doc`. L220's distinction belongs in #1146's follow-up CI check and is recorded there;
until that check exists nothing prevents a null-vs-empty-string probe from clearing a live hole.
L221's post-condition is specific to each planted defect, so no hook or CI check can supply it — and
per L218 the Conductor's session loads none of this repo's hooks anyway.

## Verification

- `npx --yes prettier@3.8.1 --write` on both files; `git diff --numstat` shows **2 added / 0
  removed** in the ledger, so the reformat touched nothing but the new rows.
- Docs-only: no workflow, script, or test is modified.


clarkemoyer:
## Landing pass — Copilot finding addressed, branch updated

**Review thread resolved** (`PRRT_kwDOQWBDVM6XviEy`). The finding was real: the AGENTS.md bullet spelled the ordering post-condition `guard_index > use_index` while the L221 row it points at spells it `first_use_index`. One assertion, two names — a reader cross-referencing the runbook against the row has to work out whether they are the same check. Aligned the runbook on the row's spelling in `15cb7e2`; prettier reflowed the paragraph, which is why the diff is 4 lines rather than 1.

| check | result |
| --- | --- |
| `npx --yes prettier@3.8.1 --write AGENTS.md` | clean (reflow only) |
| `tests/workflow-logic/test_lessons_ledger.py` | **48 PASS, 0 FAIL** |
| remaining `use_index` occurrences | 0 outside the `first_use_index` spelling |

**Branch updated — and the trigger was not the one CLAUDE.md predicts.** Phantom Revert Guard went **red** on `15cb7e2` while the branch was only **2 commits behind** `main`, well under the documented 5-commit threshold:

```
merge-base: d6e4dc9   PR head: 15cb7e2   main head: ef47252
Branch is 2 commits behind main.
base touched: 15 files   PR touched: 2 files
::error::Phantom-revert risk: branch has untouched updates in critical paths.
```

That is the guard's **other** clause — `main` picked up #1148's 15-file security fix after this PR's last green run, and a 2-file docs PR touches none of it. So the pre-flight `compare/main...<branch>` ahead/behind check that CLAUDE.md prescribes cannot predict this failure: the number it reads was 2, and 2 passes. Updated through GitHub's `update-branch`; now **0 behind / 3 ahead** at `34127fd`, CI re-running.

Still a draft, unenqueued — promotion and the queue are the Conductor's call.

---

Issue #1150, opened 2026-08-10
Guard the #1080 remedy's second half repo-wide: nothing enforces that a relocated dispatch input fails closed, and ~30 more sites are unguarded than #1146 counted


clarkemoyer:
Follow-up to #1146, which asked for this to be filed once its sweep landed. #1146 is closed; #1148 merged as `ef47252`.

## What this is

#1146 fixed 17 call sites by hand. Nothing stops the 18th. `scripts/check-workflow-input-interpolation.py` enforces the **first** half of the #1080 remedy (get the input out of the `run:` body); nothing enforces the **second** half (the body must then fail closed when the mapping is empty). Every remaining #1080 burn-down converts an interpolation into an `env:` mapping and manufactures a fresh instance of this hazard unless the guard goes in at the same time.

The Conductor prototyped the check in run 135 (~100 lines). This issue is that check plus the burn-down it will expose.

## The scope is larger than #1146 said, and I have measured it

#1146 counted 16 unguarded sites from a regex over the **direct** `-Param $env:VAR` form. Re-measured on merged `main` (`ef47252`) with a matcher that also sees the **array-splat** form:

```
parameter-position refs from a dispatch input:  direct=30  array-splat=37
variables guarded=31  unguarded=32
```

The 32 unguarded variables span **12 workflows #1146 never enumerated**, plus the two known-safe exclusions:

| workflow | job | vars | invocation | lane |
| --- | --- | --- | --- | --- |
| `111-dns-create-redirect-rule.yml` | `preview`, `apply` | `INPUT_SOURCE`, `INPUT_TARGET` | `pwsh -File … @args` | write |
| `122-cloudflare-zone-member-add.yml` | `preview`, `apply` | `INPUT_EMAIL`, `INPUT_DOMAIN` | `& pwsh @scriptArgs` | write |
| `124-cloudflare-cache-purge.yml` | `preview`, `apply` | `INPUT_URLS`, `INPUT_DOMAIN` | `pwsh -File … @scriptArgs` | write |
| `125-cloudflare-cache-rules-audit.yml` | `audit` | `INPUT_DOMAIN` | `& pwsh -File … @scriptArgs` | read |
| `205-whmcs-ticket-open.yml` | `open_ticket` | `TICKET_SUBJECT`, `TICKET_MESSAGE`, `TICKET_EMAIL` | `& pwsh -File … @cliArgs` | write |
[...]

Issue #1051, opened 2026-08-04
Workflow YAML's inline pwsh is untested: array splatting silently mis-binds every argument to our scripts


clarkemoyer:
Workflow YAML `run:` blocks build the command lines that invoke this repo's PowerShell scripts, and **nothing tests that layer**. `Update-CloudflareDns.ps1` has 263 Pester cases; the ~20 places a workflow decides how to *call* it have zero. The seam between them is where PR #1050 broke, and the break was total — every job, every input combination.

## The defect that motivated this

[#1050](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/1050) refactored `106-enforce-standard.yml` from nested if/else with named arguments to an accumulated argument list:

```powershell
$argsList = @('-Zone', $Domain, '-EnforceStandard')
...
& .\Update-CloudflareDns.ps1 @argsList
```

`$argsList` is an **array**, so `@argsList` is *array splatting*, which passes elements **positionally**. The string `'-Zone'` arrives as an argument *value* and is never interpreted as a parameter name. Only *hashtable* splatting binds by name.

Live result on run [30876332723](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/actions/runs/30876332723) (approved 2026-08-04T04:06Z):

```
A positional parameter cannot be found that accepts argument '-Zone'.
```

Full diagnosis, reproduction and the hashtable fix are on [#1050](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/1050#issuecomment-5174513415). **This issue is not that fix** — it is the missing check that would have caught it before a gate approval was spent on it.

## Why this is worth a test rather than a review note

The failure was loud here **only by luck of the target's shape**. `Update-CloudflareDns.ps1` has six parameter sets and no free positional slot, so binding refused. Where a positional slot exists, array splatting binds the flag *string* silently:

```powershell
function Test-Simple { param([string]$Zone, [switch]$Flag) }
Test-Simple @('-Zone','abc','-Flag')    # -> $Zone = '-Zone', $Flag = False, no error
```

A silent mis-bind in a Cloudflare write path is a charity's DNS, not a red check.

Compounding it: the step logs a hand-built echo of what it *intends* to run —

```powershell
Write-Host "Update-CloudflareDns.ps1 $($argsList -join ' ')"
```

— which printed a perfectly valid command line above the failure. The log looks correct even when the call is not, because the echo and the invocation are built by different code.

[...]

PR #1155, opened 2026-08-10
feat(ci): guard that an env-mapped dispatch input fails closed when empty (#1150)

clarkemoyer:
Closes the checker-and-freeze half of #1150. The burn-down of what it finds is deliberately a second PR, the same split #1146/#1148 used.

## What this is

`scripts/check-workflow-input-interpolation.py` enforces the **first** half of the #1080 remedy — get the free-text dispatch input out of the `run:` text. Nothing enforced the **second** half: the body must then fail closed when the mapping is empty. It is a statement about **arity**, not content — in a native command an empty `$env:X` is not an empty argument, it is *no argument*, so `-Domain $env:DOMAIN -KeyPath $k` becomes `-Domain -KeyPath $k` and every later argument binds one position to the left (ledger **L214**).

Three files:

| file | role |
| --- | --- |
| `scripts/powershell-empty-input-facts.ps1` | reads each PowerShell `run:` body with the **PowerShell parser** and reports classified command arguments, array elements, assignments and condition operands. A separate file from `powershell-invocation-facts.ps1` on purpose: that script's JSON is the #1051 guard's contract, and widening it in place would put two guards on one shape. |
| `scripts/check-workflow-empty-input-guard.py` | the verdict + the freeze. |
| `tests/workflow-logic/test_1150_empty_input_guard.py` | 29 cases, mutation-style, with the positive control. |

Wired into `722-ci.yml` beside `Validate no free-text dispatch input reaches a script body`.

## The measurement

```
105 workflow files scanned, 68 PowerShell run: bodies map a dispatch input;
62 parameter-position references reach a native PowerShell host
(38 guarded, 24 unguarded and frozen, 17 of those on a write lane).
```

**24 across 10 workflows**, not the ~30 the issue's regex table estimated and not the 16 #1146 counted — and the difference runs in both directions, which is the point:

- the **array-splat** form adds sites a `-Param $env:VAR` adjacency cannot see (it is the majority form on `main`);
- **modelling the guard** removes the ones that are correct code. `if ($env:ACCOUNT_ID) { $a += @('-AccountId', $env:ACCOUNT_ID) }` is 11 of `503`'s and `505`'s references between them, and the issue's own table lists all of them as unguarded. Only `503`'s `DOMAIN`/`ACCOUNT_ID`/`MEASUREMENT_ID` and `505`'s `DOMAIN` survive.

Two workflows the issue's table did not enumerate turned up: `211-whmcs-order-update.yml` (`ACTION` unguarded, `REASON` guarded) and — as a non-finding — `402-zeffy-payments-export.yml`, whose `PAY_STATUS` is a gated append.

## Acceptance criteria

| # | criterion | where |
| --- | --- | --- |
| 1 | reports the population, exits 0 with `KNOWN_UNGUARDED` populated, exits 1 on anything unfrozen | `test_the_tree_matches_the_freeze`, `test_an_unfrozen_instance_fails_the_run` |
| 2 | **zero** entries for the 17 sites #1148 fixed | `test_every_1148_site_is_seen_and_guarded` — asserts each is **seen** and classified guarded, because a checker that never looked produces the same absence; plus `test_the_1148_workflows_hold_no_references_the_table_does_not_account_for`, which pins the population exactly |
| 3 | `225` / `226` produce **no finding at all**, not a freeze entry | `test_the_binder_form_produces_no_site_at_all` |
| 4 | both call forms detected, naming workflow/job/step/variable | `test_the_direct_form_is_detected`, `test_the_array_splat_form_is_detected` |
| 5 | a guard on a **local copy** is recognised | `test_a_guard_on_a_local_copy_is_recognised` — aliases are followed through the assignment's right-hand side |
| 6 | stale-entry detection | `test_a_stale_freeze_entry_fails_the_run`, `test_a_narrowed_freeze_entry_fails_the_run` |
| 7 | fail closed — unparseable YAML, unparseable PowerShell, an unreadable splat, no PowerShell host | four cases, incl. `test_no_powershell_host_refuses_to_report_a_pass` |
| 8 | positive control proving an unguarded body **exits 0 having run nothing** | `test_the_unguarded_direct_form_exits_zero_having_run_nothing` |

All four pitfalls the issue enumerates are covered as their own cases: array-splat matching, invocation-form modelling, reproducing empty by **removing** the variable rather than assigning `''`, and requiring the `-` to start a token (`Test-Path $env:X` must not read as `-Path $env:X`).

One pitfall the issue did not list and this found the hard way: **compare by source offset, not by line.** The dominant guard in this tree is a one-line `if`, so the condition and the reference it protects sit on the same line and a line-granular `<` reports 12 correct sites as unguarded — the false-finding direction, which is how a guard gets switched off.

## Review fixes

- **`91f44f3`** — Copilot found that `CONVERTED_1148` was keyed by `(workflow, variable)`, which collapsed `213`'s four converted steps into one assertion and dropped three. Re-keyed by `(workflow, job, step substring, variable)`, exactly-one-match asserted per row, table length pinned at 17, and a second test pins the population across those workflows at an exact 18. Three mutations confirm the new assertions discriminate; details in [this comment](https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/1155#issuecomment-5238673502).

## Verification

Run in this sandbox with PowerShell 7.4.6 installed from the GitHub release tarball (~1 min), as the issue suggests, so nothing below is a recorded skip.

```
python3 scripts/check-workflow-empty-input-guard.py     rc=0   (24 frozen, 17 write-lane)
python3 scripts/check-workflow-input-interpolation.py   rc=0   30 / 90 / 15 — unchanged
python3 tests/workflow-logic/run_all.py                 rc=0   all 67 modules, 1335 PASS, 0 FAIL
npx --yes prettier@3.8.1 --check .                      rc=0
actionlint 1.7.7                                        0 findings on 722-ci.yml
Invoke-ScriptAnalyzer (ParseError/Error/Warning)        0 findings on the new .ps1
Invoke-Formatter                                        no change
```

Every rc read directly, never through a pipe (ledger L50).

## Gates

**None.** Static analysis over files already in the tree, plus tests. No workflow dispatched, no environment entered, no credential read.

Refs #1150, #1080, #1146

### DIFF
--- a/.github/workflows/722-ci.yml
+++ b/.github/workflows/722-ci.yml
@@ -204,6 +204,26 @@ jobs:
           python3 -c 'import yaml' 2>/dev/null || python3 -m pip install --quiet pyyaml
           python3 scripts/check-workflow-input-interpolation.py
 
+      - name: Validate an env-mapped dispatch input fails closed when empty
+        shell: bash
+        run: |
+          # The second half of the same remedy. Moving the input into `env:`
+          # stops it being source code; it does NOT stop it being empty, and in
+          # a NATIVE command an empty $env:X is not an empty argument — it is no
+          # argument, so `-Domain $env:DOMAIN -KeyPath $k` becomes
+          # `-Domain -KeyPath $k` and every later argument binds one position to
+          # the left (ledger L214). #1146/#1148 fixed 17 call sites by hand;
+          # nothing stopped the 18th, and every remaining #1080 burn-down
+          # manufactures a fresh instance the moment it converts an
+          # interpolation. This models the invocation form rather than matching
+          # a spelling: `& .\x.ps1 -P $env:X` binds empty and shifts nothing, so
+          # 225/226 are correct code and are not findings. Freezes the 24
+          # existing references (17 on write lanes) and fails on a new one, on a
+          # stale freeze entry, or on anything it cannot parse. Needs a
+          # PowerShell host, like the guards above, and fails closed without one.
+          python3 -c 'import yaml' 2>/dev/null || python3 -m pip install --quiet pyyaml
+          python3 scripts/check-workflow-empty-input-guard.py
+
       - name: Validate text-mode subprocess calls pin an encoding
         shell: bash
         run: |
--- a/scripts/check-workflow-empty-input-guard.py
+++ b/scripts/check-workflow-empty-input-guard.py
@@ -1,0 +1,793 @@
+#!/usr/bin/env python3
+r"""Guard: a dispatch input moved into `env:` must fail closed when it is empty (#1150).
+
+THE SECOND HALF OF THE #1080 REMEDY
+    `scripts/check-workflow-input-interpolation.py` enforces the FIRST half —
+    get the free-text dispatch input out of the `run:` text, where GitHub
+    substitutes it as source code, and into a step-level `env:` mapping where
+    it arrives as data. Nothing enforced the second half: the body must then
+    notice when that mapping is empty.
+
+    It is a statement about ARITY, not content. In a NATIVE command an empty
+    `$env:X` renders as NO ARGUMENT AT ALL, so
+
+        pwsh -NoProfile -File s.ps1 -Domain $env:DOMAIN -KeyPath $key
+
+    becomes `-Domain -KeyPath $key`, `-Domain` binds the string `-KeyPath`,
+    and every later argument shifts one position to the left. Ledger **L214**.
+    Measured on pwsh 7.4.6 and 7.6.4, both hosts agreeing, variable absent:
+
+        & pwsh -NoProfile -File s.ps1 -P $env:X -Q 'lit'   Missing an argument
+        $a = @('-P', $env:X, '-Q', 'lit'); & pwsh -File s.ps1 @a
+                                                          Missing an argument
+        & ./s.ps1 -P $env:X -Q 'lit'                       P=[] Q=[lit]  SAFE
+
+    `#1146` / `#1148` fixed 17 call sites by hand. This guard is what stops the
+    18th, and every remaining #1080 burn-down manufactures a fresh instance of
+    the hazard the moment it converts an interpolation into an `env:` mapping.
+
+WHY THE INVOCATION FORM IS MODELLED AND NOT ASSUMED
+    PowerShell's own binder accepts the empty value and shifts nothing; only a
+    native command drops it. So `& .\scripts\x.ps1 -P $env:X` is CORRECT CODE
+    and reporting it would be a false finding — `225-whmcs-domain-order-url-verify.yml`
+    and `226-whmcs-application-triage.yml` are the two live sites that a
+    form-blind matcher flags, and `226`'s `orderids` is LEGITIMATELY empty in
+    its default `report` mode. "Fixing" it would break the common path. They
+    are not findings here and they are not freeze entries either: a freeze
+    entry records a defect, and there is no defect to record.
+
+WHY BOTH CALL FORMS ARE MATCHED
+    `@('-ApiUrl', $env:X, '-OutputFile', $out)` splatted into a native command
+    drops the empty element exactly like the direct form does, and there is no
+    `-Param $env:VAR` adjacency for a regex to see. That is precisely how
+    #1146's hand count missed two live holes in `213`. On `main` the array form
+    is the majority of all parameter-position references, so a checker written
+    to the direct spelling would freeze the bigger half as compliant.
+
+WHAT COUNTS AS A GUARD
+    Anything that makes the reference non-empty at the call site, read out of
+    the AST rather than pattern-matched, because this tree spells it three ways
+    and all three are real guards:
+
+      * fail closed   if ([string]::IsNullOrWhiteSpace($env:X)) { …; exit 1 }
+      * default fill  if ([string]::IsNullOrWhiteSpace($env:X)) { $env:X = '…' }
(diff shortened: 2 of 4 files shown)


clarkemoyer:
Copilot's finding is correct and now fixed in `91f44f3`. Both of its comments (the one on line 86 and the suppressed one on line 177) are the same defect, so this is one fix.

**What was wrong.** `CONVERTED_1148` was a dict comprehension keyed by `(workflow, variable)`, and `213-whmcs-zeffy-payments-import-draft.yml` converted **four** distinct steps that all map `WHMCS_API_URL` — `Export clients`, `Export transactions`, `Export invoices`, `Lookup invoices`. Later entries overwrote earlier ones, so exactly one of the four was ever asserted. A regression in `Export invoices` would have passed as long as `Export clients` still held. That is the vacuous-green shape the list exists to prevent, one level down from where I was looking for it — the test that guards criterion 2 had the defect criterion 2 is about.

**The fix**, keyed the way `test_1146`'s own `SITES` table keys it:

- `CONVERTED_1148` is 17 rows of `(workflow, job, step substring, variable)`, and the test asserts **exactly one** matching reference per row. Zero means a converted site is not seen at all; more than one means the step substring stopped identifying a single call site and the row is no longer asserting what it names.
- `len(CONVERTED_1148) == 17` is asserted, because a short table is a criterion-2 assertion that quietly covers less than it claims — which is precisely how this got past me.
- A second test pins the population across those 14 workflows at an **exact** 18: the 17 converted plus `221`'s separately-guarded `WHMCS_API_URL`, which #1146's table covers under `QUERY`. Deliberately exact, not a floor — a `>=` would let a new unguarded reference appear inside an already-converted workflow and still read as "all 17 present and guarded", the reassuring direction.

**Discrimination measured, not assumed.** Three in-process mutations, each detected, baseline green:

| mutation | result |
| --- | --- |
| drop `213`'s `Export invoices` row — the exact defect reported | detected by both the count and the population assertion |
| rename a step substring so it matches nothing | detected, naming the row: `expected exactly one reference, found 0` |
| understate the accounted-for extra (`1` → `0`) | detected: `expected 17 … found 18` |

`tests/workflow-logic/run_all.py` rc=0, 67 modules, **1335 PASS / 0 FAIL** (the module goes 28 → 29 cases). `prettier@3.8.1 --check .` rc=0.

One correction to my own PR body while I am here: it said "27 cases". The module had **28** before this commit and has **29** now; I wrote that number from memory instead of counting it. Body updated.

---

clarkemoyer:
**Conductor run 138 — reviewed by reproducing the defect, not by reading the wiring.**

Per AGENTS.md, a guard is only reviewed once the original defect is put back and the guard is watched
to fail. Done in a throwaway worktree at `91f44f3` (`mktemp -d`, never your branch):

| probe | result |
| --- | --- |
| baseline | `rc=0` — 105 files, 68 PowerShell bodies, 62 references (38 guarded / 24 frozen / 17 write-lane) |
| test module | 29 PASS / 0 FAIL |
| **plant**: delete the 3-line `IsNullOrWhiteSpace($env:WHMCS_API_URL)` guard in `201-whmcs-export-domains.yml` | **`rc=1`**, naming `201-whmcs-export-domains.yml :: export_domains :: step 2 "Export domains (read-only)" [direct] $env:WHMCS_API_URL <- ${{ inputs.api_url }} (whmcs-prod-read)` |
| fail-closed: unparseable YAML | `rc=1`, "something it cannot parse is a finding" |
| fail-closed: **no PowerShell host** | `rc=1`, "refuses to report a pass it did not measure" — not a skip |

**The plant was asserted before its result was believed** (L-“prove the plant landed”): guard
occurrences `1 → 0`, `git diff --numstat` `0 3`, and the tree returned to 0 dirty files after each probe.

**The no-PowerShell-host probe took two attempts and the first was a probe bug, not a finding** — worth
recording because it is run 136's L220 shape. Pruning `pwsh` off `PATH` also prunes `python3`: on this
host both resolve to the *same* `WindowsApps` directory, so the run returned `rc=127` /
`env: 'python3': No such file or directory`. Read naively that is "fails closed, rc≠0" — the right
answer for the wrong reason, from a run where the checker never started. Re-probed by invoking python
through its absolute path with `PATH=/usr/bin:/bin`, after asserting `shutil.which('pwsh')` and
`shutil.which('powershell')` were both `None`. Only then is `rc=1` evidence.

**Wiring confirmed at the step level, not the job level.** `Validate Repository` being green does not
prove the new step executed. Job `93418626559` step **14 `Validate an env-mapped dispatch input fails
closed when empty` = success** — so the hosted runner does have a PowerShell host and this measured
rather than self-skipped, which is precisely the failure mode the step's own fail-closed branch exists
to prevent. No `continue-on-error` in the added block.

**Copilot's `CONVERTED_1148` finding is genuinely addressed at `91f44f3`, not merely marked outdated**
(L216 — `isOutdated=true` means re-read the file, never treat it as evidence). The table is now keyed
`(workflow, job, step substring, variable)`, `213`'s four steps are four distinct rows, `len == 17` is
pinned, and `EXTRA_GUARDED_IN_THE_SAME_STEPS = 1` accounts for `221`'s second reference explicitly
rather than absorbing it into a `>=`. Thread resolved.

Also confirmed the design call in the body: the false-finding direction here is **source-offset vs line**
comparison, and getting that wrong reports 12 correct sites as unguarded — which is how a guard gets
switched off by the people it protects. Good call splitting the burn-down into a second PR.

Promoting from draft and queueing. No gates: static analysis over files already in the tree.

Issue #1027, opened 2026-08-03
Hook rule tests have no per-rule polarity check: a rule tested only in the direction it passes is green and wrong (4th instance)


clarkemoyer:
`.claude/hooks/test_hooks.py` has 41 flat `check()` calls and no way to tell that a rule was tested only in the direction it already passes. Both #940 findings were this one shape, and both greens were meaningless.

## The defect

`check(name, script, payload, expect_block)` — the fourth argument *is* the polarity. The 41 calls are a flat list with no grouping, so nothing relates a rule to its cases, and nothing notices when every case for a new rule points the same way.

#940 shipped a `$endCursor` rule with three cases: **block / block / allow**. All three passed against `"$endcursor" not in low` — a substring test that got the actual rule wrong, allowing `$endCursorX` and `$endCursor_2` (real GraphQL variable names that `--paginate` substitutes into neither, which is the exact infinite-page-1 loop the rule exists to stop). The case that would have failed — a *near-miss* that must be blocked — was the one nobody wrote.

This is the fourth instance in this repo of a test that passed against the mutation it existed to catch (see L48).

## Why a flat count cannot catch it

"Does the file contain both polarities?" is trivially true today and will stay true forever. The property that matters is **per rule**, and rule identity is not currently represented anywhere in the test file.

It *is* recoverable, though: `guard_bash.py` reports every refusal through `block(reason)` (line 24) and each of the ~6 call sites passes a distinct reason string. Running a case tells you which rule fired.

## Acceptance criteria

1. `.claude/hooks/test_hooks.py` gains an explicit rule registry — each guard rule declared once with an id, and its cases attached to that id — so that "which rule does this case exercise" is data, not inference from the case name.
2. A meta-test asserts, for every rule in the registry: **at least one case that blocks, and at least one that is allowed.** It must fail with a message naming the rule and the missing direction.
3. A coverage assertion: every `block(` call site in `.claude/hooks/guard_bash.py` is reached by at least one registered blocking case, matched on the emitted reason rather than on the case's name. A rule added with no test must turn this red.
4. **AC3 must not be satisfied by counting.** Hard-coding the number of `block(` sites, or asserting `len(rules) == 6`, fails the moment a rule is added and teaches the next author to bump a constant. Derive the sites from the source.
[...]

PR #1157, opened 2026-08-10
fix(security): fail closed on an empty dispatch input at the 24 frozen sites (#1150)

clarkemoyer:
Refs #1150. The second half of that issue — #1155 landed the checker and the freeze; this burns the freeze down to zero and deletes it.

## What this is

#1155 measured **24 unguarded parameter-position references across 10 workflows, 17 of them on a write lane**, and froze them in `KNOWN_UNGUARDED`. A freeze entry is a defect awaiting a fix, not a baseline. All 24 are fixed here and `KNOWN_UNGUARDED` is now `{}`.

The hazard is ledger **L214**, and it is a statement about **arity, not content**: in a NATIVE PowerShell invocation an empty `$env:X` is not an empty argument, it is *no argument at all*, so `-Domain` binds the next token and every later parameter shifts one position left. `$ErrorActionPreference = 'Stop'` does not propagate a native command's failure, so the step can **exit 0 having run nothing** — which is why this is load-bearing rather than tidy.

## Two remedies, one property

The property is that the reference is never empty at the call site. Which remedy achieves it depends on the input's own declaration, matching the shapes #1148 established.

| remedy | sites | why |
| --- | --- | --- |
| fail closed — `::error::` naming the variable, `exit 1` | 23 | the input is `required: true`; a real dispatch cannot blank it, a deleted or misspelled `env:` mapping can |
| default fill — `$env:ACCOUNT_ID = '4702611686'` | 1 | 503's `account_id` is `required: false` with that documented default **and** `Mandatory` on the script's param block. Failing closed would make an optional input mandatory |

Emitted as `Write-Output '::error::…'` rather than `throw`, so the assertion is on the text and not on pwsh's error renderer (ledger **L207**).

**One site's remedy is branch-local.** 124's `urls` is `required: false` and legitimately empty when `purge_everything=true` — but in the `else` branch it is mandatory, and empty there does not mean "purge nothing", it means the element vanishes from the splat and `-UrlList` swallows the next token. The guard lives in that branch and says so.

### Sites

| workflow | job(s) | variables | lane |
| --- | --- | --- | --- |
| `111-dns-create-redirect-rule.yml` | `preview`, `apply` | `INPUT_SOURCE`, `INPUT_TARGET` | write |
| `122-cloudflare-zone-member-add.yml` | `preview`, `apply` | `INPUT_DOMAIN`, `INPUT_EMAIL` | write |
| `124-cloudflare-cache-purge.yml` | `preview`, `apply` | `INPUT_DOMAIN`, `INPUT_URLS` | write |
| `125-cloudflare-cache-rules-audit.yml` | `audit` | `INPUT_DOMAIN` | read |
| `205-whmcs-ticket-open.yml` | `open_ticket` | `TICKET_SUBJECT`, `TICKET_MESSAGE` | write |
| `211-whmcs-order-update.yml` | `order_update` | `ACTION` | write |
| `230-whmcs-record-field-set.yml` | `record_field_set` | `TARGET`, `FIELD`, `VALUE` | write |
| `305-m365-add-tenant-domain.yml` | `add-tenant-domain` | `INPUT_DOMAIN` | write |
| `503-google-gtm-provision.yml` | `provision` | `DOMAIN`, `MEASUREMENT_ID`, `ACCOUNT_ID`\* | write |
| `505-google-ga-property-provision.yml` | `provision` | `DOMAIN` | write |

\* the default-fill site. In 503/505 the guard is placed **before** the Key Vault fetch: a dispatch that cannot succeed should not mint a write-scoped credential first.

## Why the checker's green is not the evidence

`scripts/check-workflow-empty-input-guard.py` now reports `62 guarded, 0 unguarded` — but that is a statement about a static analyser, and a too-permissive guard predicate reports a clean tree for a tree full of holes, in the reassuring direction.

`tests/workflow-logic/test_1151_empty_input_burn_down.py` closes that gap by running the **real `run:` bodies out of the committed workflow files** against stub callees, asserting **both directions** per site (#1027):

- **variable removed** → the body exits 1, emits its own `::error::<VAR> is empty` line, and the callee is never reached;
- **variable present** → that line is absent and the callee **is** reached.

The second half is what makes the first half's "callee not reached" mean anything: without it, a body that could never reach its callee under this harness would satisfy the removal case for the wrong reason.

Mechanics carried over deliberately: "empty" is always reproduced by **removing** the variable, never by assigning `''` (on Windows an assigned empty string leaves the variable present and the hazard does not reproduce — #1150 pitfall 3); the environment is inherited then layered, never a fresh minimal dict (**L35**); and no failure case asserts a non-zero exit code alone — each also asserts the guard's own text *and* that PowerShell parsed the body, because a parse error exits non-zero **and** echoes the offending source line back, which can put the searched-for marker into stdout for reasons unrelated to the guard (**L203**).

### The tests were mutation-tested, not just run

Three mutations on byte snapshots (restored from the snapshot, never `git checkout --` — **L182**), each parsed as YAML before its result was believed:

| mutation | outcome |
| --- | --- |
| delete 124 `apply`'s `INPUT_URLS` guard | red — `test_each_site_fails_closed_when_its_variable_is_removed`, naming `124 :: apply :: INPUT_URLS` |
| point 503's default fill at a wrong account id | red — `test_the_default_fill_site_supplies_the_workflows_own_default` |
| neuter 205's `TICKET_SUBJECT` guard to `if ($false)` | red — `test_each_site_fails_closed_when_its_variable_is_removed`, naming `205 :: open_ticket :: TICKET_SUBJECT` |

Two distinct failure sets across three mutations, so the module discriminates rather than failing wholesale; the tree came back byte-identical and green after each.

## Also changed

- `test_the_freeze_is_not_empty_and_names_the_write_lanes` → `test_the_freeze_is_empty_because_the_burn_down_landed`. Its original job was preventing vacuity, and an empty freeze makes `compare()` vacuous in the stale direction — so the replacement asserts the three things that would otherwise be indistinguishable from a checker that stopped working: the population is non-empty, none of it is unguarded, and write lanes are still classified as write lanes.
- The checker's OK summary is now conditional on the freeze. Saying "this is a FREEZE, not an endorsement" over an **empty** freeze trains the reader to discount a green run that has actually earned its green.

## Verification

`rc` read directly, never through a pipe (**L50**). pwsh 7.4.6 installed from the release tarball, so nothing here is a recorded skip.

```
scripts/check-workflow-empty-input-guard.py     rc=0   62 guarded, 0 unguarded, freeze empty
scripts/check-workflow-input-interpolation.py   rc=0   30 / 90 / 15 — unchanged, as required
tests/workflow-logic/run_all.py                 rc=0   68 modules, 1343 PASS, 0 FAIL, 0 SKIP
tests/workflow-logic/test_1151_...py            rc=0   8 tests, + 3 mutations all detected
npx --yes prettier@3.8.1 --check .              rc=0
actionlint 1.7.7 (10 changed workflows)         rc=0
every other scripts/check-*.py                  rc=0
```

One pre-existing exception unrelated to this diff: `scripts/check-environment-protection.py` exits 1 in this sandbox because the egress proxy blocks `repos/.../environments` (`403 Access to this GitHub API path is not permitted through this proxy`). It correctly refuses to report a clean bill of health rather than passing on a read it could not complete.

No `.ps1` files changed, so PSScriptAnalyzer/Invoke-Formatter have nothing to say here.

## Gates

**None entered, none needed.** No workflow dispatched, no environment, no credential read — the tests stub `az` precisely so the 503/505 bodies reach their callee without touching Key Vault.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EaxwA3vDsoyYPg3A2nsWFg)_

### DIFF
--- a/.github/workflows/111-dns-create-redirect-rule.yml
+++ b/.github/workflows/111-dns-create-redirect-rule.yml
@@ -67,6 +67,21 @@ jobs:
           INPUT_INCLUDE_WWW: ${{ inputs.include_www }}
           INPUT_PRESERVE_QS: ${{ inputs.preserve_query_string }}
         run: |
+          # Fail closed on an empty mapping (#1150, ledger L214). The inputs below are
+          # `required: true`, so a real dispatch cannot blank them -- a deleted or
+          # misspelled `env:` mapping can. `pwsh -File ... @args` is a NATIVE command, so an
+          # empty value is not an empty argument -- it is no argument at all:
+          # `-SourceDomain` binds the NEXT token and every later parameter shifts one
+          # position left. The interpolation guard stays green either way, because nothing
+          # is interpolated.
+          if ([string]::IsNullOrWhiteSpace($env:INPUT_SOURCE)) {
+            Write-Output '::error::INPUT_SOURCE is empty: the step-level env: mapping is missing or misnamed, or the dispatch supplied a blank source_domain. Refusing to call Set-CloudflareRedirectRule.ps1 with an empty -SourceDomain.'
+            exit 1
+          }
+          if ([string]::IsNullOrWhiteSpace($env:INPUT_TARGET)) {
+            Write-Output '::error::INPUT_TARGET is empty: the step-level env: mapping is missing or misnamed, or the dispatch supplied a blank target_domain. Refusing to call Set-CloudflareRedirectRule.ps1 with an empty -TargetDomain.'
+            exit 1
+          }
           $args = @(
             '-SourceDomain', $env:INPUT_SOURCE,
             '-TargetDomain', $env:INPUT_TARGET,
@@ -100,6 +115,21 @@ jobs:
           INPUT_INCLUDE_WWW: ${{ inputs.include_www }}
           INPUT_PRESERVE_QS: ${{ inputs.preserve_query_string }}
         run: |
+          # Fail closed on an empty mapping (#1150, ledger L214). The inputs below are
+          # `required: true`, so a real dispatch cannot blank them -- a deleted or
+          # misspelled `env:` mapping can. `pwsh -File ... @args` is a NATIVE command, so an
+          # empty value is not an empty argument -- it is no argument at all:
+          # `-SourceDomain` binds the NEXT token and every later parameter shifts one
+          # position left. The interpolation guard stays green either way, because nothing
+          # is interpolated.
+          if ([string]::IsNullOrWhiteSpace($env:INPUT_SOURCE)) {
+            Write-Output '::error::INPUT_SOURCE is empty: the step-level env: mapping is missing or misnamed, or the dispatch supplied a blank source_domain. Refusing to call Set-CloudflareRedirectRule.ps1 with an empty -SourceDomain.'
+            exit 1
+          }
+          if ([string]::IsNullOrWhiteSpace($env:INPUT_TARGET)) {
+            Write-Output '::error::INPUT_TARGET is empty: the step-level env: mapping is missing or misnamed, or the dispatch supplied a blank target_domain. Refusing to call Set-CloudflareRedirectRule.ps1 with an empty -TargetDomain.'
+            exit 1
+          }
           $args = @(
             '-SourceDomain', $env:INPUT_SOURCE,
             '-TargetDomain', $env:INPUT_TARGET,
--- a/.github/workflows/122-cloudflare-zone-member-add.yml
+++ b/.github/workflows/122-cloudflare-zone-member-add.yml
@@ -66,6 +66,21 @@ jobs:
           INPUT_PERMISSION_GROUP: ${{ inputs.permission_group }}
         run: |
           $ErrorActionPreference = 'Stop'
+
+          # Fail closed on an empty mapping (#1150, ledger L214). The inputs below are
+          # `required: true`, so a real dispatch cannot blank them -- a deleted or
+          # misspelled `env:` mapping can. `& pwsh @scriptArgs` is a NATIVE command, so an
+          # empty value is not an empty argument -- it is no argument at all: `-Domain`
+          # binds the NEXT token and every later parameter shifts one position left. The
+          # interpolation guard stays green either way, because nothing is interpolated.
+          if ([string]::IsNullOrWhiteSpace($env:INPUT_DOMAIN)) {
+            Write-Output '::error::INPUT_DOMAIN is empty: the step-level env: mapping is missing or misnamed, or the dispatch supplied a blank domain. Refusing to call cloudflare-zone-member-add.ps1 with an empty -Domain.'
+            exit 1
+          }
+          if ([string]::IsNullOrWhiteSpace($env:INPUT_EMAIL)) {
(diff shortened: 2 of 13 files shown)


clarkemoyer:
Reviewed at `b7e4a91`, re-fetched immediately before posting and unchanged (L216). **Approving and queueing.**

I did not review this by reading the wiring — with `KNOWN_UNGUARDED` now `{}`, the whole protection rests on whether the checker still *detects*, so I reintroduced the defect twice.

**Control first.** Untouched PR head: `rc=0`, 105 files scanned, 62 parameter-position references, 62 guarded, 0 frozen. Every mutation below is bracketed by that control and an explicit `git restore --source=HEAD` (L209).

**M-A — delete both `INPUT_SOURCE` guards in 111** (anchor asserted present exactly twice; use preserved; `yaml.safe_load` clean). `rc=1`, naming exactly:

```
111-dns-create-redirect-rule.yml: NOT in KNOWN_UNGUARDED — apply::INPUT_SOURCE, preview::INPUT_SOURCE
[W] … :: apply :: … [array-splat] $env:INPUT_SOURCE <- ${{ inputs.source_domain }} (cloudflare-prod-write)
```

Correctly-named subset, write lane marked. **An empty freeze is not a vacuous guard.**

**M-B — the relational one, and my first attempt was a false positive worth recording.** Moving 122's `INPUT_DOMAIN` guard below its use *inside the array literal* left a dangling `,`; the checker exited 1 on `PowerShell does not parse`, which scores as a detection and is not one (L203, flattering direction). Redone by relocating the guard below the complete `& pwsh @scriptArgs` statement, with three post-conditions asserted before reading any exit code: mutant is valid YAML, the checker's own parser reports no parse failure, and `guard_index (3805) > use_index (3630)` — the relation I meant to break (L221). Result: `rc=1` naming **`preview::INPUT_DOMAIN` alone**, while `apply`'s guard — still above its use — stays silent. That is the ordering model demonstrated, not assumed, and the sibling disagreement that makes it credible.

Incidentally your `test_unparseable_powershell_is_a_finding` is what made my bad mutation legible instead of silently wrong.

**Judgment calls, both verified against the source rather than the prose:**
- 503 `account_id` — `required: false`, `default: '4702611686'` at `:26`, `[Parameter(Mandatory)][string]$AccountId` at `google-gtm-provision.ps1:24`. Default-fill is right; failing closed would make an optional input mandatory. I was going to flag the hardcoded literal at `:108` as drift-prone and `test_the_default_fill_site_supplies_the_workflows_own_default` already pins it, with a comment naming the exact hazard ("provisions into the wrong GTM account in fact").
- 124 `urls` — `required: false`, guard is inside the `else` branch only. Correct: empty there is not "purge nothing", the element vanishes from the splat.

**Suites, run locally on the PR head with a real PowerShell host (`pwsh` 7.6.4 — confirmed present and executed, not assumed):** `test_1150` all PASS, `test_1151` **8 PASS / 0 FAIL / 0 SKIP**. Zero skips is the part that matters — a self-skipping security module proves nothing (L219 family). `test_a_removed_variable_is_not_the_same_as_an_assigned_empty_string` encodes L220 directly.

No review threads, all checks green, 0 behind `main`. Queueing.

Issue #891, opened 2026-07-26
[invalid] created in error — closed immediately


clarkemoyer:
Created by mistake by an agent session (wrong tool method — a comment was intended). No content, no work tracked here. Closed immediately; nothing references it.

PR #1158, opened 2026-08-10
docs(ledger): L226/L227 — merge-queue membership is not `autoMergeRequest`, and who resolves a path

clarkemoyer:
Two rows from run 139. Both are Conductor-side findings from this run's own work, and neither is enforceable below the doc tier — the reasons are stated in each row's `Enforced by` cell rather than left implied.

**L226 — `autoMergeRequest` is silent about merge-queue membership.** Run 138 hit this, chose not to ship it, and set the bar explicitly: *"if it recurs a third time it earns one"*. It recurred twice in run 139. On #1157, `gh pr merge --auto` printed only the merge-strategy note and `--json autoMergeRequest` then read `null` with `mergeStateStatus: CLEAN` — the exact signature of a command that did nothing. It had queued: `refs/heads/gh-readonly-queue/main/pr-1157-de17662…` existed and a re-fire answered *"already queued to merge"*. Independently, ffcadmin #891 read `autoMergeRequest: false` with `state: OPEN, mergedAt: null` and merged seconds later. The remedy is the queue branch rather than the field, because its name carries the base SHA and therefore answers *which* `main` it queued onto.

**L227 — a true recorded hazard that still leaves the case that bites.** `AGENTS.md` warns that `/tmp` resolves differently for git-bash than for a Windows `python3`. Run 139 read it, and still lost two cycles: running the guard from a `mktemp -d` worktree under bash works, while the identical path passed as `subprocess.run(cwd=…)` raises `NotADirectoryError: [WinError 267]`. The working case is the one you try first, so the recorded hazard reads as not applying. The discriminator is whether a shell or `CreateProcess` resolves the path.

## Verification

- `python3 tests/workflow-logic/test_lessons_ledger.py` — **48 PASS / 0 FAIL**, control re-derived on clean `main` first (also 48/0) rather than assumed.
- `npx prettier@3.8.1 --check .` clean; `main` was already clean.
- 0 commits behind `origin/main` at push time (`de17662`).

## Two notes for the reviewer

**The diff is 93+/91− for two added rows, and 3+/1− under `--ignore-all-space`.** All of the rest is prettier re-padding every cell in that table, because these rows are wider than the previous widest and the table is column-aligned. Nothing outside the two rows changed in substance.

**The guard caught a real defect in my first draft, which is why it ran before the PR was opened rather than after CI complained.** L227's original text quoted the offending `sed` command verbatim; its three delimiter characters made a **7-cell row**, and prettier propagated that width to the whole table's delimiter row 90 lines earlier. `test_every_ledger_row_has_the_column_count_its_header_declares` named it precisely, along with three consequent failures whose real cause was the same cell shift. That episode is now recorded inside L227's own evidence cell. The rewrite asserts `len(cells) == 4` and no stray pipe before inserting.

Refs #719.


clarkemoyer:
Branch updated from `main` after #1157 merged mid-run — `Phantom Revert Guard` had gone red on the 2-commit-behind clause, and is green on the updated head. Now 0 behind / 2 ahead. `test_lessons_ledger.py` re-run on the merged head: **48 PASS / 0 FAIL**.

Left as a draft for the next run to promote, per Step 7.

clarkemoyer:
Sharpened L226 with a third observation from later in the same run.

#1159 was **armed but not yet queued** (blocked on a pending `Validate Repository`) and its `autoMergeRequest` read **populated** — `enabledAt 13:42:09Z` — the mirror image of #1157, which read `null` while genuinely queued. The reading that reconciles all three: the field tracks the outstanding auto-merge *request*, so it is populated while the PR waits to queue and clears once the request has been consumed into a queue entry.

If that is right, the field is not merely unreliable — it reads **backwards**: fullest before anything has happened, empty once it has. That is a stronger claim than the row originally made, so it is written as the inference it is, from three observations in one run, rather than as established behaviour.

Ledger guard re-run: 48 PASS / 0 FAIL.

Issue #962, opened 2026-07-31
Pinning the parent's decode without the child's encode is half a fix — 3 subprocess sites can return proc.stdout=None on Windows


clarkemoyer:
## Problem

The #945 sweep (landed as https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/953)
pinned `encoding="utf-8"` on 43 text-mode `subprocess` calls, and
`scripts/check-subprocess-encoding.py` now keeps them pinned. That fixed the **parent's decode**.
It did not fix the **child's encode**, and for a Python child on Windows those are two different
settings.

A Python child writes its stdout in the console codepage. So when the parent forces UTF-8 and the
child emits cp1252, the pin converts a silently-mojibake'd string into a hard failure:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x97 in position 200: invalid start byte
```

`0x97` is a cp1252 em dash. Worse, the exception is raised on `subprocess`'s reader **thread**, so
`subprocess.run` still returns — with `proc.stdout is None`. The caller then dies somewhere else
entirely:

```
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

which points at the caller's arithmetic, not at an encoding problem. That is a bad failure to debug.

Observed for real on 2026-07-31 (run 60) in
`tests/workflow-logic/test_workflow_number_uniqueness.py`, whose child is
`scripts/check-workflow-doc-consistency.py`. That module is fixed on
https://github.com/FreeForCharity/FFC-Cloudflare-Automation/pull/917 by also passing
`env={**os.environ, "PYTHONIOENCODING": "utf-8"}`, which makes both ends agree on every platform.

**CI never sees this** — ubuntu-latest is UTF-8 on both ends — so the suite stays green while the
same tree crashes on a maintainer's machine. Same blind spot as #945 / ledger L35.

## Scope (one PR)

Three call sites on `main` have the at-risk shape (parent pins `encoding=`, child is
`sys.executable`/`python`, child encoding unpinned):

- `tests/workflow-logic/run_all.py`
- `tests/workflow-logic/test_env_secret_reference_resolution.py`
- `tests/workflow-logic/test_powershell_command_resolution.py`

Add `env={**os.environ, "PYTHONIOENCODING": "utf-8"}` to each (importing `os` where needed).

Then extend `scripts/check-subprocess-encoding.py` so the guard covers the other half: when a
[...]

PR #1162, opened 2026-08-10
fix(tests): run_all.py must survive a module printing an unencodable character

clarkemoyer:
## Summary

`run_all.py` can die partway through the suite, print no summary, and have that read as a clean run.

#962 pinned two encoding settings — the child's encode (`PYTHONIOENCODING=utf-8` in the child env) and the parent's decode (`encoding="utf-8", errors="replace"` on `subprocess.run`). Both are still correct. There is a **third**, and it is the one that bites: having decoded a module's output cleanly, the harness **re-encodes** it at `sys.stdout.write(proc.stdout)` (`run_all.py:184`) using whatever encoding the *parent's* stdout happens to have. Redirect a run on Windows and that is cp1252, so a module printing `→` (U+2192) raises `UnicodeEncodeError` inside the module loop.

The existing comment at that call site is worth reading, because it names two of the three:

> The parent's decode and the child's encode are two settings.

**The traceback is not the damage.** The exception escapes the loop, so:

- the suite stops at whichever module reached the character first,
- every module after it never runs,
- and `::error::workflow-logic tests failed: …` **never prints**.

Anything reading that output for the summary line — a reviewer, a script, a later step — finds no failures and concludes there were none. A crashed suite and a clean suite are indistinguishable to a reader who greps for *failures* rather than for the *presence of the summary*.

### How it was found

Live, this run, while baselining #1161. The first baseline on clean `main` reported `rc=1` and **zero failing modules**; it had actually died at `test_738_fleet_smoke_drift.py` with 31 modules never executed. Re-run with `PYTHONIOENCODING=utf-8` on the parent, the same tree reports **1165 PASS / 83 FAIL / 16 failing modules**. The "zero failing modules" reading was produced by a real, silent, 31-module hole.

This is the same cp1252 family as L35 and the CLAUDE.md notes, but a **new vector**: those cover reading a file, and a one-off `python3 -c` printing to a console. This one is a committed harness re-encoding a *child's already-decoded* output, and no existing guard looks there.

## The fix

`_survive_unencodable_output()` reconfigures `sys.stdout`/`sys.stderr` with `errors="replace"` at the top of `main()`.

`errors="replace"` and **not** `encoding="utf-8"`, deliberately: re-encoding is only ever a display concern here (every comparison the file makes is on the decoded `str`), and forcing utf-8 bytes at a genuinely cp1252 console would trade a crash for mojibake on every box that has one. Replacement degrades one character and keeps the roster and the summary readable.

## Tests

Three new cases in `test_run_all_roster.py`, plus one guarding against the fix decaying.

**They cannot use the existing `_run_all_on` helper, and that is the point.** That helper pins `PYTHONIOENCODING=utf-8` on the child it launches — and here the child *is* `run_all.py`. So every existing test in this file runs the harness with a utf-8 stdout, which is exactly why this class of failure survived undetected. The new helper `_run_all_with_parent_encoding()` drives it with cp1252 instead.

| test | asserts |
| --- | --- |
| `test_the_arrow_is_genuinely_unencodable_in_cp1252` | positive control — if U+2192 ever encodes cleanly, or the fixture stops emitting it, the survival test would go green measuring nothing |
| `test_a_module_printing_an_unencodable_character_does_not_kill_the_suite` | no `UnicodeEncodeError`; **the module after the offending one still ran**; the summary line is present |
| `test_a_real_failure_is_still_reported_under_a_narrow_stdout` | a red module is still named — so the fix cannot decay into swallowing write errors, which would reach the same silent-clean outcome by another route |

The middle one asserts on the *later module* rather than only on the absence of a traceback, because the modules that never execute are the actual damage.

This runs on Linux CI: the narrow stdout is forced through `PYTHONIOENCODING=cp1252` on the child, so it does not depend on the host's console codepage.

## Test plan

- `test_run_all_roster.py`: **21 PASS / 0 FAIL** (was 18/0).
- **Mutation:** deleting the `_survive_unencodable_output()` call from `main()` flips exactly `test_a_module_printing_an_unencodable_character_does_not_kill_the_suite` and `test_a_real_failure_is_still_reported_under_a_narrow_stdout`, and leaves the positive control green. Mutant asserted to parse before its exit code was believed; anchor count checked (1) before substituting; restored from a byte-exact copy, not `read_text`/`write_text`.
- `prettier@3.8.1 --check` clean on both files; `tr -cd '\r' | wc -c` → **0** on both.
- Full `run_all.py` suite: results posted as a comment once it finishes (it exceeds a 10-minute foreground cap on this host).


clarkemoyer:
### Full-suite result (the test-plan item left open at PR time)

`run_all.py` on this Windows host, branch vs clean `origin/main` (`23204c2`), same command, `PYTHONIOENCODING=utf-8` on the parent so the baseline could complete at all:

| | PASS | FAIL | failing modules |
| --- | --- | --- | --- |
| clean `main` | 1165 | 83 | 16 |
| this branch | **1168** | 83 | 16 |

The failing-module set is **identical** — the 16 are pre-existing and environmental on this host (bash-version and git-output differences), and `+3` is exactly the three new tests. No new failures.

One detail worth stating because it is the PR's own subject: the branch run **printed its summary line**, which is the thing whose absence started this. Asserting `summary line present: 1` rather than `failures: 0` is the check the diagnosis actually calls for — a run with zero failures and a run that died before it could report any are the same output otherwise.

